Issue #71, opened 2020-01-16
Timing of block downloads leaks whether or not any notes belonging to the wallet were discovered


defuse:
This issue was originally discovered and reported by Florian Tramèr. I'm making it public since no users are currently at risk.

The issue is that the SDK will:

1. Download new blocks,
2. process those blocks, taking an amount of time that depends on how many notes belong to this wallet,
3. wait a fixed length of time,
4. make a request to lightwalletd for more new blocks.

A passive network adversary can measure the time between (1) and (4) to determine whether any notes in the blocks belonged to the wallet. To fix this, either

a. Decouple the network code from the block processing code, such that downloaded blocks are queued up to be processed on a separate thread and the processing won't affect the timing of network requests,
b. Florian's suggestion, to request blocks at absolute time intervals, i.e. instead of sleeping for 20 seconds between requests, request new blocks at times \*:00, \*:20, \*:40. *In order for this to not leak information, the block processing MUST always complete before the next request is due to be made.*

I recommend (b) since it's easier and we're about to run into even harder side-channel challenges with memo downloading and mempool access.

pacu:
@defuse I can't assure that timers will fire on 0 20 40 exactly. Apple requieres a minimum tolerance of delay for the timer, since `Timer` is not a real-time timer after all and depends on a run loop. I think that unless we can find more info on this, we will have to settle with random jitter for now. 

if we manage to set a 0 20 40 Timer scheme (there's no direct api for it) and the recommended 10% offset tolerance that Apple recommends, the offset for each user becomes information as well, doesn't it?

PR #72, opened 2020-01-17
Randomize POLL_INTERVAL

pacu:
Addresses #71 

There are better approaches at the expense of more research time.


defuse:
#71 should be kept open since the problem is not fully resolved yet.

PR #74, opened 2020-01-30
Fix - CompactBlock Processsor loops indefinitely on reorg

### LINKED ISSUE #73: CompactBlockProcessor infinite loop on reorg
```` swift
  func determineLowerBound(errorHeight: Int) -> BlockHeight {
        let offset = min(ZcashSDK.MAX_REORG_SIZE, ZcashSDK.DEFAULT_REWIND_DISTANCE * (consecutiveChainValidationErrors + 1))
        return max(errorHeight - offset, lowerBoundHeight ?? ZcashSDK.SAPLING_ACTIVATION_HEIGHT)
    }
````

that last max() should be a safeguard to rewinding past sapling activation, but has the side effect of not returning the correct rewind height.

The Fix
```
func determineLowerBound(errorHeight: Int) -> BlockHeight {
        let offset = min(ZcashSDK.MAX_REORG_SIZE, ZcashSDK.DEFAULT_REWIND_DISTANCE * (consecutiveChainValidationErrors + 1))
        return max(errorHeight - offset, self.config.walletBirthday)
    }
```
this wallet should never rewind further than its birthday which in the worst case will be SAPLING_ACTIVATION.
I was so defensive that I introduced a bug.

### PR DESCRIPTION
**Pull request by pacu**
Fixes #73

### DIFF
--- a/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
+++ b/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
@@ -323,7 +323,7 @@ public class CompactBlockProcessor {
         
         // rewind
         
-        let rewindHeight = determineLowerBound(errorHeight: height)
+        let rewindHeight = determineLowerBound(errorHeight: height, consecutiveErrors: consecutiveChainValidationErrors, walletBirthday: self.config.walletBirthday)
         guard rustBackend.rewindToHeight(dbData: config.dataDb, height: Int32(rewindHeight)) else {
             fail(rustBackend.lastError() ?? RustWeldingError.genericError(message: "unknown error rewinding to height \(height)"))
             return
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift [hunk 1]
```diff
@@ -342,11 +342,11 @@ public class CompactBlockProcessor {
         }
     }
     
-    func determineLowerBound(errorHeight: Int) -> BlockHeight {
-        let offset = min(ZcashSDK.MAX_REORG_SIZE, ZcashSDK.DEFAULT_REWIND_DISTANCE * (consecutiveChainValidationErrors + 1))
-        return max(errorHeight - offset, lowerBoundHeight ?? ZcashSDK.SAPLING_ACTIVATION_HEIGHT)
+    func determineLowerBound(errorHeight: Int, consecutiveErrors: Int, walletBirthday: BlockHeight) -> BlockHeight {
+            let offset = min(ZcashSDK.MAX_REORG_SIZE, ZcashSDK.DEFAULT_REWIND_DISTANCE * (consecutiveErrors + 1))
+            return max(errorHeight - offset, walletBirthday)
```
defuse: A reorg could happen at the wallet's birthday, so we should allow it to go back a little further, e.g. `walletBirthday - ZcashSDK.MAX_REORG_SIZE`.

PR #83, opened 2020-02-11
Feature/note spending v7

pacu:
Str4d, please review this changes! 
Apologies for my rusty rust!


pacu:
@str4d don't mind anything that's not a .rs file! thanks

@lindanlee I updated the public API docs

rust/src/lib.rs
```diff
@@ -219,6 +342,49 @@ pub extern "C" fn zcashlc_get_balance(db_data: *const u8, db_data_len: usize, ac
     unwrap_exc_or(res, -1)
 }
 
+/// Returns 1 when the address is valid and shielded. 
+/// Returns 0 in any other case
+/// Errors when the provided address belongs to another network
+#[no_mangle]
+pub unsafe extern "C" fn zcashlc_isValidShieldedAddress(
+    address: *const c_char,
+) -> u8 {
```
ebfull: It would be better to return a `bool` here.

rust/src/lib.rs
```diff
@@ -219,6 +342,49 @@ pub extern "C" fn zcashlc_get_balance(db_data: *const u8, db_data_len: usize, ac
...
+    unwrap_exc_or(res, 0)
+}
+
+
+/// Returns 1 when the address is valid and transparent. 
+/// Returns 0 in any other case
+#[no_mangle]
+pub unsafe extern "C" fn zcashlc_isValidTransparentAddress(
+    address: *const c_char,
+) -> u8 {
```
ebfull: Return a `bool` instead.

rust/src/lib.rs
```diff
@@ -168,6 +186,111 @@ pub extern "C" fn zcashlc_init_blocks_table(
...
+                ));
+            }
+        };
+
+        let encoded = encode_extended_full_viewing_key(
+            HRP_SAPLING_EXTENDED_FULL_VIEWING_KEY,
+            &extfvk,
+        );
+     
+        Ok(CString::new(encoded).unwrap().into_raw())
```
ebfull: I'm concerned that we're calling `into_raw` here which will cause memory leaks since we never free these strings later on.

pacu: What would be the right approach?

pacu: there are helper methods to free strings created on the rust side from swift. @str4d is that the approach taken on other similar functions?

ebfull: Indeed, here is an example to illustrate what needs to be done.

```rust
use std::ffi::CString;
use std::mem::forget;

type c_char = i8;

fn my_func(accounts: u32) -> *mut *mut c_char {
    // Important: the vector capacity will equal its length when it's freed
    let mut v = Vec::with_capacity(accounts as usize);
    
    for _ in 0..accounts {
        // Create the CString (performs heap allocation)
        let s = CString::new("Howdy, friend!").unwrap();
        
        // Pushes the raw pointer to the CString to the vector, consuming the CString
        v.push(s.into_raw());
    }
    
    // Grab the raw pointer to the slice
    let r = v.as_mut_ptr();
    // Forget the vector so it doesn't deallocate
    forget(v);
    
    // Return raw pointer
    r
}

unsafe fn my_func_free(r: *mut *mut c_char, accounts: u32) {
    let mut v = Vec::from_raw_parts(
        r,
        accounts as usize,
        accounts as usize
    );
    
    // Consumes v to deallocate it at the end of the loop
    for raw in v {
        let s = CString::from_raw(raw);
        drop(s);
    }
}

fn main() {
    const ACCOUNTS: u32 = 100;

    let test = my_func(ACCOUNTS);
    unsafe { my_func_free(test, ACCOUNTS); }
    
    // Bad behavior can be caused if ACCOUNTS changes!
    // let test = my_func(ACCOUNTS);
    // unsafe { my_func_free(test, ACCOUNTS * 2); }
}
```

pacu: does this apply to functions that return `*mut c_char`?  All other functions don't take the as_mut_ptr() -> forget() -> return  pattern as functions returning `*mut *mut c_char` do.

PR #85, opened 2020-02-14
Update troubleshooting

### LINKED ISSUE #81: Update readme instructions for iOS SDK to fix two errors found during test build
Issue 1: Xcode uses a non-interactive script shell, so it cannot find the ZCASH_NETWORK_ENVIRONMENT=TESTNET environment variable listed in ~/.bash_profile. There is a way to set the environment variable for the app build, and also to set the environment variable in the cocoapod dependencies. The former is not the solution, the latter is.

Issue 2: sapling-spend-params and sapling-output-params, two parameters required for the build, are not in the ZcashLightClientKit. It’s not included because we want the parameters needing to be in one place and never bundled into a repo. The user must download them from https://z.cash/downloads/sapling-spend.params
and https://z.cash/downloads/sapling-output.params  and move it to the correct folder. The readme should include these instructions in the meantime. The permanent fix is to write something in the build instructions that fetches the parameters.

pacu:
fixes #81


lindanlee:
Looks great. Nice that you gave people the option to fix manually or do a clean pod install, and the picture is a nice touch.

Issue #127, opened 2020-05-20
EXC_BAD_INSTRUCTION on Send Funds with local lightwalletd


sebastiendan:
Running the sample app connected to a local `lightwalletd` instance, the Send Funds operation fails for a specific error, showing the error in a dialog. Then after a few seconds, the app crashes with the error listed down below.

<img width="383" alt="Screen Shot 2020-05-20 at 10 58 15" src="https://user-images.githubusercontent.com/12198372/82396421-d0f26e00-9a88-11ea-9d10-78275255d6a3.png">

```
Main Thread Checker: UI API called on a background thread: -[UILabel setText:]
PID: 33795, TID: 1227073, Thread name: (none), Queue name: com.apple.root.default-qos, QoS: 0
Backtrace:
4   ZcashLightClientSample              0x000000010d4bd6c4 $s22ZcashLightClientSample18SendViewControllerC19synchronizerUpdatedyy10Foundation12NotificationVF + 852
5   ZcashLightClientSample              0x000000010d4bd78d $s22ZcashLightClientSample18SendViewControllerC19synchronizerUpdatedyy10Foundation12NotificationVFTo + 141
6   CoreFoundation                      0x00007fff23d64f4c __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 12
7   CoreFoundation                      0x00007fff23d643c5 _CFXRegistrationPost1 + 421
8   CoreFoundation                      0x00007fff23d64131 ___CFXNotificationPost_block_invoke + 193
9   CoreFoundation                      0x00007fff23e61513 -[_CFXNotificationRegistrar find:object:observer:enumerator:] + 1795
10  CoreFoundation                      0x00007fff23d63a86 _CFXNotificationPost + 950
11  Foundation                          0x00007fff2590536b -[NSNotificationCenter postNotificationName:object:userInfo:] + 59
12  ZcashLightClientKit                 0x000000010f1ead4d $s19ZcashLightClientKit15SDKSynchronizerC6notify33_01791D28F706D4525E7EE757177F698FLL8progress6heightySf_SitF + 669
13  ZcashLightClientKit                 0x000000010f1e660c $s19ZcashLightClientKit15SDKSynchronizerC16processorUpdatedyy10Foundation12NotificationVF + 1260
[...]

Issue #128, opened 2020-05-20
Verified Balance and total Balance not updated after sending


pacu:
Credits to @ToposWare

Demo App is not reflecting available funds after sending and causes misleading errors in subsequent submissions could be attempted without unspent notes yet available.

![image](https://user-images.githubusercontent.com/484008/82449162-2b87cc00-9a81-11ea-8111-0e567607a623.png)

sebastiendan:
@pacu 
Executing [this SQL request](https://github.com/str4d/librustzcash/blob/note-spending-v7/zcash_client_sqlite/src/transact.rs#L131) on my `ZcashSdk_testnet_data.db` I get no result.

I can provide that db file if that can help.

pacu:
@str4d, can you take a look?

pacu:
bringing some context over from discord conversations:

> This balance of 4.0 ZEC is most likely the result of multiple trials to get ZEC from the testnet faucet. At the time I was going through my first trials (last week), I was using a fork of testnet so didn't receive the tx/coins. After I notified you to get lightwalletd.testnet.z.cash updated I started using a local instance of lightwalletd plugged to a local full node on testnet. Now that I can confirm I'm using the right branch of your testnet (I get the right latestblock), I have the right balance containing the amount coming from the faucet.
What I'm saying with this is that these coins have probably been delivered to my address last week.

pacu:
@sebastiendan once you fixed the lightwalletd issue did you sync from scratch?

sebastiendan:
@pacu Yes, I did (meaning I at least once saw the sync happening from 0 on the sync screen). 

Is there a way on the iOS side to force a re-sync from scratch? Should I delete manually the virtual device data on my disk? (I'm new to iOS dev)

pacu:
the sync always starts from 0, we are not compensating the UI percentage. 
tap the trash can icon on the main screen and restart the app to start over from ground zero

sebastiendan:
I used the trash icon and synced against from zero.

I have now a different error:

<img width="375" alt="Screen Shot 2020-05-22 at 7 55 09" src="https://user-images.githubusercontent.com/12198372/82614052-d9ba7f80-9c01-11ea-97c4-9dc52c1dbbde.png">

pacu:
oh, so, now it's getting to your lwd instance but it's being rejected. that's interesting.

sebastiendan:
My lightwalletd instance outputs only this about the tx:

`
[...]

PR #138, opened 2020-06-04
Enable heartwood.

gmale:
We might want to make further adjustments to conditionally enable this only on testnet.

@pacu see if you can reproduce the error in #128 without this change. If not, then that means the librustzcash changes were enough to fix this. If, instead, it still reproduces, then try this branch. If that clears the issue, then we'll want to discuss how to enable this only for testnet builds.


pacu:
fixed  #127

PR #140, opened 2020-06-05
Issue 136 Null bytes in strings effectively truncate the string from …

pacu:
…librustzcash's point of view

image


defuse:
Aside from the one comment, everything else looks great!

Issue #177, opened 2020-09-02
Init accounts assertion fails   assert!(v.len() == v.capacity());


pacu:
**What is the bug?**
the problem is that assert!(v.len() == v.capacity()); can't be guaranteed, because the underlying allocator may choose to allocate the Vec with a larger capacity than is needed.

**Additional context**
rustc 1.40.0 (73528e339 2019-12-16 -> works
rustc 1.46.0 (04488afe3 2020-08-24) -> fails

The larger problem here is that in order to pass the array of strings from Rust to Swift, what I apparently attempted to do is pass a Vec of CString pointers as a raw pointer itself, and then in Swift we treat it as an unsafe buffer pointer, and then later call back into Rust to free the memory. This is technically "fine" as long as the Vec's capacity matches its length, because we're relying on the caller to specify the Vec's length when freeing it. That's most likely why I asserted here: if the assertion fails, then chunks of the Vec would not be freed.

**Solution**
The proper fix here is to have Swift allocate its equivalent of a Vec<*mut c_char>, and then pass that through to Rust which can use it as &mut[*mut c_char] to fill it, and then later to a freeing function that frees the *mut c_chars.
This involves changes in librustzcash and lib.rs

**workaround**
use  rust 1.40.0 which is known to work well

```
rustup override set 1.40.0
rustup target remove aarch64-apple-ios x86_64-apple-ios
rustup target add aarch64-apple-ios x86_64-apple-ios
```

PR #180, opened 2020-09-02
add workaround to fix rustc problem on init accounts

pacu:
#177 workaround until better solution is  implemented


lindanlee:
LGTM. Thanks for the fix.

PR #241, opened 2021-01-05
Issue 240: reorg logic fails do to ARC deallocating operation

### LINKED ISSUE #240: Operation deallocated before it could report the error causes reorg to fail.
Code exited on this check

````
 guard let validationError = error as? CompactBlockValidationError else {
                LoggerProxy.debug("Warning: validateChain operation returning generic error: \(error)")
                return
            }
```` 
CompactBlockValidationOperation was being deallocated before it could report the error down to the compact block processor. 

This rendered the wallet unusable because it is unable to recover from the error until that race condition clears and the error does not manifest itself.

pacu:
closes #240


pacu:
@ant013 important update!

PR #345, opened 2022-03-02
0.13.0-beta.1 Update changelog.md and podspec version

### LINKED ISSUE #344: release 0.13.0-beta.1
This will contain rust-less build for the SDK

pacu:
Closes #344 #346 

This PR fixes broken CI and tests. Creates Github actions to build the swift package and run the offline tests
# Author

# Reviewer


.github/workflows/swift.yml
```diff
@@ -0,0 +1,15 @@
+name: Swift
+
+on: [push]
```
ccjernigan, line 3: What's the desired behavior for when this runs?  For open pull requests?  After merge to main?

I suspect the desired implementation might be something like:
```
on:
  pull_request:
    paths-ignore:
      - '.github/ISSUE_TEMPLATE/*'
      - '.github/PULL_REQUEST_TEMPLATE.md'
      - 'changelog.md'
      - 'LICENSE'
      - 'README.md'
      - 'responsible_disclosure.md'
      - 'SWIFTLINT.md'
      - 'docs/**'
```

That would provide the behavior of running when a pull request is opened and when another commit is made to a pull request.  It will not run when certain documentation-only changes occur, reducing CI minutes used.  It also will not run when a merge is made to the main branch (assuming that merges to main are only done via pull request, so therefore don't need to be re-tested)

.travis.yml
```diff
@@ -1,19 +1,11 @@
...
     - sourcery
-install:
-  - ${TRAVIS_BUILD_DIR}/Scripts/travis/rust_setup.sh
-  - ${TRAVIS_BUILD_DIR}/Scripts/travis/ZcashLightClientSample_setup.sh
-  - curl https://z.cash/downloads/sapling-output.params > ${TRAVIS_BUILD_DIR}/ZcashLightClientKitTests/sapling-output.params
-  - curl https://z.cash/downloads/sapling-output.params > ${TRAVIS_BUILD_DIR}/ZcashLightClientKitTests/sapling-spend.params
-script:
-  - swiftlint
-  - travis_wait 60 xcodebuild -quiet -UseModernBuildSystem=NO -workspace ./Example/ZcashLightClientSample/ZcashLightClientSample.xcworkspace -scheme ZcashLightClientSample -destination platform\=iOS\ Simulator,OS\=14.5,name\=iPhone\ 8 build
+    - swiftlint
```
ccjernigan, line 11: Note that swiftlint isn't being pinned to a specific version.  It should be mostly OK, unless they add a new check that suddenly starts breaking our builds.  The last time I know this happened was about a year or so ago.  This could be something we consider filing a followup issue to improve.

.github/workflows/swift.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+  build:
+
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
ccjernigan, line 11: Checkout v3 just was released this week, so we can bump this version.

Also we can enhance security of this a bit by using the git SHA instead of the version.  The versions are really git tags, which means they could theoretically be changed.  GitHub has some buried documentation on this: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

It is top of mind, because this is a change I'll be making to the Android repos this coming week.

ccjernigan: https://github.com/actions/checkout/releases/tag/v3.0.0 so I think the SHA is `a12a3943b4bdde767164f792f33f40b04645d846 `

ccjernigan:
```suggestion
    - uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
       timeout-minutes: 1
```

.github/workflows/swift.yml
```diff
@@ -0,0 +1,15 @@
...
+    - name: Build
+      run: swift build -v
```
ccjernigan, line 13: Optional, although it might be worth considering an explicit timeout on the build steps.  If we get a bad runner or GitHub Actions is having an incident, sometimes builds can take forever and waste CI minutes.  Setting an explicit timeout creates some guardrails to keep the steps from running for too long.

The downside is that over time, we occasionally might have to increase the timeout as our codebase grows and builds take longer.

I chose the current suggested timeouts based on the run time of the action, plus some buffer.

```suggestion
    - name: Build
      timeout-minutes: 10
      run: swift build -v
```

.github/workflows/swift.yml
```diff
@@ -0,0 +1,15 @@
...
+    - name: Run tests
+      run: swift test --filter OfflineTests
```
ccjernigan, line 15:
```suggestion
    - name: Run tests
      timeout-minutes: 5
      run: swift test --filter OfflineTests
```

changelog.md
```diff
@@ -1,3 +1,20 @@
+# 0.13.0-beta.1
+- [Enhancement] PR #338. Rust-less build. Check for new documentation on how to benefit from this huge change
+- [Enhancement] Swift Package Manager Support!
```
ccjernigan, line 3: 🥳

ccjernigan (changes requested):
It's super exciting to see the new release coming with SPM support.

Everything looks fine—this should work as-is.  I have some comments to improve performance and security of the github actions build though.

.github/workflows/swift.yml
```diff
@@ -0,0 +1,15 @@
+name: Swift
+
+on: [push]
+
+jobs:
+  build:
```
ccjernigan, line 6: We can also harden the job a bit, by preventing the child steps from having write access to the repository.

    permissions:
      contents: read
```suggestion
  build:
      permissions:
          contents: read
```

.travis.yml
pacu: Oh, I removed this and I'm installing swiftlint via homebrew and using the default xcode setting

.github/workflows/swift.yml
ccjernigan: There's a minor syntax error that's failing the build.  Need `permissions:` e.g.
```
  build:
    permissions:
      contents: read
```

.travis.yml
ccjernigan: Maybe I don't understand the Travis syntax.  I currently read this as using Homebrew to install sorcery and swiftlint.  Does Travis look up something in the Xcode project when invoking Homebrew to ensure a specific version of SwiftLint is used??

Or are you saying that because the .swiftlint configuration file is set, that it shouldn't introduce a breaking change?

The problem I encountered about a year ago was that I was using Apple's `masterClock` API for AVFoundation, and SwiftLint added a new check for the word "master" and then it broke my builds until I could update the .swiftlint file to bypass this check.

From what I understand, I don't think the current .swiftlint rules would prevent similar breaking changes in the future.

A solution to that would be to somehow pin a specific SwiftLint version, but I think we can file a followup issue since I don't expect this type of breakage to happen frequently.

ccjernigan (approved):

.travis.yml
pacu: image
Yes, you are right.  This specific target Trivis is building does run swiftlint on a build phase. our usage of swift lint is not strict so it just warns you. 

If the version of homebrew is different (probable because the project one is outdated, since we use the version available on homebrew) it won't run but in terms of failure of success, it doesn't matter in this case.

ccjernigan (approved):

PR #393, opened 2022-06-10
[#384] Adopt Type Safe Memos in the FFI and SDK

### LINKED ISSUE #384: Adopt Type Safe Memos in the FFI and SDK

pacu:
Closes #384

This makes the SDK use the Memo and MemoBytes types across the FFI and the
public interface of the SDKSynchronizer. Addresses the comments by @defuse  on
issue https://github.com/zcash/ZcashLightClientKit/issues/378.
These changes depend on zcash-light-client-ffi repo changes managing
memos as byte arrays instead of string encoding of memos.

There are some compromises in terms of database representation of memos
and their exposure in transaction representing entities because we
wanted to avoid changing the database at this moment and be retro-
compatible with earlier versions and maintain deprecated APIs for wallets
to have time to adopt.

Memo and MemoBytes are WIP and subject to change. we are exploring
making Memo an opaque type based on a struct and drop the enum approach
to make the interface less throwing.

Fix: wrap new functions and deprecate old API. catch Memo errors

update get_sent and get_received memo functions from rust welding
point to branch in FFI repo

# Author

# Reviewer


Package.resolved
```diff
@@ -84,11 +84,11 @@
       },
       {
         "package": "libzcashlc",
-        "repositoryURL": "https://github.com/zcash-hackworks/zcash-light-client-ffi.git",
+        "repositoryURL": "https://github.com/zcash-hackworks/zcash-light-client-ffi",
         "state": {
-          "branch": null,
-          "revision": "b7e8a2abab84c44046b4afe4ee4522a0fa2fcc7f",
-          "version": "0.0.3"
+          "branch": "make-memos-great-again",
```
ccjernigan, line 89: 😂

Tests/DarksideTests/AdvancedReOrgTests.swift
```diff
@@ -303,7 +303,7 @@ class AdvancedReOrgTests: XCTestCase {
             spendingKey: coordinator.spendingKeys!.first!,
             zatoshi: sendAmount,
             toAddress: testRecipientAddress,
-            memo: "test transaction",
+            memo: try! Memo(string: "test transaction"),
```
ccjernigan, line 306: Why do two of these test cases use `try!` while the rest do not?

Tests/DarksideTests/BalanceTests.swift
```diff
@@ -892,7 +892,12 @@ class BalanceTests: XCTestCase {
                 */
                 XCTAssertEqual(Int64(confirmedTx.value), self.sendAmount)
                 XCTAssertEqual(confirmedTx.toAddress, self.testRecipientAddress)
-                XCTAssertEqual(confirmedTx.memo?.asZcashTransactionMemo(), memo)
+                do {
+                    let confirmedMemo = try confirmedTx.memo.intoMemoBytes().intoMemo()
+                    XCTAssertEqual(confirmedMemo, memo)
+                } catch {
+                    XCTFail("failed retrieving memo from confirmed transaction. Errod: \(error.localizedDescription)")
```
ccjernigan, line 899: Errord -> Error

```suggestion
                    XCTFail("failed retrieving memo from confirmed transaction. Error: \(error.localizedDescription)")
```

Sources/ZcashLightClientKit/DAO/PendingTransactionDao.swift
```diff
@@ -57,7 +58,7 @@ struct PendingTransaction: PendingTransactionEntity, Decodable, Encodable {
             raw: entity.raw,
             id: entity.id,
             value: entity.value,
-            memo: entity.memo,
+            memo: entity.memo == nil ? Data(MemoBytes.empty().bytes) : entity.memo,
```
ccjernigan, line 61: What is the type of `entity.memo`?  Are we able to leverage the Optional extensions we have defined here?

ccjernigan (commented):
I know a bit less about the memo and Rust internals.  Overall I think this looks like a good set of enhancements.  I had a few minor comments inline.

Tests/OfflineTests/PendingTransactionRepositoryTests.swift
```diff
@@ -174,6 +174,6 @@ class PendingTransactionRepositoryTests: XCTestCase {
     }
     
     private func mockTransaction() -> PendingTransactionEntity {
-        PendingTransaction(value: Int.random(in: 1 ... 1_000_000), toAddress: recipientAddress, memo: nil, account: 0)
+        PendingTransaction(value: Int.random(in: 1 ... 1_000_000), toAddress: recipientAddress, memo: .empty(), account: 0)
```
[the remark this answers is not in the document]
pacu: Yes, I don't recall why I randomized this but, I doubt it can't be removed. Good catch. thanks!

Sources/ZcashLightClientKit/DAO/PendingTransactionDao.swift
pacu: not actually. this type is Data. Is kept that way so I don't have to change the underlying code :/

pacu: Unfortunately this is of type `Data`. I didn't want to change the underlying encoding so I didn't touch anything of the persistence layer since we will probably make more changes in the next sprints

Tests/DarksideTests/AdvancedReOrgTests.swift
pacu: good catch! thanks

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -297,25 +297,72 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
+    ) -> Memo? {
+        let dbData = dbData.osStr()
+
+        guard let rawMemo = zcashlc_get_received_memo(dbData.0, dbData.1, idNote, networkType.networkId) else { return nil }
+
+        let byteArray = Array(UnsafeMutableBufferPointer(start: rawMemo, count: MemoBytes.capacity))
+
+        defer {
+            zcashlc_byte_array_free(rawMemo, UInt(MemoBytes.capacity))
+        }
```
str4d, line 361: Ditto.

str4d (commented):
I haven't reviewed the entire PR, just the pieces that interact with https://github.com/zcash-hackworks/zcash-light-client-ffi/pull/27.

Sources/ZcashLightClientKit/Model/Memo.swift
```diff
@@ -187,3 +187,25 @@ extension String {
         self = s
     }
 }
```
defuse, line 189: There are a couple issues in `unpaddedRawBytes` above as well:

1. `firstNullByte` should be called `lastNullByte`.
2. I think this leaves one null byte in the string, because `...` is inclusive, when memo strings shouldn't ever end in a null byte (that's interpreted as padding and stripped away)

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -297,25 +297,72 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
     }
+
+    static func getReceivedMemo(
+        dbData: URL,
+        idNote: Int64,
+        networkType: NetworkType
+    ) -> Memo? {
+        let dbData = dbData.osStr()
+
+        guard let rawMemo = zcashlc_get_received_memo(dbData.0, dbData.1, idNote, networkType.networkId) else { return nil }
```
defuse, line 355: This code no longer works with the code in https://github.com/zcash-hackworks/zcash-light-client-ffi/pull/27/files#diff-f64796ccc388f201986ea7eaaf07bbefbd8ac1a56bb995fb505da80879872a39R862 (I think it's because str4d's suggestion was implemented there, but this code hasn't been updated yet)

defuse (changes requested):

Sources/ZcashLightClientKit/Model/Memo.swift
defuse: That function might not handle the completely full memo case (no null bytes) properly. It seems like it would set `firstNullByte` to 512 and then try to `return [UInt8](bytes[0 ... firstNullByte])` which would try to access out-of-bounds memory.

I'm also confused what the `else { return [UInt8](bytes[0 ... 1]) }` is for, why do we want to return the first two bytes in that case? Or am I misunderstanding?

defuse:
Requested some bug fixes, overall this is a great improvement!

LukasKorba (approved):
same as Carter, I'm not so familiar with the layer but overall it LGTM. The try! has been already caught by Carter

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -297,25 +297,72 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
+    ) -> Memo? {
+        let dbData = dbData.osStr()
+
+        guard let rawMemo = zcashlc_get_sent_memo(dbData.0, dbData.1, idNote, networkType.networkId) else { return nil }
+
+        let byteArray = Array(UnsafeMutableBufferPointer(start: rawMemo, count: MemoBytes.capacity))
+
+        defer {
+            zcashlc_byte_array_free(rawMemo, UInt(MemoBytes.capacity))
+        }
```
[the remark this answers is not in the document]
pacu: Good catch! Solved!

I'm creating a contiguous byte array of `MemoBytes.capacity` size and passing a mutable pointer to the rust function, then we copy those bytes (if successful) to the `MemoBytes` struct

Sources/ZcashLightClientKit/Model/Memo.swift
```diff
@@ -187,3 +187,25 @@ extension String {
         self = s
```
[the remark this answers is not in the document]
pacu: Uhmm. Pretty late to the party, I found this block form PSPDFKit about Swift and Null on character on Strings.
https://pspdfkit.com/blog/2020/null-character-strings/

If I understand correctly, we shouldn't be worried about a text string like `let string = "A\u{0}BC"` being passed to this initializer, but we should care about trimming trailing, null-bytes. Is that correct?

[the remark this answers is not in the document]
pacu: @defuse Is this what you are referring to with this?

>Memo strings are still not allowed to contain null bytes at the end of the string, since those are considered to be padding and stripped away in the parsing. Instead of checking for null bytes anywhere, we should now be checking for null bytes at the end.

````Swift
func testTrimTrailingNullBytes() throws {
        let nullTrailedString = "This Is a memo with text and trailing null bytes\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}\u{0}"

        let nonNullTrailedString = "This Is a memo with text and trailing null bytes"

        let trimmedString = String(nullTrailedString.reversed().drop(while: { $0 == "\u{0}"}).reversed())

        XCTAssertEqual(trimmedString, nonNullTrailedString)
    }
````

Shall null-trailed Strings be "sanitized" or should the constructor _throw_ an exception if trailing null-bytes are found?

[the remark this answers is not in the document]
Comment (defuse) [reply Sources/ZcashLightClientKit/Model/Memo.swift:187]: > If I understand correctly, we shouldn't be worried about a text string like let string = "A\u{0}BC" being passed to this initializer, but we should care about trimming trailing, null-bytes. Is that correct?

Right.

With the way the API worked previously, we needed the...

```
        guard !string.containsCStringNullBytesBeforeStringEnding() else {
            throw MemoBytes.Errors.invalidUTF8
        }
...

...code to make sure that the memo won't be truncated. When we were using C strings to pass memos to the rust layer, any null byte would truncate the string. Now, with these changes, we don't have that problem, so we can delete this code.

Now, a memo is allowed to contain null bytes inside the string, but isn't allowed to contain null bytes at the end of the string, because technically those will be considered to be padding and will get cut off by the padding-stripper. So we need to check that:

```
let trimmedString = String(string.reversed().drop(while: { $0 == "\u{0}"}).reversed())
guard trimmedString == string else {
    throw MemoBytes.Errors.endsWithNullBytes
}
```

This will make sure that if someone tries to send a memo with trailing null bytes, it will fail for the sender, instead of it appearing to work and the trailing null bytes getting dropped on the receiver's side.

Now we need something like

Sources/ZcashLightClientKit/Model/Memo.swift
pacu: Thanks for pointing this out!

I made changes to avoid this and also added tests.

This function is kind of odd but works like this: 
1.  enumerate the byte array in an (index, byte) pair
2. reverse the enumerated sequence
3. get the index of the first non-null byte in the pair of the reversed enumerated sequence
4. return an array of the slice resulting of index 0 to the found index. Else return the first byte.

Sources/ZcashLightClientKit/DAO/PendingTransactionDao.swift
```diff
@@ -105,7 +105,7 @@ class PendingTransactionSQLDAO: PendingTransactionRepository {
         static var raw = Expression<Blob?>("raw")
         static var id = Expression<Int>("id")
         static var value = Expression<Int>("value")
-        static var memo = Expression<Blob?>("memo")
+        static var memo = Expression<Blob>("memo")
```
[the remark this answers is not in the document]
pacu: thanks! I should have reverted this!

pacu:
https://github.com/actions/virtual-environments/issues/5836 
awaiting for xcode 13.4.1 on github actions

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -458,35 +458,12 @@ public class SDKSynchronizer: Synchronizer {
     
     // MARK: Synchronizer methods
 
-    @available(*, deprecated, message: "This function will be removed soon, use the one reveiving a `Zatoshi` value instead")
-    public func sendToAddress(
```
str4d, line 462: I'm assuming this was removed because it was left there to avoid changing the argument types, but now another argument is forced to change type so there's no benefit to further deprecation over removal.

Sources/ZcashLightClientKit/Model/Memo.swift
```diff
@@ -74,7 +76,11 @@ public struct MemoText: Equatable {
 
 public struct MemoBytes: Equatable {
     public enum Errors: Error {
+        /// Invalid UTF-8 Bytes where detected when attempting to create a Text Memo
```
str4d, line 79:
```suggestion
        /// Invalid UTF-8 Bytes were detected when attempting to create a Text Memo
```

Sources/ZcashLightClientKit/Transaction/PersistentTransactionManager.swift
```diff
@@ -94,6 +95,14 @@ class PersistentTransactionManager: OutboundTransactionManager {
...
                     result(.failure(TransactionManagerError.updateFailed(pendingTransaction)))
                 }
+            } catch MemoBytes.Errors.invalidUTF8 {
+                DispatchQueue.main.async {
+                    result(.failure(TransactionManagerError.shieldingEncodingFailed(pendingTransaction, reason: "Memo contains invalid UTF-8 bytes")))
+                }
+            } catch MemoBytes.Errors.tooLong(let length) {
+                DispatchQueue.main.async {
+                    result(.failure(TransactionManagerError.shieldingEncodingFailed(pendingTransaction, reason: "Memo is too long. expected 512 bytes, received \(length)")))
+                }
```
str4d, line 105: Does `MemoBytes.Errors.endsWithNullBytes` need to be special-cased here as well, or is it okay for that error case to fall through to the default catch?

Sources/ZcashLightClientKit/DAO/PendingTransactionDao.swift
```diff
@@ -7,7 +7,8 @@
 
 import Foundation
 import SQLite
-struct PendingTransaction: PendingTransactionEntity, Codable {
+struct PendingTransaction: PendingTransactionEntity, Decodable, Encodable {
```
str4d, line 10: I don't know what this change is doing; it doesn't seem like it should affect memo usage.

Sources/ZcashLightClientKit/DAO/PendingTransactionDao.swift
```diff
@@ -57,7 +58,7 @@ struct PendingTransaction: PendingTransactionEntity, Codable {
             raw: entity.raw,
             id: entity.id,
             value: entity.value,
-            memo: entity.memo,
+            memo: entity.memo == nil ? Data(MemoBytes.empty().bytes) : entity.memo,
```
str4d, line 61: Elsewhere, I thought that optional `Data` was mapped to `MemoBytes.empty()` if not present. Does that mean `PendingTransaction.memo` can represent an empty memo as either `null` or `Data(MemoBytes.empty()`, or is `null` used for something else? I don't know if it matters that an empty memo can be represented in two ways within this part of the SDK.

Tests/DarksideTests/BalanceTests.swift
```diff
@@ -83,7 +83,7 @@ class BalanceTests: XCTestCase {
             spendingKey: spendingKey,
             zatoshi: maxBalance,
             toAddress: try Recipient(testRecipientAddress, network: self.network.networkType),
-            memo: "test send \(self.description) \(Date().description)",
+            memo: try Memo(string: "this is a test"),
```
str4d, line 86: Is it intended that this memo stops having field substitutions?

Tests/DarksideTests/BalanceTests.swift
```diff
@@ -511,7 +511,7 @@ class BalanceTests: XCTestCase {
             spendingKey: spendingKey,
             zatoshi: sendAmount,
             toAddress: try Recipient(testRecipientAddress, network: self.network.networkType),
-            memo: "test send \(self.description) \(Date().description)",
+            memo: try Memo(string: "this is a test"),
```
str4d, line 514: Same question here.

str4d (approved):
utACK 0282d81584888009127174a49e196a6c6ec2f502 with a few questions.

PR #445, opened 2022-07-29
[#444] Syncing Restarts to zero when the wallet is wiped and synced f…

### LINKED ISSUE #444: Syncing Restarts to zero when the wallet is wiped and synced from zero in one go.
precondition: this only happens if the birthday provided does not match the one on disk an a lower height is used.
steps:

1. Wipe wallet.
2. let it sync 💤 (this takes a while)
3. watch for when finishes downloading.
4. validation fails at checkpoint height. 
5. wallet rewinds to checkpoint height and  wallet restarts downloading

pacu:
…rom zero in one go.

This commit changes the way walletBirthday is stored in the synchronizer and intinitializer. Wallets syncing from creation/restore would have a problem where the birthday stored in the Blocks Table would be the one corresponding to the chekpoint found a the time of syncing, but the compact block downloader would start downloading blocks from the height provided by the user when the wallet was restored. This would cause a `validationFailed` error at the height checkpoint.height + 1 and restart downloading from checkpoint.height and then resume correctly.

Darksidewalletd test setUp was changed re be able to reproduce the error since injection guaranteed correctness and it was not possible to reproduce the error with it since the wallet birthday height provided matched exactly with the checkpoint on disk and that's the only case that avoided this error.

Closes #444

# Author

# Reviewer

### DIFF
--- a/Tests/DarksideTests/AdvancedReOrgTests.swift
+++ b/Tests/DarksideTests/AdvancedReOrgTests.swift
@@ -35,7 +35,7 @@ class AdvancedReOrgTests: XCTestCase {
         try super.setUpWithError()
         coordinator = try TestCoordinator(
             seed: seedPhrase,
-            walletBirthday: birthday,
+            walletBirthday: birthday + 50, //don't use an exact birthday, users never do.
             channelProvider: ChannelProvider(),
             network: network
         )
--- a/Sources/ZcashLightClientKit/Initializer.swift
+++ b/Sources/ZcashLightClientKit/Initializer.swift
@@ -78,6 +78,8 @@ public class Initializer {
     private(set) var downloader: CompactBlockDownloader
     private(set) var network: ZcashNetwork
     private(set) public var viewingKeys: [UnifiedViewingKey]
+    /// The effective birthday of the wallet based on the height provided when initializing
+    /// and the checkpoints available on this SDK
     private(set) public var walletBirthday: BlockHeight
 
     /**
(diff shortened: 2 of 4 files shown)


Tests/DarksideTests/AdvancedReOrgTests.swift
```diff
@@ -35,7 +35,7 @@ class AdvancedReOrgTests: XCTestCase {
         try super.setUpWithError()
         coordinator = try TestCoordinator(
             seed: seedPhrase,
-            walletBirthday: birthday,
+            walletBirthday: birthday + 50, //don't use an exact birthday, users never do.
```
ccjernigan, line 38: Non blocking, since this is a test case.

But should this be plus or should it be minus?  Typically we'd expect a birthday behind, not ahead, right?

ccjernigan (approved):

Tests/DarksideTests/AdvancedReOrgTests.swift
pacu: Well the thing is that these tests use a synthetic blockchain that starts at 663150.  There's nothing before that height. 
users using a birthday that doesn't match a checkpoint will end up using a checkpoint that's lower than their input.

pacu: I don't think so. because that already happens when selecting the effective birthday with the checkpoint height. 
We should review the way ECC Wallet android and Secant wallets  proppose the birthday of restored wallets without a height. I recall that the android reference wallet suggested a "better. birthday" when the first transaction is found", if the suggested height is the mined height of the transaction and it matches an existing breakpoint of the wallet it will not catch the first transaction. 

That's what we should check.

pacu: If I have a transaction on block 850000 and there's a checkpoint for that height and one before at 845000, the latter should be chosen. I don't know if that is what we are currently doing

PR #500, opened 2022-08-24
463 migrate lightwallet service to async await

pacu:
Draft: PoC on how to return cancellable AsyncStream 

https://www.swiftbysundell.com/articles/async-sequences-streams-and-combine/

### DIFF
--- a/Sources/ZcashLightClientKit/Service/LightWalletGRPCService.swift
+++ b/Sources/ZcashLightClientKit/Service/LightWalletGRPCService.swift
@@ -301,6 +301,7 @@ extension LightWalletGRPCService: LightWalletServiceBlockingAPI {
 // MARK: - LightWalletServiceNonBlockingAPI
 
 extension LightWalletGRPCService: LightWalletServiceNonBlockingAPI {
+
     public func getInfo(result: @escaping (Result<LightWalletdInfo, LightWalletServiceError>) -> Void) {
         compactTxStreamer.getLightdInfo(Empty()).response.whenComplete { completionResult in
             switch completionResult {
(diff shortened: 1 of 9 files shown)


Sources/ZcashLightClientKit/Service/LightWalletGRPCService.swift
```diff
@@ -454,30 +455,38 @@ extension LightWalletGRPCService: LightWalletServiceNonBlockingAPI {
         }
     }
     
-    public func fetchUTXOsAsync(for tAddress: String, height: BlockHeight) async throws -> [UnspentTransactionOutputEntity] {
+    public func fetchUTXOsAsync(for tAddress: String, height: BlockHeight) async throws -> AsyncThrowingStream<UnspentTransactionOutputEntity, Error> {
```
LukasKorba, line 458: I think this new change is actually mixing 2 approaches. It should either be 
```swift
public func fetchUTXOsAsync(for tAddress: String, height: BlockHeight) async throws -> [UnspentTransactionOutputEntity]  
```
(which is the previous implementation) OR
```swift
public func fetchUTXOsAsync(for tAddress: String, height: BlockHeight) -> AsyncThrowingStream<UnspentTransactionOutputEntity, Error> 
```
(removed async throws)

Both do the same job. The 1st one simply gets all UnspentTransactionOutputEntity in async context, caller is just waiting for the result and for loop is hidden inside the method. 2nd one returns AsyncThrowingStream and let caller to for loop it. 

`async throws` does nothing in this context in the 2nd method.

Sources/ZcashLightClientKit/Service/LightWalletGRPCService.swift
```diff
@@ -454,30 +455,38 @@ extension LightWalletGRPCService: LightWalletServiceNonBlockingAPI {
...
+                            script: reply.script,
+                            valueZat: Int(reply.valueZat),
+                            height: Int(reply.height),
+                            spentInTx: nil
+                        )
+                    )
+                }
+
+                continuation.finish()
+            }
```
LukasKorba, line 488: This continuation is actually never throwing an error. The for loop should be wrapped in do catch or return value `AsyncThrowingStream<UnspentTransactionOutputEntity, Error>` should be changed to `AsyncThrowingStream<UnspentTransactionOutputEntity, Never>` I guess.

LukasKorba, line 488: I think it should be like this:

```swift
return AsyncThrowingStream { continuation in
            Task {
                continuation.onTermination = { _ in
                    utxoStreamCall.cancel()
                }
                
                do {
                    for try await reply in utxoStreamCall.responseStream {
                        continuation.yield(
                            UTXO(
                                id: nil,
                                address: tAddress,
                                prevoutTxId: reply.txid,
                                prevoutIndex: Int(reply.index),
                                script: reply.script,
                                valueZat: Int(reply.valueZat),
                                height: Int(reply.height),
                                spentInTx: nil
                            )
                        )
                    }
                    continuation.finish(throwing: nil)
                } catch {
                    continuation.finish(throwing: error)
                }
            }
        }
```

Sources/ZcashLightClientKit/Service/LightWalletGRPCService.swift
```diff
@@ -454,30 +455,38 @@ extension LightWalletGRPCService: LightWalletServiceNonBlockingAPI {
...
+
+        let utxoStreamCall = compactTxStreamerAsync.makeGetAddressUtxosStreamCall(arg)
+
+        return AsyncThrowingStream { continuation in
+            Task {
+                continuation.onTermination = { @Sendable _ in
+                    utxoStreamCall.cancel()
+                }
+
+                for try await reply in utxoStreamCall.responseStream {
```
LukasKorba, line 472: Are we sure the the error in case on `try` will be propagated?

LukasKorba, line 472: Also this gRPC API and its termination might be probably replaced by
```swift
let stream = compactTxStreamerAsync.getAddressUtxosStream(arg)

        return AsyncThrowingStream { continuation in
            Task {
                do {
                    for try await reply in stream {
```

LukasKorba: UPD: I tried it and without do-catch the error is not propagated

PR #535, opened 2022-09-19
Use UnifiedSpendingKey for shielding and Spending

### LINKED ISSUE #534: Use UnifiedSpendingKey for shielding and Spending

pacu:
Closes #534 

image

# Author

- [x] Documentation: Did you update Docs as appropiate? (E.g [README.md](../blob/main/README.md), etc.)

# Reviewer


Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -684,21 +687,24 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
+        defer { zcashlc_string_free(zaddrCStr) }
+
+        guard let uAddr = String(validatingUTF8: zaddrCStr) else {
+            throw RustWeldingError.unableToDeriveKeys
+        }
+
+        return UnifiedAddress(validatedEncoding: uAddr)
     }
     
     static func deriveTransparentAddressFromSeed(
```
nuttycom, line 702: Same with this?

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -711,19 +717,23 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
+        defer { zcashlc_string_free(tAddrCStr)}
+
+        guard let tAddr = String(validatingUTF8: tAddrCStr) else {
+            throw RustWeldingError.unableToDeriveKeys
+        }
+
+        return TransparentAddress(validatedEncoding: tAddr)
     }
     
     static func deriveTransparentAccountPrivateKeyFromSeed(
```
nuttycom, line 732: And this.

Sources/ZcashLightClientKit/Tool/DerivationTool.swift
```diff
@@ -108,125 +61,39 @@ public class DerivationTool: KeyDeriving {
...
-        }
-        do {
-            return try rustwelding.deriveUnifiedFullViewingKeyFromSeed(seed, numberOfAccounts: numberOfAccounts, networkType: networkType)
-        } catch {
-            throw KeyDerivationErrors.derivationError(underlyingError: error)
-        }
+    /// Given a spending key, return the associated viewing key.
+    /// - Parameter spendingKey: the `UnifiedSpendingKey` from which to derive the `UnifiedFullViewingKey` from.
+    /// - Returns: the viewing key that corresponds to the spending key.
+    public func deriveUnifierFullViewingKey(from spendingKey: UnifiedSpendingKey) throws -> UnifiedFullViewingKey {
```
nuttycom, line 75:
```suggestion
    public func deriveUnifiedFullViewingKey(from spendingKey: UnifiedSpendingKey) throws -> UnifiedFullViewingKey {
```

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
 }
 
 protocol ZcashRustBackendWelding {
-    /**
-    gets the latest error if available. Clear the existing error
-    */
+    /// clears the cached utxos for the given address from the specified height on
```
nuttycom, line 35: Can you add a little bit of documentation about how/when this is supposed to be used?

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+    /// [`UnifiedSpendingKey`] for the newly created account.  The caller should manage the memory of
+    /// (and store) the returned spending keys in a secure fashion.
+    ///
+    /// If `seed` was imported from a backup and this method is being used to restore a
+    /// previous wallet state, you should use this method to add all of the desired
+    /// accounts before scanning the chain from the seed's birthday height.
+    ///
+    /// By convention, wallets should only allow a new account to be generated after funds
+    /// have been received by the currently-available account (in order to enable
+    /// automated account recovery).table from a given seed and a number of accounts
```
nuttycom, line 56:
```suggestion
    /// automated account recovery).
```

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+    /// - Parameters:
+    ///     - dbData: location of the data db
+    ///     - seed: byte array of the zip32 seed
+    ///  - Returns: The `UnifiedSpendingKey` structs for the number of accounts created
```
nuttycom, line 60: Missing `networkType`

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+    ///
+    static func createAccount(dbData: URL, seed: [UInt8], networkType: NetworkType) throws -> UnifiedSpendingKey
+
+    /// Creates a transaction to the given address from the given account
+    /// - Parameter dbData: URL for the Data DB
+    /// - Parameter usk: `UnifiedSpendingKey`y for the account the spend will be done from
```
nuttycom, line 66:
```suggestion
    /// - Parameter usk: `UnifiedSpendingKey` for the account that controls the funds to be spent.
```

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+    /// - Parameter to: recipient address
+    /// - Parameter value: transaction amount in Zatoshi
+    /// - Parameter memo: the `Memo` for this transaction
+    /// - Parameter spendParamsPath: path escaped String for the filesystem locations where the spend parameters are located
+    /// - Parameter outputParamsPath: path escaped String for the filesystem locations where the output parameters are located
```
nuttycom, line 71: An API question: it seems like `dbData`, `spendParamsPath`, and `outputParamsPath` (and maybe `networkType`) could all be member variables of an object, and then `createToAddress` could just be a method that takes the USK, the address, the value and the memo. 

I like to separate things that are configuration-oriented from business logic in my applications; I'm not sure if that applies here.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+    ///   - networkType: network type of this key
+    static func getCurrentAddress(
+        dbData: URL,
+        account: Int32,
+        networkType: NetworkType
+    ) throws -> UnifiedAddress
+
+    /// Returns the nearest height where a rewind is possible. Currently prunning gets rid of sapling witnesses older
+    /// than 100 blocks. So in order to reconstruct the witness tree that allows to spend notes from the given wallet
+    /// the rewind can't be more than 100 block or back to the oldest unspent note that this wallet contains.
```
nuttycom, line 134:
```suggestion
    /// the rewind can't be more than 100 blocks or back to the oldest unspent note that this wallet contains.
```

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+    ///   - dbData: location of the data db
+    ///   - account: index of the given account
+    ///   - networkType: network type of this key
+    static func getCurrentAddress(
+        dbData: URL,
+        account: Int32,
+        networkType: NetworkType
+    ) throws -> UnifiedAddress
+
+    /// Returns the nearest height where a rewind is possible. Currently prunning gets rid of sapling witnesses older
```
nuttycom, line 132:
```suggestion
    /// Returns the nearest height where a rewind is possible. Currently pruning gets rid of sapling witnesses older
```

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+    /// than 100 blocks. So in order to reconstruct the witness tree that allows to spend notes from the given wallet
+    /// the rewind can't be more than 100 block or back to the oldest unspent note that this wallet contains.
+    /// - Parameters:
+    ///   - dbData: location of the data db file
+    ///   - height: height you would like to rewind to.
+    static func getNearestRewindHeight(dbData: URL, height: Int32, networkType: NetworkType) -> Int32
```
nuttycom, line 138: It's not entirely clear to me from the documentation how this is intended to be used. Is this basically just so that one can guard a rewind operation? If so, should we alternately consider and API that simply attempts a rewind, and does this check internally (and returns a failure if the requested rewind can't be performed?)

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+        networkType: NetworkType
+    ) -> String?
+
+    /// get sent memo from note
+    /// - Parameters:
+    ///   - dbData: location of the data db file
+    ///   - idNote: note_id of note where the memo is located
+    ///   - networkType: network type of this key
+    /// - Returns: a `Memo` if any
+    static func getSentMemo(
```
nuttycom, line 195: Should we consider unifying the `getSentMemo` and `getReceivedMemo` APIs? It seems like this is something that should be handled on the backend.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+        dbData: URL,
+        address: String,
+        networkType: NetworkType
+    ) throws -> Int64
+
+    /// Returns the transparent receiver within the given Unified Address, if any.
+    // - Parameter uAddr: a `UnifiedAddress`
+    /// - Returns a `TransparentAddress` if any
+    /// - Throws `receiverNotFound` when the receiver is not found. `invalidUnifiedAddress` if the UA provided is not valid
+    static func getTransparentReceiver(for uAddr: UnifiedAddress) throws -> TransparentAddress
```
nuttycom, line 216: Should this instead return a `TransparentAddress?`, and only throw if the UA is invalid? That seems like the more appropriate way to separate concerns; it's not an error that a UA doesn't contain a transparent receiver.

Nit: I would group this with the `getSaplingReceiver` function.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+        dbData: URL,
+        idNote: Int64,
+        networkType: NetworkType
+    ) -> Memo?
+
+    /// Returns the Sapling receiver within the given Unified Address, if any.
+    /// - Parameter uAddr: a `UnifiedAddress`
+    /// - Returns a `SaplingAddress` if any
+    /// - Throws `receiverNotFound` when the receiver is not found. `invalidUnifiedAddress` if the UA provided is not valid
+    static func getSaplingReceiver(for uAddr: UnifiedAddress) throws -> SaplingAddress
```
nuttycom, line 175: I think that this should return `SaplingAddress?` instead. It's not an error to not have a Sapling receiver on a UA.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
-    gets the latest error message from librustzcash. Does not clear existing error
-    */
+    /// gets the latest error message from librustzcash. Does not clear existing error
     static func getLastError() -> String?
 
-    /**
-    initializes the data db
-    - Parameter dbData: location of the data db sql file
-    */
+    /// initialize the accounts table from a set of unified full viewing keys
```
nuttycom, line 224: I would document that is is essentially only intended for situations where you're restoring from seed, and that `createAccount` should be used otherwise.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -32,43 +32,210 @@ public enum DbInitResult {
...
+    /// - Parameter dbData: location of the data db
+    /// - Parameter ufvks: an array of UnifiedFullViewingKeys
+    static func initAccountsTable(dbData: URL, ufvks: [UnifiedFullViewingKey], networkType: NetworkType) throws -> Bool
+
+    /// initializes the data db
```
nuttycom, line 229: Document that this will perform the migrations required to update the data database to have the required schema and contents.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -77,182 +244,96 @@ protocol ZcashRustBackendWelding {
...
-        - saplingTree: hash of the sapling tree
-    */
-    // swiftlint:disable function_parameter_count
+    /// initialize the blocks table from a given checkpoint (birthday)
+    /// - Parameters:
+    ///   - dbData: location of the data db
+    ///   - height: represents the block height of the given checkpoint
+    ///   - hash: hash of the merkle tree
+    ///   - time: in milliseconds from reference
+    ///   - saplingTree: hash of the sapling tree
```
nuttycom, line 267: Missing `networkType`

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -77,182 +244,96 @@ protocol ZcashRustBackendWelding {
...
-    initialize the blocks table from a given checkpoint (birthday)
-    - Parameters:
-        - dbData: location of the data db
-        - height: represents the block height of the given checkpoint
-        - hash: hash of the merkle tree
-        - time: in milliseconds from reference
-        - saplingTree: hash of the sapling tree
-    */
-    // swiftlint:disable function_parameter_count
+    /// initialize the blocks table from a given checkpoint (birthday)
```
nuttycom, line 261: I don't see an argument that can be used to provide the birthday height? Or is `birthday` the wrong term here?

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -77,182 +244,96 @@ protocol ZcashRustBackendWelding {
...
-    */
-    static func getNearestRewindHeight(dbData: URL, height: Int32, networkType: NetworkType) -> Int32
 
-    /**
-    rewinds the compact block storage to the given height. clears up all derived data as well
-    - Parameters:
-        - dbData: location of the data db file
-        - height: height to rewind to. DON'T PASS ARBITRARY HEIGHT. Use getNearestRewindHeight when unsure
-    */
+    /// rewinds the compact block storage to the given height. clears up all derived data as well
```
nuttycom, line 302: This documentation is a little confusing, because I think of the "compact block storage" as the cache db. 
```suggestion
    /// Resets the state of the database to only contain block and transaction information up to the 
    /// specified height.
```

nuttycom: I think that this should also document what happens if the user has sent transactions in blocks that would be invalidated by this rewind.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -77,182 +244,96 @@ protocol ZcashRustBackendWelding {
...
+    /// - Parameters:
+    ///   - dbData: location of the data db file
+    ///   - height: height to rewind to. DON'T PASS ARBITRARY HEIGHT. Use getNearestRewindHeight when unsure
```
nuttycom, line 305: As suggested above, I don't think that this should be two calls; instead, the rewind should return failure if too large of a rewind is requested. Maybe the height parameter should be optional, and then if it's not provided the `getNearestRewindHeight` call should be done internally?

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -77,182 +244,96 @@ protocol ZcashRustBackendWelding {
...
+    /// For brand-new light client databases, this function starts scanning from the Sapling
+    /// activation height. This height can be fast-forwarded to a more recent block by calling
+    /// [`zcashlc_init_blocks_table`] before this function.
+    /// Scanned blocks are required to be height-sequential. If a block is missing from the
+    /// cache, an error will be signalled.
+    ///
+    /// - Parameters:
+    ///   - dbCache: location of the compact block cache db
+    ///   - dbData:  location of the data db file
+    ///   - limit: scan up to limit blocks. pass 0 to set no limit.
```
nuttycom, line 324: Missing documentation for the `networkType` parameter.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -77,182 +244,96 @@ protocol ZcashRustBackendWelding {
...
+
+    /// Scans new blocks added to the cache for any transactions received by the tracked
+    /// accounts.
+    /// This function pays attention only to cached blocks with heights greater than the
+    /// highest scanned block in `db_data`. Cached blocks with lower heights are not verified
+    /// against previously-scanned blocks. In particular, this function **assumes** that the
+    /// caller is handling rollbacks.
+    /// For brand-new light client databases, this function starts scanning from the Sapling
+    /// activation height. This height can be fast-forwarded to a more recent block by calling
+    /// [`zcashlc_init_blocks_table`] before this function.
```
nuttycom, line 317: `zcashlc_init_blocks_table` shouldn't be mentioned here explicitly.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -262,176 +343,50 @@ protocol ZcashRustBackendWelding {
...
-        - spendParamsPath: path escaped String for the filesystem locations where the spend parameters are located
-        - outputParamsPath: path escaped String for the filesystem locations where the output parameters are located
-    */
-    // swiftlint:disable function_parameter_count
+
+    /// Creates a transaction to shield all found UTXOs in cache db.
+    /// - Parameter dbCache: URL for the Cache DB
+    /// - Parameter dbData: URL for the Data DB
+    /// - Parameter account: the account index that will originate the transaction
+    /// - Parameter usk: `UnifiedSpendingKey` that spend transarent funds and where the funds will be shielded to.
```
nuttycom, line 359: Document that this must correspond to the provided account. If your `UnifiedSpendingKey` model contains account information, then the `account` parameter should be omitted at this layer entirely.

nuttycom (changes requested):
Overall, most of these comments are minor. However, I have one more substantive design suggestion: I think that the `ZcashRustBackendWelding` API should be constructed in a more object-oriented fashion, where the methods are focused on the domain operations, and the `dataDb`, various spend parameter paths, and network information all be provided to a fallible function that constructs the object that represents the capability to perform these operations. As a secondary benefit to this approach, the data db initialization call could be performed as part of this constructor function.

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -408,389 +288,382 @@ class ZcashRustBackend: ZcashRustBackendWelding {
         return (try? MemoBytes(contiguousBytes: contiguousMemoBytes)).flatMap { try? $0.intoMemo() }
     }
 
-    static func getReceivedMemo(
+    static func getTransparentBalance(
         dbData: URL,
-        idNote: Int64,
+        address: String,
```
nuttycom, line 293: Does this make sense to go address-by-address? Or should the transparent balance be returned on an account-by-account basis?

I guess that per-address is relevant because we want shielding operations to usually not shield across multiple transparent addresses, because doing so links the associated UAs as belonging to the same wallet.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
pacu: I think these suggestions make sense but are out of scope of what we are trying to achieve at the moment

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
pacu: I think that the balance makes sense on an account basis. But I've never thought of shielding in terms of rolling addresses. Would grabbing all the funds from all addresses and shielding them provide insight on those addresses? OTOH, wouldn't fragmenting per-address eventually make a shielding threshold unreachable?

Sources/ZcashLightClientKit/Tool/DerivationTool.swift
pacu: I need to do a rename all. It's pending. thanks

Package.swift
```diff
@@ -16,7 +16,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/grpc/grpc-swift.git", from: "1.8.0"),
         .package(url: "https://github.com/stephencelis/SQLite.swift.git", from: "0.13.0"),
-        .package(name:"libzcashlc", url: "https://github.com/zcash-hackworks/zcash-light-client-ffi", branch: "bin/librustzcash_0_7"),
+        .package(name:"libzcashlc", url: "https://github.com/zcash-hackworks/zcash-light-client-ffi", branch: "bin/librustzcash_0_7")
```
nuttycom, line 19: Would it make sense to use an explicit hash reference, rather than a branch ID? I think that would make things a little easier with Xcode, because then when that hash changes xcode will know it needs to fetch the new version?

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -246,24 +245,24 @@ public class Initializer {
...
-    public func getAddress(index account: Int = 0) -> String? {
-        try? accountRepository.findBy(account: account)?.address
-    }
+//    /**
+//    get address from the given account index
+//    - Parameter account:  the index of the account
+//    */
+//    public func getAddress(index account: Int = 0) -> String? {
+//        try? accountRepository.findBy(account: account)?.address
+//    }
```
LukasKorba, line 265: do we want to keep the commented out code here?

LukasKorba (approved):
LGTM.. just a small non-blocking comment, haven't run the tests though

Sources/ZcashLightClientKit/Initializer.swift
pacu: nope! thanks!

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
pacu: Do you mean something like this? (note: I haven't tested this)

````Swift
static func rewindToHeight(
        dbData: URL,
        height: Int32?,
        networkType: NetworkType
    ) throws -> Bool {
        let nearestRewind = getNearestRewindHeight(dbData: dbData, height: 0, networkType: networkType)

        let dbData = dbData.osStr()


        if let height = height {
            guard height < nearestRewind else {
                throw RustWeldingError.invalidRewind(suggestedHeight: nearestRewind)
            }

            return zcashlc_rewind_to_height(dbData.0, dbData.1, height, networkType.networkId)
        } else {
            return zcashlc_rewind_to_height(dbData.0, dbData.1, nearestRewind, networkType.networkId)
        }
    }
````

File: Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift (hunk 24 above)
Comment (pacu) [reply Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift]: Ok I implemented the account based transparent balance that addresses this

File: Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift (hunk 12 above)
Comment (pacu) [reply Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift:225]: probably out of scope for now

File: Package.swift (hunk 25 above)
Comment (pacu) [reply Package.swift:19]: Xcode hates us 😹

File: Example/ZcashLightClientSample/ZcashLightClientSample/Get Address/GetAddressViewController.swift [hunk 27]
```diff
@@ -11,24 +11,26 @@ import ZcashLightClientKit
 class GetAddressViewController: UIViewController {
     @IBOutlet weak var unifiedAddressLabel: UILabel!
     @IBOutlet weak var tAddressLabel: UILabel!
-    @IBOutlet weak var spendingKeyLabel: UILabel! // THIS SHOULD BE SUPER SECRET!!!!!
-
-    // THIS SHOULD NEVER BE STORED IN MEMORY
-    // swiftlint:disable:next implicitly_unwrapped_optional
-    var spendingKey: SaplingExtendedSpendingKey!
+    @IBOutlet weak var saplingAddress: UILabel! // THIS SHOULD BE SUPER SECRET!!!!!
```
nuttycom, line 14: This comment is no longer valid.

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -408,392 +293,420 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
+        guard zcashlc_is_valid_viewing_key([CChar](key.utf8CString), networkType.networkId) else {
             if let error = lastError() {
                 throw error
             }
-            return nil
+            return false
         }
 
-        defer { zcashlc_string_free(extfvk) }
+        return true
```
nuttycom, line 442: nit: I think that this would read more clearly as:
```suggestion
        let result = zcashlc_is_valid_viewing_key([CChar](key.utf8CString), networkType.networkId)
        if let error = lastError() {
            throw error
        } else {
            return result
        }
```

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -408,392 +293,420 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
-
-        guard
-            let ffiEncodedKeysPtr = zcashlc_derive_extended_full_viewing_keys(
-                seed,
-                UInt(seed.count),
-                accounts,
-                networkType.networkId
-            )
-        else {
+        guard zcashlc_is_valid_sapling_extended_spending_key([CChar](key.utf8CString), networkType.networkId) else {
```
nuttycom, line 450: Same comment here about the use of `guard` when returning boolean values in general.

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -408,392 +293,420 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
-            let itemPointer = ufvksArrayPointer.advanced(by: item)
+        var ffiUfvks = [FFIEncodedKey]()
+        for ufvk in ufvks {
+            guard !ufvk.encoding.containsCStringNullBytesBeforeStringEnding() else {
+                throw RustWeldingError.invalidInput(message: "`UFVK` contains null bytes.")
+            }
             
-            guard let encoding = String(validatingUTF8: itemPointer.pointee.encoding) else {
-                throw RustWeldingError.unableToDeriveKeys
+            guard try self.isValidUnifiedFullViewingKey(ufvk.encoding, networkType: networkType) else { // TODO Fix
```
nuttycom, line 502: What is this TODO about?

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -408,392 +293,420 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
+            [CChar](hash.utf8CString),
+            time,
+            [CChar](saplingTree.utf8CString),
+            networkType.networkId
+        ) != 0 else {
             if let error = lastError() {
                 throw error
             }
-            return nil
+            throw RustWeldingError.dataDbInitFailed(message: "Unknown Error")
```
nuttycom, line 570: This seems like a programming error rather than an initialization failure, isn't it? Should this be an assertion instead? Is `lastError` expected to be fallible in such a way that it requires this fallback?

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -16,243 +16,430 @@ enum RustWeldingError: Error {
...
+    /// needing to put them back again to restore the balance (if they weren't spent)
+    /// - Parameters:
+    ///   - dbData: location of the data db file
+    ///   - address: the address of the UTXO
+    ///   - sinceheight: clear the UXTOs from that address on
+    ///   - networkType: network type of this key
+    /// - Returns: the amount of UTXOs cleared or -1 on error
+    static func clearUtxos(
+        dbData: URL,
+        address: String,
```
nuttycom, line 46: nit: it'd be nice to have a better type for transparent addresses.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -16,243 +16,430 @@ enum RustWeldingError: Error {
...
+    /// - Parameter to: recipient address
+    /// - Parameter value: transaction amount in Zatoshi
+    /// - Parameter memo: the `Memo` for this transaction
+    /// - Parameter spendParamsPath: path escaped String for the filesystem locations where the spend parameters are located
+    /// - Parameter outputParamsPath: path escaped String for the filesystem locations where the output parameters are located
+    /// - Parameter networkType: network type of this key
+    static func createToAddress(
+        dbData: URL,
+        usk: UnifiedSpendingKey,
+        to address: String,
```
nuttycom, line 89: This on the other hand really is a polymorphic address string.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
nuttycom: Yes, that looks good.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -16,243 +16,430 @@ enum RustWeldingError: Error {
...
+        limit: UInt32,
+        networkType: NetworkType
+    ) -> Bool
+
+
+    /// puts a UTXO into the data db database
+    /// - Parameters:
+    ///   - dbData: location of the data db file
+    ///   - txid: the txid bytes for the UTXO
+    ///   - index: the index of the UTXO
```
nuttycom, line 438: The `script` argument is not documented

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -487,18 +484,15 @@ public class SDKSynchronizer: Synchronizer {
...
     ) {
         // let's see if there are funds to shield
-        let derivationTool = DerivationTool(networkType: self.network.networkType)
-
+        let accountIndex = Int(spendingKey.account)
         do {
-            let tAddr = try derivationTool.deriveTransparentAddressFromAccountPrivateKey(transparentAccountPrivateKey, index: 0)
 
-            let tBalance = try utxoRepository.balance(address: tAddr.stringEncoded, latestHeight: self.latestDownloadedHeight())
+            let tBalance = try self.getTransparentBalance(accountIndex: accountIndex)
```
nuttycom, line 495: This is fine for now; we'll need to rework shielding according to ZIP 315 later.

nuttycom (changes requested):
Just a couple more changes, and https://github.com/zcash-hackworks/zcash-test-vectors/pull/90 should have the updated test vectors that you need to make the remainder of the tests work.

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
```diff
@@ -1023,34 +1022,41 @@ extension CompactBlockProcessor {
...
-        return try? SaplingAddress(encoding: zAddress, network: self.config.network.networkType)
+        getUnifiedAddres(accountIndex: accountIndex)?.saplingReceiver()
     }
     
     public func getTransparentAddress(accountIndex: Int) -> TransparentAddress? {
-        guard let tAddress = try? accountRepository.findBy(account: accountIndex)?.transparentAddress else { return nil }
-
-        return TransparentAddress(validatedEncoding: tAddress)
-
+        getUnifiedAddres(accountIndex: accountIndex)?.transparentReceiver()
```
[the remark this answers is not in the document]
pacu: Oh thanks! I had changed that but apparently I lost that change somehow

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
```diff
@@ -1023,34 +1022,41 @@ extension CompactBlockProcessor {
...
+                    address: address.stringEncoded,
+                    networkType: config.network.networkType)
+                ),
+            total: Zatoshi(
+                try rustBackend.getTransparentBalance(
+                    dbData: config.dataDb,
+                    address: address.stringEncoded,
+                    networkType: config.network.networkType
+                )
+            )
```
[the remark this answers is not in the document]
pacu: yep. that's done!

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -10,378 +10,263 @@ import Foundation
...
         }
-        return result
+
+        defer { zcashlc_string_free(addressCStr) }
+
+        guard let address = String(validatingUTF8: addressCStr) else {
+            throw RustWeldingError.unableToDeriveKeys
+        }
+
+        return UnifiedAddress(validatedEncoding: address, network: networkType)
```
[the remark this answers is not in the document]
pacu: It is but I had to include it because the underlying functions need it. Having an environment variable that hits 2 or 3 calls deep it's even worse than including the network type.

nuttycom (approved):
utACK with one minor comment: would it be easier to avoid the build failures if we did something like this in `Package.swift` instead?

```
-        .package(name:"libzcashlc", url: "https://github.com/zcash-hackworks/zcash-light-client-ffi", branch: "bin/librustzcash_0_7")
+        .package(name:"libzcashlc", url: "https://github.com/zcash-hackworks/zcash-light-client-ffi", revision: "e8fbb84c1bec44af9dbef7e27c85f25e8f51a5af")
```

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
pacu: It depends on code that it's nullable from the C bindings. I'd rather have this unknown error to catch a weird malformed error from Rust than crashing the client application

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
pacu: should this be account based as well?

pacu: should this be account based now?

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
pacu: opened this issue

https://app.zenhub.com/workspaces/wallet-5db8aa0244512d0001e0968e/issues/zcash/zcashlightclientkit/548

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
```diff
@@ -1023,34 +1022,41 @@ extension CompactBlockProcessor {
 
 extension CompactBlockProcessor {
     public func getUnifiedAddres(accountIndex: Int) -> UnifiedAddress? {
```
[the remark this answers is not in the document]
pacu: The compact block processor is basically a lower level synchronizer. 
The synchronizer is aware of OS dependent events that the processor shouldn't care about. Addresses are not affected by that. That's why it had these methods

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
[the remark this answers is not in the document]
pacu: resolved

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -246,7 +246,7 @@ protocol ZcashRustBackendWelding {
     static func getTransparentReceiver(for uAddr: UnifiedAddress) throws -> TransparentAddress?
 
     /// gets the latest error if available. Clear the existing error
-    static func lastError() -> RustWeldingError?
```
nuttycom, line 38: I'm not sure that this is the correct change; the type of this now implies that some error must have occurred. It is a little more ergonomic, perhaps, but it doesn't seem right from a type perspective.

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -336,10 +321,12 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
     }
     
-    static func lastError() -> RustWeldingError? {
-        guard let message = getLastError() else { return nil }
+    static func lastError() -> RustWeldingError {
+        defer { zcashlc_clear_last_error() }
 
-        zcashlc_clear_last_error()
+        guard let message = getLastError() else {
+            return RustWeldingError.dataDbInitFailed(message: "Unknown Error")
```
nuttycom, line 328: `dataDbInitFailed` doesn't seem like the right error here.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
nuttycom: I think this got unresolved by the force-push.

nuttycom (changes requested):
The change to `lastError` doesn't make sense to me. Let's discuss.

nuttycom (approved):
utACK

pacu:
`testReOrgRemovesOutboundTxAndIsNeverMined` test fixed

PR #559, opened 2022-10-05
Discussion: New transaction data types

pacu:
Not for merging at all

### DIFF
--- a/Playgrounds/Transactions.playground/Contents.swift
+++ b/Playgrounds/Transactions.playground/Contents.swift
@@ -1,0 +1,142 @@
+import ZcashLightClientKit
+import Foundation
+
+/**
+ let expected_views = vec![
+            "CREATE VIEW v_transactions AS
+            SELECT id_tx,
+                   mined_height,
+                   tx_index,
+                   txid,
+                   expiry_height,
+                   raw,
+                   SUM(value) + MAX(fee) AS net_value,
+                   SUM(is_change) > 0 AS has_change,
+                   SUM(memo_present) AS memo_count
+            FROM (
+                SELECT transactions.id_tx            AS id_tx,
+                       transactions.block            AS mined_height,
+                       transactions.tx_index         AS tx_index,
+                       transactions.txid             AS txid,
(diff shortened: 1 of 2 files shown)


Playgrounds/Transactions.playground/Contents.swift
```diff
@@ -0,0 +1,177 @@
...
+    // MISSING: transaction kind. Is it outbound or inbound?
+    var id: Int
+    var minedHeight: BlockHeight
+    var tx_index: Int
+    var txid: Data
+    var expiryHeight: BlockHeight
+    var raw: Data
+    var netValue: Zatoshi
+    var hasChange: Bool
+    var memoCount: Int
```
pacu, line 67:
```suggestion
    var memoCount: Int
    var blocktime: TimeInterval  // MISSING: ability to show the date and time of the transaction on a listing
```

Playgrounds/Transactions.playground/Contents.swift
```diff
@@ -0,0 +1,177 @@
...
+                FROM   transactions
+                       JOIN sent_notes ON transactions.id_tx = sent_notes.tx
+            )
+            GROUP BY id_tx",
+
+ */
+
+
+struct TransactionOverview {
+    // MISSING: transaction kind. Is it outbound or inbound?
```
nuttycom, line 58: "transaction kind" isn't a coherent concept; it's almost always both inbound and outbound, because it sends value and receives change. However, "internal only" might be useful? For that, I think that I should add the fee paid as a field, and a transaction is then considered internal-only if net_value == -fee.

nuttycom: Actually, I guess if a transaction contains no sent notes, then it's outgoing.

pacu: In most cases it could be deduced that transactions with no  change are inbound, but that's not true. When you send MAX_FUNDS you have zero change outbound transactions. We should define a way to define the direction of a transaction unequivocally

PR #565, opened 2022-10-12
[#523] Make a CompactBlockProcessor an Actor

### LINKED ISSUE #523: Make a CompactBlockProcessor an Actor
I was investigating what it would take to make a CompactBlockProcessor an Actor - a new concurrency type. Such change usually consists of 2 things:

### Class -> Actor update
In order to update the CBP to an Actor, following methods must be updated:
```swift
// should be possible by using new getInfoAsync
func validateServer(completionBlock: @escaping (() -> Void))

// We need to re-thing the timer
private func setTimer()

// Conforming method from EnhancementStreamDelegate
func transactionEnhancementProgressUpdated(_ progress: EnhancementProgress)
```

### Use of an Actor
An Actor by its principle an async context so any use of a processor must be in awaited async context. SDKSynchronizer is using processor heavily but for now in synchronized context only.

LukasKorba:
Closes #523 

- Sample app refactored for the processor being an actor
- tests refactored as well
- dark side tests fixed
- code cleanup
- review of the code and last touches

# Author

# Reviewer

### DIFF
--- a/Example/ZcashLightClientSample/ZcashLightClientSample/DemoAppConfig.swift
+++ b/Example/ZcashLightClientSample/ZcashLightClientSample/DemoAppConfig.swift
@@ -16,6 +16,7 @@ enum DemoAppConfig {
     static var birthdayHeight: BlockHeight = ZcashSDK.isMainnet ? 935000 : 1386000
     
     static var seed = try! Mnemonic.deterministicSeedBytes(from: "live combine flight accident slow soda mind bright absent bid hen shy decade biology amazing mix enlist ensure biology rhythm snap duty soap armor")
+    
     static var address: String {
         "\(host):\(port)"
     }
--- a/Example/ZcashLightClientSample/ZcashLightClientSample/Get UTXOs/GetUTXOsViewController.swift
+++ b/Example/ZcashLightClientSample/ZcashLightClientSample/Get UTXOs/GetUTXOsViewController.swift
@@ -29,11 +29,13 @@ class GetUTXOsViewController: UIViewController {
 
         self.transparentAddressLabel.text = tAddress
 
-        // swiftlint:disable:next force_try
-        let balance = try! AppDelegate.shared.sharedSynchronizer.getTransparentBalance(accountIndex: 0)
-        
-        self.totalBalanceLabel.text = NumberFormatter.zcashNumberFormatter.string(from: NSNumber(value: balance.total.amount))
-        self.verifiedBalanceLabel.text = NumberFormatter.zcashNumberFormatter.string(from: NSNumber(value: balance.verified.amount))
+        Task { @MainActor in
+            // swiftlint:disable:next force_try
+            let balance = try! await AppDelegate.shared.sharedSynchronizer.getTransparentBalance(accountIndex: 0)
+            
+            self.totalBalanceLabel.text = NumberFormatter.zcashNumberFormatter.string(from: NSNumber(value: balance.total.amount))
+            self.verifiedBalanceLabel.text = NumberFormatter.zcashNumberFormatter.string(from: NSNumber(value: balance.verified.amount))
+        }
     }
     
     @IBAction func shieldFunds(_ sender: Any) {
--- a/Sources/ZcashLightClientKit/Block/Processor/CompactBlockDownload.swift
+++ b/Sources/ZcashLightClientKit/Block/Processor/CompactBlockDownload.swift
@@ -16,7 +16,7 @@ extension CompactBlockProcessor {
     ) async throws {
         try Task.checkCancellation()
         
-        setState(.downloading)
+        state = .downloading
         
         var buffer: [ZcashCompactBlock] = []
         var targetHeightInternal: BlockHeight?
--- a/Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
+++ b/Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
@@ -103,7 +103,7 @@ public enum CompactBlockProgress {
 }
 
 protocol EnhancementStreamDelegate: AnyObject {
-    func transactionEnhancementProgressUpdated(_ progress: EnhancementProgress)
+    func transactionEnhancementProgressUpdated(_ progress: EnhancementProgress) async
 }
 
 public protocol EnhancementProgress {
@@ -213,7 +213,7 @@ public extension Notification.Name {
 
 /// The compact block processor is in charge of orchestrating the download and caching of compact blocks from a LightWalletEndpoint
 /// when started the processor downloads does a download - validate - scan cycle until it reaches latest height on the blockchain.
-public class CompactBlockProcessor {
+public actor CompactBlockProcessor {
 
     /// Compact Block Processor configuration
     ///
@@ -312,31 +312,12 @@ public class CompactBlockProcessor {
         case synced
     }
     
-    // TODO: this isn't an Actor even though it looks like a good candidate, the reason:
-    // `state` lives in both sync and async environments. An Actor is demanding async context only
-    // so we can't take the advantage unless we encapsulate all `state` reads/writes to async context.
-    // Therefore solution with class + lock works for us butr eventually will be replaced.
-    // The future of CompactBlockProcessor is an actor (we won't need to encapsulate the state separately), issue 523,
-    // https://github.com/zcash/ZcashLightClientKit/issues/523
-    public class ThreadSafeState {
-        private var state: State = .stopped
-        let lock = NSLock()
-        
-        func setState(_ newState: State) {
-            lock.lock()
-            defer { lock.unlock() }
-            state = newState
-        }
-        
-        public func getState() -> State {
-            lock.lock()
-            defer { lock.unlock() }
-            return state
+    public internal(set) var state: State = .stopped {
+        didSet {
+            transitionState(from: oldValue, to: self.state)
         }
     }
     
-    public internal(set) var state = ThreadSafeState()
-
     var config: Configuration {
         willSet {
             self.stop()
@@ -348,7 +329,7 @@ public class CompactBlockProcessor {
     }
 
     var shouldStart: Bool {
-        switch self.state.getState() {
+        switch self.state {
         case .stopped, .synced, .error:
             return !maxAttemptsReached
         default:
@@ -386,7 +367,7 @@ public class CompactBlockProcessor {
     ///  - storage: concrete implementation of `CompactBlockStorage` protocol
     ///  - backend: a class that complies to `ZcashRustBackendWelding`
     ///  - config: `Configuration` struct for this processor
-    convenience init(
+    init(
         service: LightWalletService,
         storage: CompactBlockStorage,
         backend: ZcashRustBackendWelding.Type,
@@ -407,7 +388,7 @@ public class CompactBlockProcessor {
     /// Initializes a CompactBlockProcessor instance from an Initialized object
     /// - Parameters:
     ///     - initializer: an instance that complies to CompactBlockDownloading protocol
-    public convenience init(initializer: Initializer) {
+    public init(initializer: Initializer) {
         self.init(
             service: initializer.lightWalletService,
             storage: initializer.storage,
@@ -446,12 +427,6 @@ public class CompactBlockProcessor {
     
     deinit {
         cancelableTask?.cancel()
-    }
-
-    func setState(_ newState: State) {
-        let oldValue = state.getState()
-        state.setState(newState)
-        transitionState(from: oldValue, to: newState)
     }
 
     static func validateServerInfo(
@@ -500,7 +475,7 @@ public class CompactBlockProcessor {
     /// triggers the blockProcessorStartedDownloading notification
     ///
     /// - Important: subscribe to the notifications before calling this method
-    public func start(retry: Bool = false) throws {
+    public func start(retry: Bool = false) async {
         if retry {
             self.retryAttempts = 0
             self.processingError = nil
@@ -509,12 +484,12 @@ public class CompactBlockProcessor {
         }
 
         guard shouldStart else {
-            switch self.state.getState() {
+            switch self.state {
             case .error(let e):
                 // max attempts have been reached
                 LoggerProxy.info("max retry attempts reached with error: \(e)")
                 notifyError(CompactBlockProcessorError.maxAttemptsReached(attempts: self.maxAttempts))
-                setState(.stopped)
+                state = .stopped
             case .stopped:
                 // max attempts have been reached
                 LoggerProxy.info("max retry attempts reached")
@@ -529,7 +504,7 @@ public class CompactBlockProcessor {
             return
         }
 
-        self.nextBatch()
+        await self.nextBatch()
     }
 
     /**
@@ -545,14 +520,13 @@ public class CompactBlockProcessor {
         cancelableTask?.cancel()
 
         self.retryAttempts = 0
-        setState(.stopped)
     }
 
     /**
     Rewinds to provided height.
     If nil is provided, it will rescan to nearest height (quick rescan)
     */
-    public func rewindTo(_ height: BlockHeight?) throws -> BlockHeight {
+    public func rewindTo(_ height: BlockHeight?) async throws -> BlockHeight {
         self.stop()
 
         let lastDownloaded = try downloader.lastDownloadedBlockHeight()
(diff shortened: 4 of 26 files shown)


Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
```diff
@@ -449,8 +426,8 @@ public class CompactBlockProcessor {
     }
 
     func setState(_ newState: State) {
-        let oldValue = state.getState()
-        state.setState(newState)
+        let oldValue = state
```
Chlup, line 429: This code is fine as it is. 

I only have a suggestion. What if we drop this method and call `transitionState(from: oldValue, to: newState)` from state gettter/and setter?

Something like:
```
public internal(set) var state: State = .stopped {
    didSet { transitionState(from: oldValue, to: newState) }
}
```

Feel free to ignore this suggestion. Maybe it's bad idea because we introduce kinda side effect.

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
```diff
@@ -597,27 +574,25 @@ public class CompactBlockProcessor {
...
-                )
-                completionBlock()
-            } catch let error as LightWalletServiceError {
-                self.severeFailure(error.mapToProcessorError())
-            } catch {
-                self.severeFailure(error)
-            }
+    func validateServer(completionBlock: @escaping (() -> Void)) async {
+        do {
+            let info = try await self.service.getInfo()
```
Chlup, line 579: Are we ok with this no longer running on main thread?

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
```diff
@@ -766,41 +741,39 @@ public class CompactBlockProcessor {
...
+    private func nextBatch() async {
         setState(.downloading)
-        Task { @MainActor [self] in
-            do {
-                let nextState = try await NextStateHelper.nextStateAsync(
-                    service: self.service,
-                    downloader: self.downloader,
-                    config: self.config,
-                    rustBackend: self.rustBackend
+        do {
```
Chlup, line 746: Are we ok with this no longer run on main thread?

Sources/ZcashLightClientKit/Block/Utils/NotificationCenterDefault.swift
```diff
@@ -0,0 +1,30 @@
+//
+//  File.swift
+//  
+//
+//  Created by Lukáš Korba on 12.10.2022.
+//
+
+import Foundation
+
+struct NotificationCenterDefault {
```
Chlup, line 10: Maybe we could do extension of `NotificationCenter` instead of introducing new struct. And we could call new methods `postOnMainThread(....` and `postNotificationOnMainThread(.....` to be super explicit about what is going. Now it's not clear in code on first sight what is going on. But this more about naming and code style. This code works fine.

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -709,38 +711,40 @@ public class SDKSynchronizer: Synchronizer {
...
-                            total: initializer.getBalance()
-                        ),
-                        transparentBalance: (try? self.getTransparentBalance(accountIndex: 0)) ?? WalletBalance.zero,
-                        syncStatus: status,
-                        latestScannedHeight: self.latestScannedHeight
-                    )
-                ]
-            )
+            Task { @MainActor in
+                await NotificationCenterDefault.post(
```
Chlup, line 719: I ma little bit lost here. `NotificationCenterDefault.post` isn't async method. So how `await` can be used here?

Tests/DarksideTests/AdvancedReOrgTests.swift
```diff
@@ -101,16 +101,23 @@ class AdvancedReOrgTests: XCTestCase {
...
-            completion: { synchro in
-                synchronizer = synchro
-                initialVerifiedBalance = synchro.initializer.getVerifiedBalance()
-                initialTotalBalance = synchro.initializer.getBalance()
-                preTxExpectation.fulfill()
-                shouldContinue = true
-            },
-            error: self.handleError
-        )
+        try await withCheckedThrowingContinuation { continuation in
```
Chlup, line 104: Can you please describe a bit why `withCheckedThrowingContinuation` must be used? I don't mean in the code. I mean here in the comment. I am not exactly sure what is going on here and I would like to understand it. Thanks.

Tests/TestUtils/TestCoordinator.swift
```diff
@@ -136,7 +136,9 @@ class TestCoordinator {
     }
     
     func stop() throws {
-        synchronizer.stop()
+        Task {
```
Chlup, line 139: I am thinking about this change. Previously code was executed synchronously. So first stop was called and then variables were set to nil. Now probably `stop` method exists earlier than `synchronizer.stop()` is called. Are we ok with this?

Chlup (changes requested):

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
pacu: I think we are as long as the tests pass

Sources/ZcashLightClientKit/Block/Utils/NotificationCenterDefault.swift
pacu: I agree with this code being and extension with more explicit naming

Tests/DarksideTests/AdvancedReOrgTests.swift
pacu: This uses a continuation that belongs to this task. If you just await this the XCWait won't work because the expectation will be waited for in a different context

Tests/TestUtils/TestCoordinator.swift
pacu: this should immediately stop everything

pacu (changes requested):

Sources/ZcashLightClientKit/Block/Processor/CompactBlockDownload.swift
```diff
@@ -9,7 +9,7 @@
 import Foundation
 
 extension CompactBlockProcessor {
-    func compactBlockStreamDownload(
+    public func compactBlockStreamDownload(
```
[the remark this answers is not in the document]
LukasKorba: good catch, back to internal

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
LukasKorba: The method is not used anywhere in the SDK or wallet. ideally remove it but after talk with android team I decided to leave it there as is - they claimed maybe some devs are using method we don't. Anyway, the API changed as well co I removed the completion block and left the method available.
```swift
func validateServer() async { ...
```

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
LukasKorba: I asked myself the same and spent a lot of time testing it. I think nothing is dependent on main thread unless it touches UI. That is triggered by notifications only which we deliver on main thread now. So after testing and these facts I believe we are safe.

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
LukasKorba: The await is actually for the 
```swift
self.getTransparentBalance()
``` 
inside the post. I put the await right before the call and removed @MainActor as we explicitly handle main thread inside of the post.

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
```diff
@@ -500,7 +477,7 @@ public class CompactBlockProcessor {
     /// triggers the blockProcessorStartedDownloading notification
     ///
     /// - Important: subscribe to the notifications before calling this method
-    public func start(retry: Bool = false) throws {
+    public func start(retry: Bool = false) async throws {
```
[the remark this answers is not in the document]
LukasKorba: Not possible in this case as the start method is touching so many isolated properties.

```
All methods and mutable properties inside an actor are isolated to that actor by default, which means they cannot be accessed directly from code that’s external to the actor. Access to constant properties is automatically allowed because they are inherently safe from race conditions, but if you want you can make some methods excepted by using the nonisolated keyword.

Actor methods that are non-isolated can access other non-isolated state, such as constant properties or other methods that are marked non-isolated. However, they cannot directly access isolated state like an isolated actor method would; they need to use await instead.
```

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
LukasKorba: All recent changes of state handling were only because we needed thread safe solution and processor wasn't an actor.  The code before any async refactor for the state was 
```swift
public private(set) var state: State = .stopped {
        didSet {
            transitionState(from: oldValue, to: self.state)
        }
    }
```

and we can revert back to it thx to the Actor refactor.

Chlup (approved):

Sources/ZcashLightClientKit/Block/Processor/FigureNextBatch.swift
```diff
@@ -7,7 +7,7 @@
 
 import Foundation
 
-extension CompactBlockProcessor {
+public extension CompactBlockProcessor {
```
pacu, line 10: Why is this needed to be public?

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -151,23 +151,25 @@ public class SDKSynchronizer: Synchronizer {
...
         try self.initializer.initialize()
-        try self.blockProcessor.setStartHeight(initializer.walletBirthday)
+        try await self.blockProcessor.setStartHeight(initializer.walletBirthday)
         self.status = .disconnected
     }
 
     /// Starts the synchronizer
     /// - Throws: CompactBlockProcessorError when failures occur
-    public func start(retry: Bool = false) throws {
+    public func start(retry: Bool = false) async throws {
```
pacu, line 172: I have a couple of questions about implementation on the client side.
How would a wallet use this API? 
What happens if the Task context that calls start() is cancelled? will the Synchronizer stop all the tasks in-flight?

pacu (changes requested):
Good work! Left some questions about how clients would use the API

Sources/ZcashLightClientKit/Block/Processor/FigureNextBatch.swift
LukasKorba: removed

Tests/DarksideTests/SynchronizerTests.swift
```diff
@@ -102,14 +102,14 @@ final class SynchronizerTests: XCTestCase {
             XCTFail("Failed with error: \(testError)")
         })
 
-        DispatchQueue.main.asyncAfter(deadline: .now() + 5) {
-            self.coordinator.synchronizer.stop()
-        }
+        try await Task.sleep(nanoseconds: 5_000_000_000)
+        await self.coordinator.synchronizer.stop()
```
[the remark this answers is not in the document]
LukasKorba: solved

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
LukasKorba: Synchronizer changed to have start and stop without async context

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -80,33 +80,33 @@ public protocol Synchronizer {
...
     /// Extended Viewing Keys and a wallet birthday found in the initializer object
-    func prepare() throws
+    func prepare() async throws
 
     ///Starts this synchronizer within the given scope.
     ///
     ///Implementations should leverage structured concurrency and
     ///cancel all jobs when this scope completes.
-    func start(retry: Bool) throws
+    func start(retry: Bool) async throws
```
pacu, line 89: This doesn't match the implementation. 🧐

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -80,33 +80,33 @@ public protocol Synchronizer {
...
 
     /// Stop this synchronizer. Implementations should ensure that calling this method cancels all jobs that were created by this instance.
-    func stop() throws
+    func stop() async throws
```
pacu, line 92: Same as above

pacu (changes requested):
I think there are some changes that weren't pushed because synchronizer protocol still has start and stop async

Example/ZcashLightClientSample/ZcashLightClientSample/DemoAppConfig.swift
```diff
@@ -13,9 +13,12 @@ import MnemonicSwift
 enum DemoAppConfig {
     static var host = ZcashSDK.isMainnet ? "lightwalletd.electriccoin.co" : "lightwalletd.testnet.electriccoin.co"
     static var port: Int = 9067
-    static var birthdayHeight: BlockHeight = ZcashSDK.isMainnet ? 935000 : 1386000
+    static var birthdayHeight: BlockHeight = 2_000_000//ZcashSDK.isMainnet ? 935000 : 1386000
```
Chlup, line 16: We probably don't want to merge these changes.

Chlup (changes requested):

Sources/ZcashLightClientKit/Synchronizer.swift
LukasKorba: good catch, fixed (wonder why compiler is ok, no warnings, no nothing :()

Sources/ZcashLightClientKit/Synchronizer.swift
LukasKorba: good catch, fixed (wonder why compiler is ok, no warnings, no nothing :()

Example/ZcashLightClientSample/ZcashLightClientSample/DemoAppConfig.swift
LukasKorba: reverted

Chlup (approved):

Tests/TestUtils/TestCoordinator.swift
```diff
@@ -200,6 +200,12 @@ class TestCoordinator {
     }
 }
 
+extension CompactBlockProcessor {
```
pacu, line 203: [non-blocking] I don't understand this change.

pacu (approved):
Approved with non blocking comments.

Tests/TestUtils/TestCoordinator.swift
LukasKorba: The TestCoordinator modifies the processor's `config` but processor is actor-isolated context and write can't be done directly. You can read the value by awaiting it but can't modify it.

```swift
let c = await processor.config // is ok
await processor.config = newConfig // won't compile and honestly the syntax even doesn't make sense
await processor.modifyConfig(newConfig) // is ok
```

Example/ZcashLightClientSample/ZcashLightClientSample/Get UTXOs/GetUTXOsViewController.swift
```diff
@@ -29,11 +29,13 @@ class GetUTXOsViewController: UIViewController {
 
         self.transparentAddressLabel.text = tAddress
 
-        // swiftlint:disable:next force_try
-        let balance = try! AppDelegate.shared.sharedSynchronizer.getTransparentBalance(accountIndex: 0)
-        
-        self.totalBalanceLabel.text = NumberFormatter.zcashNumberFormatter.string(from: NSNumber(value: balance.total.amount))
-        self.verifiedBalanceLabel.text = NumberFormatter.zcashNumberFormatter.string(from: NSNumber(value: balance.verified.amount))
+        Task { @MainActor in
```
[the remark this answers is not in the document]
LukasKorba: I personally never found a use case for MainActor.run API. Here it just throws 
```
Cannot pass function of type '@Sendable () async throws -> ()' to parameter expecting synchronous function type
```
What we need is to spin up the task to update the UI and this syntax does exactly that.

Issue #613, opened 2022-11-09
Provide background tasks helpers


pacu:
We want to provide SDK clients with ways to spin up tasks to download and or sync blocks 

Since this is an application function. We must provide what others will need to use in their own tasks.

Additionally, create documentation that shows how to leverage this: 
- API docs 
- demo app samples
- tutorials 

Bonus points. Evaluate if this can be a sub module of the SDK that can be left out if not wanted.

PR #626, opened 2022-11-15
[#613] Add support for background task to sample app

Chlup:
- `SyncBlocksViewController` now uses shared synchronizer instead of locally created block processor.
- When app goes to background background task is created so SDK has time to stop the work.

# Author

# Reviewer

### DIFF
--- a/Example/ZcashLightClientSample/ZcashLightClientSample/AppDelegate.swift
+++ b/Example/ZcashLightClientSample/ZcashLightClientSample/AppDelegate.swift
@@ -6,9 +6,11 @@
 //  Copyright © 2019 Electric Coin Company. All rights reserved.
 //
 
+import Combine
+import MnemonicSwift
+import NotificationBubbles
 import UIKit
 import ZcashLightClientKit
-import NotificationBubbles
 
 var loggerProxy = SampleLogger(logLevel: .debug)
 
@@ -18,6 +20,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
     private var wallet: Initializer?
     private var synchronizer: SDKSynchronizer?
+    
+    var taskIdentifier: UIBackgroundTaskIdentifier = .invalid
+    var shouldStartSyncOnAppForeground = false
     
     var sharedSynchronizer: SDKSynchronizer {
         if let sync = synchronizer {
@@ -98,10 +103,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func applicationDidEnterBackground(_ application: UIApplication) {
         // Use this method to release shared resources, save user data, invalidate timers, and store enough application state information to restore your application to its current state in case it is terminated later.
         // If your application supports background execution, this method is called instead of applicationWillTerminate: when the user quits.
+
+        startBackgroundWork(on: application)
     }
     
     func applicationWillEnterForeground(_ application: UIApplication) {
         // Called as part of the transition from the background to the active state; here you can undo many of the changes made on entering the background.
+        do {
+            try sharedSynchronizer.applicationWillEnterForeground()
+        } catch {
+            loggerProxy.error("Failed to start synchronisation after coming to foreground. \(error)")
+        }
     }
     
     func applicationDidBecomeActive(_ application: UIApplication) {
@@ -112,6 +124,33 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Called when the application is about to terminate. Save data if appropriate. See also applicationDidEnterBackground:.
     }
 }
+
+// MARK: - Background
+
+extension AppDelegate {
+    func startBackgroundWork(on application: UIApplication) {
+        guard taskIdentifier == .invalid else { return }
+
+        self.taskIdentifier = application.beginBackgroundTask(
+            withName: "ZcashLightClientKit.SDKSynchronizer",
+            expirationHandler: { [weak self] in
+                loggerProxy.info("BackgroundTask Expiration Handler Called")
+                self?.endBackgroundWork()
+            }
+        )
+
+        sharedSynchronizer.applicationDidEnterBackground(resumeOnForeground: true) { [weak self] in
+            self?.endBackgroundWork()
+        }
+    }
+
+    @objc func endBackgroundWork() {
+        UIApplication.shared.endBackgroundTask(taskIdentifier)
+        taskIdentifier = .invalid
+    }
+}
+
+// MARK: - Misc
 
 /**
 The functions below are convenience functions for THIS SAMPLE APP.
--- a/Example/ZcashLightClientSample/ZcashLightClientSample/Send/SendViewController.swift
+++ b/Example/ZcashLightClientSample/ZcashLightClientSample/Send/SendViewController.swift
@@ -46,7 +46,7 @@ class SendViewController: UIViewController {
         super.viewDidAppear(animated)
         Task { @MainActor in
             do {
-                try await synchronizer.start(retry: false)
+                try synchronizer.start(retry: false)
                 self.synchronizerStatusLabel.text = SDKSynchronizer.textFor(state: synchronizer.status)
             } catch {
                 self.synchronizerStatusLabel.text = SDKSynchronizer.textFor(state: synchronizer.status)
--- a/Sources/ZcashLightClientKit/Synchronizer.swift
+++ b/Sources/ZcashLightClientKit/Synchronizer.swift
@@ -90,6 +90,23 @@ public protocol Synchronizer {
     /// Stop this synchronizer. Implementations should ensure that calling this method cancels all jobs that were created by this instance.
     func stop() throws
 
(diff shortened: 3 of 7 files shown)


Example/ZcashLightClientSample/ZcashLightClientSample/Send/SendViewController.swift
```diff
@@ -46,7 +46,7 @@ class SendViewController: UIViewController {
         super.viewDidAppear(animated)
         Task { @MainActor in
             do {
-                try await synchronizer.start(retry: false)
+                try synchronizer.start(retry: false)
```
pacu, line 49: I think we can remove the Task enclosing now that the await is gone

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -90,6 +90,23 @@ public protocol Synchronizer {
...
+    /// Synchronizer need some time to stop a work. So before calling this method use `UIApplication.shared.beginBackgroundTask` to request some
+    /// background execution time from operating system.
+    ///
+    /// - Parameters:
+    ///   - resumeSyncWhenAppGoesToForeground: Indicates if sync process should automatically start after app goes back to foreground. To support
+    ///                                        this call `applicationWillEnterForeground` when app will enter foreground.
+    ///   - backgroundWorkFinished: Closure is called when synchronizer finishes stop of the work.
+    func applicationDidEnterBackground(resumeSyncWhenAppGoesToForeground: Bool, backgroundWorkFinished: @escaping () -> Void)
+
+    /// Call this wehn application will enter foreground.
```
pacu, line 104:
```suggestion
    /// Call this when application will enter foreground.
```

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -90,6 +90,23 @@ public protocol Synchronizer {
...
 
+    /// Call this when application did enter background. Synchronizer stops it's work.
+    ///
+    /// Synchronizer need some time to stop a work. So before calling this method use `UIApplication.shared.beginBackgroundTask` to request some
+    /// background execution time from operating system.
+    ///
+    /// - Parameters:
+    ///   - resumeSyncWhenAppGoesToForeground: Indicates if sync process should automatically start after app goes back to foreground. To support
+    ///                                        this call `applicationWillEnterForeground` when app will enter foreground.
+    ///   - backgroundWorkFinished: Closure is called when synchronizer finishes stop of the work.
```
pacu, line 101:
```suggestion
    ///   - backgroundWorkFinished: Closure is called when synchronizer finally stops
```

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -90,6 +90,23 @@ public protocol Synchronizer {
...
+    func applicationDidEnterBackground(resumeSyncWhenAppGoesToForeground: Bool, backgroundWorkFinished: @escaping () -> Void)
```
pacu, line 102:
```suggestion
    func applicationDidEnterBackground(resumeOnForeground: Bool, synchronizerStopped: @escaping () -> Void)
```

I think these names are more concise. Also stopped doesn't imply that synchronizer finished syncing, it just refers to it stopping.

Tests/TestUtils/TestCoordinator.swift
```diff
@@ -307,7 +307,7 @@ enum TestSynchronizerBuilder {
         )
 
         let synchronizer = try SDKSynchronizer(initializer: initializer)
-        if case .seedRequired = try await synchronizer.prepare(with: seed) {
+        if case .seedRequired = try synchronizer.prepare(with: seed) {
```
pacu, line 310: let's remove the `async` from line 294 since this function no longer requires being called on an async context

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -90,6 +90,23 @@ public protocol Synchronizer {
...
+
+    /// Call this wehn application will enter foreground.
+    ///
+    /// When syncing process was previously interrupted by calling `applicationDidEnterBackground` with `resumeSyncWhenAppGoesToForeground` parameter
+    /// set to `true` then this method starts syncing process. Otherwise nothing happens.
+    func applicationWillEnterForeground() throws
```
pacu, line 108: I don't think this needs to be part to the `protocol` since it's iOS dependent.

pacu (changes requested):
I think this does not close #613, just one of its items.

Example/ZcashLightClientSample/ZcashLightClientSample/AppDelegate.swift
```diff
@@ -98,10 +105,21 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
     
     func applicationWillEnterForeground(_ application: UIApplication) {
         // Called as part of the transition from the background to the active state; here you can undo many of the changes made on entering the background.
+
+        if shouldStartSyncOnAppForeground {
+            do {
+                loggerProxy.info("Starting sync after returning to foreground.")
+                try sharedSynchronizer.start()
+            } catch {
+                loggerProxy.error("Failed to start synchronisation after comming to foreground. \(error)")
```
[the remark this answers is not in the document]
Chlup: Fixed

Example/ZcashLightClientSample/ZcashLightClientSample/AppDelegate.swift
```diff
@@ -113,6 +131,49 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 }
 
+// MARK: - Background
+
+extension AppDelegate {
+    func startBackgroundWork() {
```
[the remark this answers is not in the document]
Chlup: About "Maybe creating some helper class that SDK clients can use directly to spin up the tasks":
   I am not sure about this. Management of background task is one call which creates it. And then another call which invalidates it. It's super simple. I can create a helper which wraps these two calls but I don't see a point in that. There will still be two calls on app side. I don't see a reasonable way how could we simplify this.

About that documentation:
  What is your idea that should be in this documentation? In my code I start background task only to get time to stop synchronizer. So I can write suggestion that this background task should be used only to stop synchronizer. Basically only to use `applicationDidEnterBackground`. Where this documentation should be?

Sources/ZcashLightClientKit/Synchronizer.swift
Chlup: Fixed

Sources/ZcashLightClientKit/Synchronizer.swift
Chlup: Fixed

Sources/ZcashLightClientKit/Synchronizer.swift
Chlup: If the app uses `Synchronizer` protocol instead of `SDKSynchronizer` type then this feature won't be available. And if we require developers to use `SDKSynchronizer` type then what is the purpose of `Synchronizer` protocol?

Example/ZcashLightClientSample/ZcashLightClientSample/AppDelegate.swift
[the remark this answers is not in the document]
pacu: > About "Maybe creating some helper class that SDK clients can use directly to spin up the tasks":
I am not sure about this. Management of background task is one call which creates it. And then another call which invalidates it. It's super simple. I can create a helper which wraps these two calls but I don't see a point in that. There will still be two calls on app side. I don't see a reasonable way how could we simplify this.

I trust your expertise and judgement on this. 

> Where this documentation should be?
we place dedicated docs con `/docs` and refer to it in the main README.md as something worth checking out.
Also add a CHANGELOG.md entry about these changes.

> What is your idea that should be in this documentation? 
I think we need to provide guidance on:
- how to handle the scenario of user backgrounding the app and what the app should do about the SDKSynchronizer
- how to handle the scenario of the user returning to the app and what it should do about the SDKSynchronizer.
- How to create long processing tasks that allow iOS to wake up the client app to sync blocks and what are the ideal conditions when that must happen (e.g.: device connected to power and Wi-Fi)
- How to create app refreshes for scanning some bit of blocks (when the app has caught up with latest blocks and would benefit from such app refreshes)

We should support this scenarios with code samples. (these could be either in the Demo app or ECC Wallet, whatever you think it is more content rich and fruitful)

Sources/ZcashLightClientKit/Synchronizer.swift
pacu: Its original purpose was to be a OS-Agnostic interface that could host many implementations but we never got to that.  :(

pacu (changes requested):
Answered comments! 👍

pacu (changes requested):

Chlup:
Let's put this on side bench for now. We can reopen it when we get back to it.

PR #648, opened 2022-11-23
[#646] SDK sync process resumes correctly

### LINKED ISSUE #646: Unable to correctly resume sync process
SDK is unable to properly resume sync process. The main problem is that SDK computes one range and then use it for all actions necessary for sync. 

This is what is currently happening inside SDK:
1. Let's say we want to sync blocks 2000000 - 2100000.
2. Sync process starts. Range which should be used to do sync is computed in `NextStateHelper.nextStateAsync`. This range is computed based on which blocks are downloaded or which blocks have been scanned. This is the first root cause. No blocks are downloaded or scanned yet so the computed range is 2000000 - 2100000.
3. Computed range is passed to `processNewBlocks()` method. And this range is used for all actions necessary for sync (download, validation, scan, enhancement, fetch UTXO). And this is the second root cause. 
4. Sync process is interrupted while downloading blocks for some reason.
5. Sync process starts again. Previously SDK was able to download 20000 blocks. So now computed range for sync is 202000 - 210000.
6. Computed range is passed to `processNewBlocks()`. And this range is used for all actions necessary for sync. And here the problem happens. All the actions works with range from 2020000. Which is ok for download. But all other actions will ignore first 20000 blocks. 

Steps to reproduce:
1. Start the SDK Sample app.
2. Open "Sync Blocks" and start syncing.
3. Kill the app during downloading phase.
4. Run the app again and start syncing again.
5. All the blocks will be downloaded. But only blocks with height => height that was used on second sync will be scanned and enhanced.

### LINKED ISSUE #660: Any rewind causes full rescan
In #618 removing of cache DB after sync was introduced. Which is ok on it's own. But it doesn't work correctly with rewind. 

To compute sync range height of last downloaded block is used. And this height is loaded from cache DB. After #618 only the latest block is kept in cache DB so next sync range is correctly computed. When rewind happens that one latest block is removed and there is no other. So there is no last downloaded block and full sync range is used in next sync. And this happens for any rewind. And that is bad.

Fastest fix is to disable removing of cache DB. Proper fix is to find different way how to track what was downloaded independently on cache DB state.

Chlup:
Closes #646 
Closes #660 

- Previously we had one range for each sync which was used for each phase of sync process. Newly there is separate range for each phase of the sync process.
- I added `InternalSyncProgress` utility. This utility tracks progress of some phases. And it is able to compute right ranges which should be used for syncing.
- Some unused download code from `CompactBlockProcessor` is removed.

# Author

# Reviewer

- [x] Checklist review: Did you go through the code with the [Code Review Guidelines](../blob/main/CODE_REVIEW_GUIDELINES.md) checklist?
- [x] Ad hoc review: Did you perform an ad hoc review?  _In addition to a first pass using the code review guidelines, do a second pass using your best judgement and experience which may identify additional questions or comments. Research shows that code review is most effective when done in multiple passes, where reviewers look for different things through each pass._
- [x] Automated tests: Did you review the automated tests?
- [x] Manual tests: Did you review the manual tests?_You will find manual testing guidelines under our [manual testing section](../blob/mater/docs/testing/manual_testing)_
- [x] Run the app: Did you run the app and try the changes? While the CI server runs the app to look for build failures or crashes, humans running the app are more likely to notice unexpected log messages, UI inconsistencies, or bad output data.


Sources/ZcashLightClientKit/Block/Utils/InternalSyncProgress.swift
```diff
@@ -0,0 +1,121 @@
+//
+//  File.swift
```
LukasKorba, line 2: Pls name the file (when we have the header it should be at least meaningful)

Tests/TestUtils/InternalSyncProgressMemoryStorage.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  File.swift
```
LukasKorba, line 2: Her as well, renamer File.swift

LukasKorba (approved):
LGTM but haven't tested, the whole topic is pretty complex and I take it as workaround before we introduce some proper solution (state machine?).

Sources/ZcashLightClientKit/Block/Utils/InternalSyncProgress.swift
pacu: This bug in Xcode is incredible.

Sources/ZcashLightClientKit/Block/Utils/InternalSyncProgress.swift
```diff
@@ -0,0 +1,121 @@
...
+        case latestUTXOFetchedHeight
+    }
+
+    enum TypeOfWork {
+        /// There is nothing to do. Everything is synced.
+        case none
+        /// These is something to sync.
+        case sync
+        /// Some local height is higher then latest block height from backend API.
+        case higherValue
```
pacu, line 38: I think this is a debug / log worthy situation.

```suggestion
        case wait(latestScannedHeight: BlockHeight, latestBlockHeight: BlockHeight)
```

Sources/ZcashLightClientKit/Block/Utils/InternalSyncProgress.swift
```diff
@@ -0,0 +1,121 @@
...
+        Key.allCases.forEach { key in
+            let finalRewindHeight = min(self.get(key), to)
+            self.set(finalRewindHeight, key)
+        }
+    }
+
+    /// Computes if there is something to sync.
+    ///
+    /// - Parameters:
+    ///   - latestBlockHeight: Latest height fetched from backend API.
```
pacu, line 65:
```suggestion
    ///   - latestBlockHeight: Latest height fetched from LightWalletd API.
```

Sources/ZcashLightClientKit/Block/Utils/InternalSyncProgress.swift
```diff
@@ -0,0 +1,121 @@
...
+    ///   - latestDownloadedHeight: The height of latest downloaded block.
+    ///   - latestScannedHeight: The of the latest scanned block.
+    /// - Returns: Which type of work must be done.
+    func typeOfWorkRemaining(latestBlockHeight: BlockHeight, latestDownloadedHeight: BlockHeight, latestScannedHeight: BlockHeight) -> TypeOfWork {
```
pacu, line 69: I don't see any particular tests of this function that has many OR sentences and can be tricky. 
I also think it could have better documentation of the intent so that any further changes to it can be checked against the documentation and verify that the logic we intend to have here still holds

Sources/ZcashLightClientKit/Block/Utils/InternalSyncProgress.swift
```diff
@@ -0,0 +1,121 @@
...
+            return .none
+        }
+    }
+
+    func computeSyncRanges(
+        birthday: BlockHeight,
+        latestBlockHeight: BlockHeight,
+        latestDownloadedBlockHeight: BlockHeight,
+        latestScannedHeight: BlockHeight
+    ) -> SyncRanges {
```
pacu, line 90: we should add docs about what this computation is and what the returned values mean.

Sources/ZcashLightClientKit/Block/Processor/FetchUnspentTxOutputs.swift
```diff
@@ -50,6 +54,8 @@ extension CompactBlockProcessor {
                         height: utxo.height,
                         networkType: config.network.networkType
                     ) ? refreshed.append(utxo) : skipped.append(utxo)
+
+                    await internalSyncProgress.set(utxo.height, .latestUTXOFetchedHeight)
```
pacu, line 58: I think the `computeSyncRanges` and `typeOfWorkRemaining` might make assumptions that are incorrect because of how "fetch" works. 

we fetch and `upsert` every UTXO available for the addresses that the user has derived and given out for receiving transparent funds. There's caveat here. If the user doesn't shield its funds, this code will always receive the same UTXOs. These will eventually get "old" and reflect a height that's far away from the chain tip. 

Does a `utxo.height` that is far from the latest mined block mean that the wallet is not synced? Absolutely not. It's just that the wallet hasn't bothered to shield those funds. 

There are re-org considerations as well. Line 73 below shows that the `range.upperbound` is saved as `latestUTXOFetchedHeight`. What are we keeping this partial result on line 58 for?

If there's a re-org that removes UTXOs from blocks we might not be capturing that because we are storing a `latestUTXOFetchedHeight` of a UTXO that does not exist anymore or that it might appear on a later block.

How does `internalSyncProgress` handle re-orgs?

Sources/ZcashLightClientKit/Block/Utils/InternalSyncProgress.swift
Chlup: Renamed

Sources/ZcashLightClientKit/Block/Utils/InternalSyncProgress.swift
Chlup: I changed a code a this enum is now gone.

Sources/ZcashLightClientKit/Block/Utils/InternalSyncProgress.swift
Chlup: I updated the documentation and added tests for InternalSyncProgress.

Sources/ZcashLightClientKit/Block/Processor/FetchUnspentTxOutputs.swift
Chlup: > I think the `computeSyncRanges` and `typeOfWorkRemaining` might make assumptions that are incorrect because of how "fetch" works.
> 
> we fetch and `upsert` every UTXO available for the addresses that the user has derived and given out for receiving transparent funds. There's caveat here. If the user doesn't shield its funds, this code will always receive the same UTXOs. These will eventually get "old" and reflect a height that's far away from the chain tip.
> 
> Does a `utxo.height` that is far from the latest mined block mean that the wallet is not synced? Absolutely not. It's just that the wallet hasn't bothered to shield those funds.

I am not sure if I understand this. My thinking was that if we fetch UTXO for some height then we are done for this height. And no new UTXOs can be find at that height. So it's enough to fetch UTXO for that height just once.

> 
> There are re-org considerations as well. Line 73 below shows that the `range.upperbound` is saved as `latestUTXOFetchedHeight`. What are we keeping this partial result on line 58 for?

Because exception can be thrown during processing of each UTXO. And that latest processed height wouldn't be stored. It can work without that. SDK just fetches already fetch UTXOs on next sync.

> 
> If there's a re-org that removes UTXOs from blocks we might not be capturing that because we are storing a `latestUTXOFetchedHeight` of a UTXO that does not exist anymore or that it might appear on a later block.
> 
> How does `internalSyncProgress` handle re-orgs?

`InternalSyncProgess` has `rewind()` method. It is called in `validationFailed()`.

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
```diff
@@ -605,38 +598,58 @@ public actor CompactBlockProcessor {
...
-                try await compactBlockBatchScanning(range: range)
-                try await compactBlockEnhancement(range: range)
-                try await fetchUnspentTxOutputs(range: range)
                 try await handleSaplingParametersIfNeeded()
                 try await removeCacheDB()
                 
                 if !Task.isCancelled {
-                    await processBatchFinished(range: range)
+                    let lastScannedHeightAfterSync = try transactionRepository.lastScannedHeight()
+                    await processBatchFinished(range: lastScannedHeightBeforeSync...lastScannedHeightAfterSync)
```
Chlup, line 652: @pacu Could you please check this? Previously we had one range so it was super easy to have range which is passed to `processBatchFinished()` method. But now we have multiple ranges. I think that I did right think but may I overlooked something.

Sources/ZcashLightClientKit/Block/Processor/FetchUnspentTxOutputs.swift
pacu: Thanks for the response. I'll take a look at `validationFailed` and `rewind` !

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
```diff
@@ -583,8 +575,9 @@ public actor CompactBlockProcessor {
 
         // clear cache
         try downloader.rewind(to: BlockHeight(rewindHeight))
```
pacu, line 577:
```suggestion
        let rewindHeight = BlockHeight(rewindHeight)
        try downloader.rewind(to: rewindHeight)
        await internalSyncProgress.rewind(to: rewindHeight)

        self.lastChainValidationFailure = nil
        self.lowerBoundHeight = try? downloader.lastDownloadedBlockHeight()
        return rewindHeight
```

Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift
pacu: uhhmm it seem so but how do you think the processor should react when this guard statement is negative?


````Swift
  internal func processBatchFinished(range: CompactBlockRange) async {
        guard processingError == nil else {
            await retryProcessing(range: range)
            return
        }
````

Comment (pacu) [reply Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift:652]: shall we pass  syncRanges instead?

**Review by pacu — CHANGES_REQUESTED** [review]
I responded the questions. I think this is good progress

File: Sources/ZcashLightClientKit/Block/Processor/FetchUnspentTxOutputs.swift (hunk 7 above)
Comment (pacu) [reply Sources/ZcashLightClientKit/Block/Processor/FetchUnspentTxOutputs.swift:58]: I think this is doing the right thing but I'll try to test it as well

File: Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift (hunk 8 above)
Comment (Chlup) [reply Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift:652]: Problem of SyncRanges is that ranges inside are optionals. `nil` instead of range means that we don't want to do specific phase. For example when everything is already downloaded then range for download is `nil`. And what should be passed when all are for example `nil`? Which range should be used?

Comment (pacu) [reply Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift:653]: do we have a copy of the original `SyncRanges` at this point in the code? I think the main problem is that we originally coded a function that says "Finished! .... oh no wait.. there's an error" because of how NSOperations fail asynchronously and cancel lazily and unreliable. Now that we don't use them, I believe we can split the `finished` from  the `failed` and this dilemma will be resolved.

Comment (pacu) [reply Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift:653]: If we do have a copy of the original range we failed at, we could patch this by adding that to the parameters passed and the retry would pass that to start over. Otherwise probably refactoring this is for the better

**Review by pacu — CHANGES_REQUESTED** [review]
@Chlup sorry for the delay. I think I might thought of a path forward to the processingFinished dilemma. Please let me know your thoughts

File: Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift (hunk 8 above)
Comment (Chlup) [reply Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift]: I tried to change this. And then I realised. I don't think that `retryProcessing()` is ever called. It is called from `processBatchFinished()` only when `processingError` is not `nil`. `processingError` is set only in `serverFailure()` and `fail()` methods. But none of these two methods is called inside sync process. 

What do you think about this? Did I miss something? 

I am thinking about how to approach this. Either we say it's ok and remove it. Or we say it is a bug and it must be fixed. But it may mean some kind of error tracking overhaul.

Comment (pacu) [reply Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift]: > I tried to change this. And then I realised. I don't think that retryProcessing() is ever called. It is called from processBatchFinished() only when processingError is not nil. processingError is set only in serverFailure() and fail() methods. But none of these two methods is called inside sync process.

I think this might be the case since we moved away from NSOperations now task that fail throw errors instead of reporting them lazily. My take is that it might be safe to remove it.

**Comment by pacu** [comment]
I tested this using `master` of ECC Wallet. And found a bug.

When you do a quick re-scan (profile -> Rescan Wallet -> Quick Re-scan) the wallet shows no balance and the SDK reports being synced without rescanning

````
2022-12-05 18:38:45.388734-0300 ECC-Wallet-Testnet[8507:260681] [] InternalSyncProgress[86] - computeNextState(latestBlockHeight:latestDownloadedBlockHeight:latestScannedHeight:walletBirthday:) - DEBUG 🐞 - Init numbers:
latestDownloadedHeight:  2116088
latestScannedHeight:     2116089
latestEnhancedHeight:    0
latestUTXOFetchedHeight: 0
````


I think it's because the scanned height should be updated to the latestScannedheight that results of rewinding.

**Review by pacu — CHANGES_REQUESTED** [review]
See comments on ECC Wallet tests

File: Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift (hunk 8 above)
Comment (Chlup) [reply Sources/ZcashLightClientKit/Block/Processor/CompactBlockProcessor.swift]: Ok. Code removed. Thanks for advice.

**Comment by Chlup** [comment]
> I tested this using `master` of ECC Wallet. And found a bug.
> 
> When you do a quick re-scan (profile -> Rescan Wallet -> Quick Re-scan) the wallet shows no balance and the SDK reports being synced without rescanning
> 
> ```
> 2022-12-05 18:38:45.388734-0300 ECC-Wallet-Testnet[8507:260681] [] InternalSyncProgress[86] - computeNextState(latestBlockHeight:latestDownloadedBlockHeight:latestScannedHeight:walletBirthday:) - DEBUG 🐞 - Init numbers:
> latestDownloadedHeight:  2116088
> latestScannedHeight:     2116089
> latestEnhancedHeight:    0
> latestUTXOFetchedHeight: 0
> ```
> 
> I think it's because the scanned height should be updated to the latestScannedheight that results of rewinding.

@pacu This is because `latestScannedHeight` is taken from `TransactionRepository` and that doesn't support rewinding. I suggest to handle this as separate issue. It's not regression of this change and it was happening before. I wouldn't like to make this PR even bigger. I can create ticket and try to solve it separately. Is that Ok with you?

**Comment by Chlup** [comment]
@pacu So I did some changes. Now #660 is fixed in this code. Latest downloaded block is no longer tracked by using cache DB but it's tracked in `InternalSyncProgress`. Thanks to this cache DB can be completely removed after sync. I did some testing and it seems fine to me. Please check it out.

**Review by pacu — CHANGES_REQUESTED** [review]
I tested the changes on ECC Wallet Testnet, was able to quick re-scan as I wasn't able to.

Most of the darksidewalletd tests passed, except for `TransactionEnhancementTests` which never gets to finish. This might be giving us some interesting caveat we are not catching. 

````
2022-12-10 15:16:22.564835-0300 xctest[8175:67599] [INFO ℹ️] CompactBlockProcessor.swift - nextBatch() - Line: 790 -> Lightwalletd might be syncing: latest downloaded block height is: 687059 while latest blockheight is reported at: 663200
```

pacu (approved):
image
All tests pass.

Issue #556, opened 2022-10-04
Refactor TransactionRepository to use DB Views


pacu:
work from branch `feature/zip-316...` 

The rust layer exposes views on the DataDb to return lists of transactions
see:
https://github.com/zcash/librustzcash/blob/main/zcash_client_sqlite/src/wallet/init.rs#L426-L516

Refactor `TransactionRepository` to maintain but use the views from the Db instead.

view `v_transactions` for listing transactions

for `v_tx_received`  for received transactions

for `v_tx_sent` for sent transactions

The views are not equal to the existing queries but similar. adapt the existing data types to what's needed tot reflect the views.

Reference:
The roomoval PR has the changes on the entities that now use the data derived from views

https://github.com/zcash/zcash-android-wallet-sdk/pull/739/files

https://github.com/zcash/zcash-android-wallet-sdk/tree/main/sdk-lib/src/main/java/cash/z/ecc/android/sdk/internal/db/derived

https://github.com/zcash/ZcashLightClientKit/pull/559

PR #674, opened 2022-12-14
[#556] Define structures for db view and add new repository methods

Chlup:
- Structures that represent transactions DB views are newly define.
- New structures are:
  - `TransactionNG.Overview`
  - `TransactionNG.Received`
  - `TransactionNG.Sent`

- `TransactionNG` is used as temporary name. `Transaction` is existing structure in the code. `TransactionNG` will be eventually renamed to `Transaction` once current `Transaction` is removed from code.

- New methods that are using new structures are added to `TransactionRepository`. Those are slightly changed compared to predecessors. Those are now little bit simpler to use.

# Author

# Reviewer

### DIFF
--- a/Sources/ZcashLightClientKit/DAO/TransactionDao.swift
+++ b/Sources/ZcashLightClientKit/DAO/TransactionDao.swift
@@ -89,6 +89,10 @@ class TransactionSQLDAO: TransactionRepository {
     var dbProvider: ConnectionProvider
     var transactions = Table("transactions")
     private var blockDao: BlockSQLDAO
+
+    fileprivate let transactionsView = View("v_transactions")
+    fileprivate let receivedTransactionsView = View("v_tx_received")
+    fileprivate let sentTransactionsView = View("v_tx_sent")
     
     init(dbProvider: ConnectionProvider) {
         self.dbProvider = dbProvider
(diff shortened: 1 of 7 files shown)


Sources/ZcashLightClientKit/DAO/TransactionDao.swift
```diff
@@ -415,6 +419,105 @@ extension TransactionSQLDAO {
...
+        let query = sentTransactionsView
+            .order(TransactionNG.Sent.Column.minedHeight.asc, TransactionNG.Sent.Column.id.asc)
+            .filter(TransactionNG.Sent.Column.rawID == Blob(bytes: rawID.bytes)).limit(1)
+
+        return try execute(query) { try TransactionNG.Sent(row: $0) }
+    }
+
+    private func execute<Entity>(_ query: View, createEntity: (Row) throws -> Entity) throws -> Entity {
+        let entities: [Entity] = try execute(query, createEntity: createEntity)
+        guard let entity = entities.first else { throw TransactionRepositoryError.notFound }
```
pacu, line 506: Why would we consider an Error that a query returns no results? Wouldn't it be better to return an Optional<Entity>?

pacu (changes requested):

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
Chlup: We could. It just seems to me like using two different technologies to report some exceptional state. It means that if you call this API you have to handle exception and Optional. But there is not hard technical reason why to do this in this way. I just think that it will be easier to use in this way. If you think that we should use Optionals here no problem at all. I can change it.

pacu: Seems reasonable. let's leave it this way

pacu (approved):

pacu (approved):

Sources/ZcashLightClientKit/Entity/TransactionEntity.swift
```diff
@@ -6,6 +6,155 @@
...
+        public let raw: Data
+        public let rawID: Data
+        public let value: Zatoshi
+    }
+}
+
+extension TransactionNG.Overview {
+    enum Column {
+        static let id = Expression<Int>("id_tx")
+        static let minedHeight = Expression<Int>("mined_height")
```
ccjernigan, line 63: Heights are technically UInt32 in the Rust code, so we should probably make this Int64 in the Swift code and then convert to our BlockHeight typesafe object.

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
```diff
@@ -415,6 +419,105 @@ extension TransactionSQLDAO {
...
+        return entity
+    }
+
+    private func execute<Entity>(_ query: View, createEntity: (Row) throws -> Entity) throws -> [Entity] {
+        let entities = try dbProvider
+            .connection()
+            .prepare(query)
```
ccjernigan, line 513: I don't know enough about how you're internally maintaining the database connections.  Is the dbProvider maintaining a long-lived connection?

Would it make sense in this case to use something like `.lazy()` so that the entire table doesn't have to be queried at once?

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
```diff
@@ -415,6 +419,105 @@ extension TransactionSQLDAO {
...
+            .filter(transaction.blocktime > TransactionNG.Sent.Column.blockTime && transaction.index > TransactionNG.Sent.Column.index)
+            .limit(limit)
+
+        return try execute(query) { try TransactionNG.Sent(row: $0) }
+    }
+
+    func findSent(in range: BlockRange, limit: Int) throws -> [TransactionNG.Sent] {
+        let query = sentTransactionsView
+            .filter(
+                TransactionNG.Sent.Column.minedHeight >= Int(range.start.height) &&
```
ccjernigan, line 487: Heights shouldn't be cast to Int for the comparison, since BlockHeight is UInt32.

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
```diff
@@ -415,6 +419,105 @@ extension TransactionSQLDAO {
     }
 }
 
+extension TransactionSQLDAO {
+    func find(id: Int) throws -> TransactionNG.Overview {
+        let query = transactionsView
+            .order(TransactionNG.Overview.Column.minedHeight.asc, TransactionNG.Overview.Column.id.asc)
```
ccjernigan, line 425: If we only expect to return a single item, does it make sense to include the `order()` call?

ccjernigan (commented):
I'm not as familiar with the iOS code base, but I do have a few comments/questions inline below.

Sources/ZcashLightClientKit/Entity/TransactionEntity.swift
Chlup: In the iOS codebase we have `BlockHeight` data type which is just an alias to Int. Any device on which our code can currently run has 64bit CPU so both Int and Int64 can hold same values. I switched from `Int` to `BlockHeight` at height related rows so it's little bit more clear.

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
Chlup: It has "lazy" behavior inside. Checkout `SimpleConnectionProvider` to see real implementation of how connection is created and provided.

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
Chlup: It's similar case as [above](https://github.com/zcash/ZcashLightClientKit/pull/674#discussion_r1050141458). I switched to `BlockHeight` from `Int` to be more clear.

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
Chlup: No it doesn't. Good catch.

Chlup:
@ccjernigan Thanks for the review!

ccjernigan:
> @ccjernigan Thanks for the review!

You're welcome!  I don't have any other questions—thanks for helping me understand the iOS codebase a bit better too 😄

ccjernigan (approved):

PR #827, opened 2023-03-10
[#826] Change how the SDK is initialised

### LINKED ISSUE #826: Change how the SDK is initialised
I was thinking about how SDK is being initialised and what are the consequences for the Secant. Currently we can’t create instance of `SDKSynchronizer` right away when the Secant starts. And we have to deal with sdk synchronizer being optional inside the Secant. We can create instance of `SDKSynchronizer` after we read data from keychain. It’s because `Initializer` requires `walletBirthday` and `viewingKeys` in it’s constructor. And both of these values are used in `Initializer.initialize()`  method. If these values would be passed to some functions instead of being constructor parameters it could make our life easier.

If we rework a bit how SDK is initialised we could improve how we work with the SDK inside the Secant. For example we would unblock https://github.com/zcash/secant-ios-wallet/issues/497. I suggest don’t do any changes in logic just move few params from constructors to functions.

Chlup:
Closes #826

- `viewingKeys` and `walletBirthday` are removed from `Initializer` constuctor. These parameters are moved to `SDKSynchronizer.prepare` function.
- Constructor of the `SDKSynchronizer` no longer throws exception.
- Any value emitted from `lastState` stream before `SDKSynchronizer.prepare` is called has `latestScannedHeight` set to 0.
- `Initializer.initialize` function isn't public anymore. To initialize SDK call `SDKSynchronizer.prepare` instead.
- Real birthday is now known after the `SDKSynchronizer.prepare` is called. But this function isn't async and therefore it's hard to pass birthday from here to `CompactBlockProcessor` in synchronous manner. We must be sure that it's safe to call start() after prepare function finishes.
        For this reason `CompactBlockProcessor` has `walletBirthdayProvider` instead of `walletBirthday`. And this new parameter is closure which reads birthday from `Initiliazer` on each call. Thanks to this `CompactBlockProcessor` has always the right birthday.
- Some cleanup is done in `TestCoordinator`.

# Author

# Reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# unreleased 
+- [#826] Change how the SDK is initialized
+    
+    - `viewingKeys` and `walletBirthday` are removed from `Initializer` constuctor. These parameters
+are moved to `SDKSynchronizer.prepare` function.
+    - Constructor of the `SDKSynchronizer` no longer throws exception.
+    - Any value emitted from `lastState` stream before `SDKSynchronizer.prepare` is called has 
+`latestScannedHeight` set to 0.
+    - `Initializer.initialize` function isn't public anymore. To initialize SDK call `SDKSynchronizer.prepare` 
+instead. 
+
+
 # 0.19.1-beta
 ## Checkpoints added
 Mainnet
--- a/Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
+++ b/Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
@@ -151,7 +151,8 @@ actor CompactBlockProcessor {
         public var maxBackoffInterval = ZcashSDK.defaultMaxBackOffInterval
         public var maxReorgSize = ZcashSDK.maxReorgSize
         public var rewindDistance = ZcashSDK.defaultRewindDistance
-        public var walletBirthday: BlockHeight
+        let walletBirthdayProvider: () -> BlockHeight
+        public var walletBirthday: BlockHeight { walletBirthdayProvider() }
         public private(set) var downloadBufferSize: Int = 10
         private(set) var network: ZcashNetwork
         private(set) var saplingActivation: BlockHeight
@@ -171,7 +172,7 @@ actor CompactBlockProcessor {
             retries: Int,
             maxBackoffInterval: TimeInterval,
             rewindDistance: Int,
-            walletBirthday: BlockHeight,
+            walletBirthdayProvider: @escaping () -> BlockHeight,
             saplingActivation: BlockHeight,
             network: ZcashNetwork
         ) {
@@ -185,7 +186,7 @@ actor CompactBlockProcessor {
             self.retries = retries
             self.maxBackoffInterval = maxBackoffInterval
             self.rewindDistance = rewindDistance
-            self.walletBirthday = walletBirthday
+            self.walletBirthdayProvider = walletBirthdayProvider
             self.saplingActivation = saplingActivation
             self.cacheDbURL = cacheDbURL
             assert(downloadBatchSize >= scanningBatchSize)
@@ -197,7 +198,7 @@ actor CompactBlockProcessor {
             spendParamsURL: URL,
             outputParamsURL: URL,
             saplingParamsSourceURL: SaplingParamsSourceURL,
-            walletBirthday: BlockHeight,
+            walletBirthdayProvider: @escaping () -> BlockHeight,
             network: ZcashNetwork
         ) {
             self.fsBlockCacheRoot = fsBlockCacheRoot
@@ -205,7 +206,7 @@ actor CompactBlockProcessor {
             self.spendParamsURL = spendParamsURL
             self.outputParamsURL = outputParamsURL
             self.saplingParamsSourceURL = saplingParamsSourceURL
-            self.walletBirthday = walletBirthday
+            self.walletBirthdayProvider = walletBirthdayProvider
             self.saplingActivation = network.constants.saplingActivationHeight
             self.network = network
             self.cacheDbURL = nil
@@ -340,7 +341,7 @@ actor CompactBlockProcessor {
     /// Initializes a CompactBlockProcessor instance from an Initialized object
     /// - Parameters:
     ///     - initializer: an instance that complies to CompactBlockDownloading protocol
-    init(initializer: Initializer) {
+    init(initializer: Initializer, walletBirthdayProvider: @escaping () -> BlockHeight) {
         self.init(
             service: initializer.lightWalletService,
             storage: initializer.storage,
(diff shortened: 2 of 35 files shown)


pacu (changes requested):
Great job!

Small request lets make `Birthday` more meaningful with some validation of the range.

Sources/ZcashLightClientKit/Model/Birthday.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+//  Created by Michal Fousek on 10.03.2023.
+//
+
+import Foundation
+
+/// Wrapper over simple birthday value. Wallet birthday value isn't known when instance of the `SDKSynchronizer` is created. It's known after
+/// `SDKSynchronizer.prepare()` is called. And then it's hard to pass birthday value to different places of the SDK. So let's use instance of this
+/// class and just change underlying value.
+public class Birthday {
```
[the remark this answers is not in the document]
Chlup: I probably didn't want to go to this place with this type. In reality I just need class wrapper for the Int value. And because I am using it for wallet birthday I called it `Birthday`. It also invalid values from birthday perspective. When `Initializer` is created birthday is set to 0 before `initialize()` is called. Maybe it would be better to take one step back and don't use `Birthday` name for this.

CHANGELOG.md
```diff
@@ -8,6 +20,7 @@ Sources/ZcashLightClientKit/Resources/checkpoints/mainnet/2007500.json
 Sources/ZcashLightClientKit/Resources/checkpoints/mainnet/2010000.json
 ````
 
+<<<<<<< HEAD
```
pacu, line 23: I think this swizzled in

Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
```diff
@@ -197,15 +198,15 @@ actor CompactBlockProcessor {
             spendParamsURL: URL,
             outputParamsURL: URL,
             saplingParamsSourceURL: SaplingParamsSourceURL,
-            walletBirthday: BlockHeight,
+            walletBirthday: ReferenceContainer<BlockHeight>,
```
pacu, line 201: I don't understand how this change. 

Why is this reference container needed?

Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
```diff
@@ -151,7 +151,8 @@ actor CompactBlockProcessor {
         public var maxBackoffInterval = ZcashSDK.defaultMaxBackOffInterval
         public var maxReorgSize = ZcashSDK.maxReorgSize
         public var rewindDistance = ZcashSDK.defaultRewindDistance
-        public var walletBirthday: BlockHeight
+        let birthday: ReferenceContainer<BlockHeight>
```
pacu, line 154: Why do we need these two attributes that seem very similar in purpose?

pacu (changes requested):
I'm not getting the sense of some of the changes, probably due to the lack of some context. 

I don't think I might be getting the problem so I can't say it's the right approach or not

Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
Chlup: I tried to explain it in PR description:

Real birthday is now known after the `SDKSynchronizer.prepare` is called. But this function isn't async and therefore it's hard to pass birthday from here to `CompactBlockProcessor` in synchronous manner. We must be sure that it's safe to call `start()` after `prepare` function finishes. For this reason I created `ReferenceContainer` class which wraps birthday height. And only one instance of this class is used. So when `Initializer` stores computed birthday in this instance it's available in `CompactBlockProcessor`. It's not the cleanest way how to do this but for now it's probably simplest and fastest way how to do this.

Chlup: ReferenceContrainer is gone. I changed the initial implementation to:

`CompactBlockProcessor` has `walletBirthdayProvider` instead of `walletBirthday`. And this new parameter is closure which reads birthday from `Initiliazer` on each call. Thanks to this `CompactBlockProcessor` has always the right birthday.

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -96,12 +95,11 @@ public class Initializer {
...
-    public let viewingKeys: [UnifiedFullViewingKey]
-    /// The effective birthday of the wallet based on the height provided when initializing
-    /// and the checkpoints available on this SDK
-    private(set) public var walletBirthday: BlockHeight
 
-    private var lowerBoundHeight: BlockHeight
+    /// The effective birthday of the wallet based on the height provided when initializing and the checkpoints available on this SDK.
+    ///
+    /// This contains valid value only after `initialize` function is called.
+    private(set) public var walletBirthday: BlockHeight = .zero
```
pacu, line 102: If you already know what the network is, our "conceptual Zero" value is sapling activation height. Our SDK won't deal with any shielded funds from sprout.

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -301,7 +283,7 @@ public class Initializer {
     /// - Parameter seed: ZIP-32 Seed bytes for the wallet that will be initialized
     /// - Throws: `InitializerError.dataDbInitFailed` if the creation of the dataDb fails
     /// `InitializerError.accountInitFailed` if the account table can't be initialized. 
-    public func initialize(with seed: [UInt8]?) throws -> InitializationResult {
+    func initialize(with seed: [UInt8]?, viewingKeys: [UnifiedFullViewingKey], walletBirthday: BlockHeight) throws -> InitializationResult {
```
pacu, line 286: this makes me think that there should be a different initializer. providing viewing keys and seed at the same time does not make sense for us, since we are not currently supporting spend capabilities and view-only keys from different seeds. 

if a user provides viewing keys that originate from other bytes than the seed bytes provided, the behaviour will be undefined

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -175,13 +179,23 @@ public class SDKSynchronizer: Synchronizer {
         }
     }
 
-    public func prepare(with seed: [UInt8]?) throws -> Initializer.InitializationResult {
-        if case .seedRequired = try self.initializer.initialize(with: seed) {
+    public func prepare(
```
pacu, line 182: same comment as for `initialize` method

pacu (changes requested):

pacu:
Reviewed and notice there are a few more changes to do in regards to initialization

Sources/ZcashLightClientKit/Initializer.swift
Chlup: This value won't ever be used. Until `prepare()` is called (and this value is filled with real birthday) the `SDKSynchronizer` will be in `unprepared` state and refused to do anything. But to make this better I replace `.zero` with: 
`Checkpoint.birthday(with: 0, network: network).height`. 

Is that ok?

pacu: Yes that will make it default to sapling activation

Sources/ZcashLightClientKit/Initializer.swift
Chlup: My change just moves `viewingKeys` and `walletBirthday` params from constructor to this method. I didn't created any new flows, logic or scenarios. Nothing prevented you before to pass viewing keys into constructor and then different seed into this function or to `prepare`.

pacu (approved):

LukasKorba (approved):

Issue #888, opened 2023-03-30
Consider making RustBackend an Actor


pacu:
Given the situation where we can have many synchronizers running at once and rustwelding having async methods, we could probably explore making that an Swift Actor and move away from static methods. 

The rust backend actor would
- be constructed for a given network and paths as constants
- CompactBlockProcessor should own it and be the exclusive user of it.
- functions that don't need IO could be `nonisolated`

PR #898, opened 2023-04-03
[#888] Make actor from ZcashRustBackendWelding

Chlup:
This PR isn't final thing. Documentation isn't updated. Tests are broken. Some tests for new stuff are missing. I want to do all of this in future PR so this one doesn't become even bigger. It's huge even now. This PR will be merged to the feature branch for now.

@LukasKorba public API of the `DerivationTool` changed. Even the way how it can be instanced changed. All this is because of async changes in `ZcashRustBackend`. Please check it out if it is ok to use in the Secant in future. Thanks.

- `ZcashRustBackend` is actor now. So majority of methods in this actor are now async.
- Some methods stayed `static` in `ZcashRustBackend`. It would be hard to pass instance of the `ZcashRustBackend` to the places where these methods are used in static manner. And it would change lot of APIs. But it isn't problem from technical perspective because these methods would be `nonisolated` otherwise.
- Methods `lastError()` and `getLastError()` in `ZcashRustBackend` are now private. This makes sure that ther won't be aby race condition between other methods and these two error methods.
- All the methods for which was `lastError()` used in code now throw error. So `lastError()` is no longer needed outside of the `ZcashRustBackend`.
- There are in the public API related to `DerivationTool`.
- `DerivationTool` now requires instance of the `ZcashRustBackend`. And `ZcashRustBackend` isn't public type. So `DerivationTool` doesn't have any public constructor now. It can be created only via `Initializer.makeDerivationTool()` instance method.
- `deriveUnifiedSpendingKey()` and `deriveUnifiedFullViewingKey()` in `DerivationTool` are now async. It is because these are using `ZcashRustBackend` inside. `DerivationTool` offers alternative (closure and combine) APIs. But downside is that there is no sync API to dervie spending key or viewing key.
- Some methods of the `DerivationTool` are now static. These methods don't use anything that requires instance of the `DerivationTool` inside.

# Author

# Reviewer


Sources/ZcashLightClientKit/CombineSynchronizer.swift
```diff
@@ -8,13 +8,6 @@
 import Combine
 import Foundation
 
-/* These aliases are here to just make the API easier to read. */
-
-// Publisher which emitts completed or error. No value is emitted.
-public typealias Completable<E: Error> = AnyPublisher<Void, E>
```
pacu, line 14: does this change belong here?

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -375,23 +351,23 @@ class ZcashRustBackend: ZcashRustBackendWelding {
...
     static func isValidTransparentAddress(_ address: String, networkType: NetworkType) -> Bool {
         guard !address.containsCStringNullBytesBeforeStringEnding() else {
             return false
         }
 
         return zcashlc_is_valid_transparent_address([CChar](address.utf8CString), networkType.networkId)
     }
-    
+
     static func isValidSaplingExtendedFullViewingKey(_ key: String, networkType: NetworkType) -> Bool {
```
pacu, line 371: I think that even though at a glance it makes sense to keep this static, then it's weird how DerivationTool has both instance and Type.

I'd say we make this func of Rust backend and remove the static nature of them. 

Any thoughts or things I'm missing?

Sources/ZcashLightClientKit/Model/WalletTypes.swift
```diff
@@ -136,7 +136,7 @@ public struct TransparentAddress: Equatable, StringEncoded, Comparable {
     /// - Throws: `KeyEncodingError.invalidEncoding`when the provided encoding is
     /// found to be invalid
     public init(encoding: String, network: NetworkType) throws {
-        guard DerivationTool(networkType: network).isValidTransparentAddress(encoding) else {
+        guard DerivationTool.isValidTransparentAddress(encoding, networkType: network) else {
```
pacu, line 139: I wonder if we could extract address encoding validation to a separate struct or class that avoids Derivation tool relying on both Type and actor of rustwelding

pacu (changes requested):
Great work! I think this good progress also surfaced other problems we had and were masking with the static Type. It would be good that we address them as well in this opportunity.

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
Chlup: That was my original idea. I tried it and the amount of changes exploded. Those events stayed static mostly because those are used in `WalletTypes.swift` to validate addresses.

Sources/ZcashLightClientKit/Model/WalletTypes.swift
Chlup: I agree that we should address it. I am just not sure if in this PR or even in this ticket. The amount of changes is big already. I think that this kind of change is out of scope of this ticket.

pacu: Agreed. Let's follow up on that then! 💪

pacu (changes requested):
LGTM!

One small thing we are missing is CHANGELOG.md updates for public API changes.

pacu (approved):
Changelog will be update on later PR

PR #915, opened 2023-04-07
Split key derivation from Rust backend actor

pacu:

# Author

# Reviewer


Sources/ZcashLightClientKit/Rust/ZcashKeyDerivationBackend.swift
```diff
@@ -0,0 +1,320 @@
...
+                UInt(seed.count),
+                accountIndex,
+                networkType.networkId
+            )
+        }
+
+        defer { zcashlc_free_binary_key(binaryKeyPtr) }
+
+        guard let binaryKey = binaryKeyPtr?.pointee else {
+            throw lastError() ?? .genericError(message: "No error message available")
```
Chlup, line 231: This is the reason why we can't do this decoupling. In my attempt to solve this problem I did almost the same thing as you did in this PR. But now there can be race condition when accessing error. When everything was in one object access to the error was protected by actor. Now we have two objects and this access isn't protected anymore. And different threads calling rust may rewrite the errors. And each instance of the `DerivationTool` makes new instance of the `ZcashKeyDerivationBackend`. Which means more instances that can call rust and access error without any lock. We are basically where we were before introducing `ZcashRustBackend` as actor. 

The one thing that we could do is to remove actor and any kind of async from anything related to `ZcashRustBackend`. And introduce lock that will be shared between any instance of any object that uses rusts. And this lock will protect the Rust and we can do this decoupling. 

What do you think?

Chlup (changes requested):

Sources/ZcashLightClientKit/Rust/ZcashKeyDerivationBackend.swift
Chlup: @pacu I added one more commit to this branch. `ZcashKeyDerivationBackend` and `ZcashRustBackend` and now both structs. No actor is used. And both use one lock (`rustBackendGlobalLock`). Thanks to this we have thread safety when accessing rust and we can decouple rust as much as we can. And we can have any number of the instances as we can. 

From the outside world rust is called in async functions so it's not blocking when called via public API.

Tests are fine. Some are failing but not because this change but probably because I moved `testTempDirectory` to `Environment`. I will do another PR to fix this.

Please tell me what you think.

pacu: > This is the reason why we can't do this decoupling. In my attempt to solve this problem I did almost the same thing as you did in this PR. But now there can be race condition when accessing error.

if we always get the error before returning from the async function would the thread be reused before then? 

I recall @str4d saying errors were per-thread and not per-process.

Chlup: Oooooh. You are right. I was thinking about this so much that I become confused. I will retract my commit. We can "undraft" this PR.

Chlup (approved):

PR #1036, opened 2023-05-09
[#1035] Create ZcashError.unknown

### LINKED ISSUE #1035: Create ZcashError.unknown
Throwing functions throws `any Error` and our goal is to throw ZcashError in the whole SDK. The issue is that any app adopting SDK using throwing functions end up in 
```swift
do {
// ...
} catch error is ZcashError {
// all ok here
} catch {
// any Error here
}
```
Passing error in the apps would be complicated. We can improve this experience by presenting ZcashError.unknown and mapping any Error into it. 

That would result in:

```swift
do {
// ...
} catch {
guard error is ZcashError else { 
    throw error.mapToZcashError()
}
throw error 
}
```

LukasKorba:
Closes #1035 

- new ZcashError.unknown that wraps and handles any Error added

# Author

# Reviewer

### DIFF
--- a/Sources/ZcashLightClientKit/Extensions/Error+ZcashError.swift
+++ b/Sources/ZcashLightClientKit/Extensions/Error+ZcashError.swift
@@ -1,0 +1,18 @@
+//
+//  Error+ZcashError.swift
+//  
+//
+//  Created by Lukáš Korba on 09.05.2023.
+//
+
+import Foundation
+
+public extension Error {
+    func toZcashError() -> ZcashError {
+        if let zcashError = self as? ZcashError {
+            return zcashError
+        } else {
+            return ZcashError.unknown(self)
+        }
+    }
+}
--- a/Sources/ZcashLightClientKit/Error/ZcashError.swift
+++ b/Sources/ZcashLightClientKit/Error/ZcashError.swift
@@ -11,6 +11,10 @@ error originates. And it can help with debugging.
 import Foundation
 
 public enum ZcashError: Equatable, Error {
+    /// Some error happened that is not handled as `ZcashError`.
+    /// Use case is to map `any Error`(in the apps for example) to this one so it can be assured only `ZcashError` is floating through apps.
+    /// ZUNKWN0001
+    case unknown(_ error: Error)
     /// Updating of paths in `Initilizer` according to alias failed.
     /// ZINIT0001
     case initializerCantUpdateURLWithAlias(_ url: URL)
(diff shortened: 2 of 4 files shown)


pacu (changes requested):
My main objection to this is that it defeats the purpose of errors having locality and being pin pointed to there they occurred. 

Even a more local but yet generic error has some value of providing clients some code location.

Being reported an unknown error would be undesirable to us, because:
1- our app throws unknown errors.
2- we don't get any information beforehand that the user can capture to fix it.

Chlup:
> My main objection to this is that it defeats the purpose of errors having locality and being pin pointed to there they occurred.
> 
> Even a more local but yet generic error has some value of providing clients some code location.
> 
> Being reported an unknown error would be undesirable to us, because: 1- our app throws unknown errors. 2- we don't get any information beforehand that the user can capture to fix it.

I agree. Problem is that our SDK emits and throws `Error` type. Not `ZcashError`. It may be possible to change error types in streams. But it's impossible to specify error type when throwing error from method. So we need some easy way how to convert `Error` type to `ZcashError`. Because in fact all the errors generated by the SDK are `ZcashError`. Without `unknown` error there will be some Optional handling when doing `Error` -> `ZcashError` conversion. `unknown` is there to not need to handle Optionals when doing that conversion. And it should never be used directly. Like no function should explicitly throw `ZcashError.unknown`. It will be used only in case that we made a mistake and somewhere is thrown something else than `ZcashError`. And it will be used only in case that client uses `toZcashError()` function. 

I checked the PR and `toZcashError()` is wrongly implemented due to misunderstanding. Implementation should looks like this:

LukasKorba:
Main reason we added this is to use TCA actions
```swift
case someFailure(ZcashError)
```

so we need to ensure ZcashError is possible to pass from a throwing function. If it's not a ZcashError, we don't have a tool to wrap it from the app side.

pacu:
I guess we don't have much wiggle room here and will have to live with it.

Sources/ZcashLightClientKit/Extensions/Error+ZcashError.swift
```diff
@@ -0,0 +1,18 @@
...
+//  Error+ZcashError.swift
+//  
+//
+//  Created by Lukáš Korba on 09.05.2023.
+//
+
+import Foundation
+
+public extension Error {
+    func toZcashError() -> ZcashError {
```
pacu, line 11: Also explain when this should be used and why

PR #1155, opened 2023-07-02
Fix inverted sync status flags

### LINKED ISSUE #1154: Logic error in `SyncStatus` enum
### Describe the issue
In `Synchronizer.swift`, there is an enum `SyncStatus` that provides the following 3 convenience properties:

```
public var isSyncing: Bool {
        guard case .syncing = self else { return true }
        return false
}
    
public var isSynced: Bool {
       guard case .upToDate = self else { return true }
       return false
}

public var isPrepared: Bool {
       guard case .unprepared = self else { return false }
       return true
}
```

There are a couple issues with this:
1. The logic is simply wrong. Each of these is inverted.
2. Using guard here is confusing (and probably what led to this logic error). Generally, guard is used to bail early if a precondition is unmet. In this case, we're mapping a case value to a Bool flag. Better to use a simple if check, which is in fact what the analogous `InternalSyncStatus` does. In fact, one could argue the implementation for both should be the same for consistency.

#### If so, please list the steps to reproduce below:
1. Try to check on the sync status by using one of these utility properties.

### Expected behaviour
Tell us what should happen
In my use case, if the SDK is syncing (thus isSyncing should return true), I want to show an indeterminate progress bar. 

### Actual behaviour + errors
Tell us what happens instead including any noticeable error output (any messages
displayed on-screen when e.g. a crash occurred)

The app syncs with no loading indicator, and when the sync finished, the loading indicator shows (because the logic is reversed)

- App Version: 
- iOS Version:
- Device: (if applies)

tw0po1nt:
Closes #1154 

# Author

- [x] Documentation: Did you update Docs as appropiate? (E.g [README.md](../blob/main/README.md), etc.)
- [x] Did you provide Screenshots of what the App looks like before and after your changes as part of the description of this PR? (only applicable to UI Changes)

# Reviewer

### DIFF
--- a/Sources/ZcashLightClientKit/Synchronizer.swift
+++ b/Sources/ZcashLightClientKit/Synchronizer.swift
@@ -336,17 +336,26 @@ public enum SyncStatus: Equatable {
     case error(_ error: Error)
     
     public var isSyncing: Bool {
-        guard case .syncing = self else { return true }
+        if case .syncing = self {
+            return true
+        }
+        
         return false
     }
     
     public var isSynced: Bool {
-        guard case .upToDate = self else { return true }
+        if case .upToDate = self {
+            return true
+        }
+        
         return false
     }
 
     public var isPrepared: Bool {
-        guard case .unprepared = self else { return false }
+        if case .unprepared = self {
+            return false
+        }
+        
         return true
     }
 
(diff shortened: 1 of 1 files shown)


Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -336,17 +336,26 @@ public enum SyncStatus: Equatable {
     case error(_ error: Error)
     
     public var isSyncing: Bool {
-        guard case .syncing = self else { return true }
+        if case .syncing = self {
```
pacu, line 339: Good catch @tw0po1nt. 

This should have broken a few tests. 🥶

LukasKorba: None actually, at least in Zcash case. The computed variables are not used anywhere in the Zcash code.

Issue #1247, opened 2023-09-10
[SBS] DarkSideTests checks and fixes


LukasKorba:
In order to make Spend-before-Sync branch buidable, the #1242 needed to be done. But DarkSideTests probably need deeper thoughts and research how to modify it. So most of the refactored code will be probably commented out and marked with this ticket to review.

nuttycom:
I'm reopening this ticket so that the work that needs to be done can be tracked in DAG for post-release cleanup/reconsideration.

PR #1257, opened 2023-09-11
Review PR for changes to ZcashLightClientKit for fast spendability.

nuttycom:
This PR is not intended to be merged; instead, it's intended to allow for a unified UI for review of the `feature/DAG-sync` changes that includes changes already merged to `main`.


Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
```diff
@@ -638,7 +649,7 @@ extension CompactBlockProcessor {
         retryAttempts = 0
         consecutiveChainValidationErrors = 0
 
-        let lastScannedHeight = await latestBlocksDataProvider.latestScannedHeight
+        let lastScannedHeight = await latestBlocksDataProvider.maxScannedHeight
         // Some actions may not run. For example there are no transactions to enhance and therefore there is no enhance progress. And in
         // cases like this computation of final progress won't work properly. So let's fake 100% progress at the end of the sync process.
         await send(event: .progressUpdated(1))
```
nuttycom, line 655: I can't comment on the relevant line in the context of this PR, but is the `validationFailed` function below still live code?

Sources/ZcashLightClientKit/Block/Download/BlockDownloader.swift
```diff
@@ -201,11 +201,7 @@ actor BlockDownloaderImpl {
 
         let pushMetrics: (BlockHeight, Date, Date) -> Void = { [metrics] lastDownloadedBlockHeight, startTime, finishTime in
             metrics.pushProgressReport(
-                progress: BlockProgress(
-                    startHeight: totalProgressRange.lowerBound,
-                    targetHeight: totalProgressRange.upperBound,
-                    progressHeight: Int(lastDownloadedBlockHeight)
-                ),
+                progress: 0,
```
nuttycom, line 204: What does it imply for this progress value to be hardcoded to `0`?

Sources/ZcashLightClientKit/Block/Enhance/BlockEnhancer.swift
```diff
@@ -135,11 +135,7 @@ extension BlockEnhancerImpl: BlockEnhancer {
             }
             
             metrics.pushProgressReport(
-                progress: BlockProgress(
-                    startHeight: range.lowerBound,
-                    targetHeight: range.upperBound,
-                    progressHeight: range.upperBound
-                ),
+                progress: 0,
```
nuttycom, line 138: Same question here, what does this hardcoded `0` mean?

Sources/ZcashLightClientKit/Block/FetchUnspentTxOutputs/UTXOFetcher.swift
```diff
@@ -86,11 +86,7 @@ extension UTXOFetcherImpl: UTXOFetcher {
         }
 
         metrics.pushProgressReport(
-            progress: BlockProgress(
-                startHeight: 0,
-                targetHeight: 1,
-                progressHeight: 1
-            ),
+            progress: 0,
```
nuttycom, line 89: Please document the meaning of these hardcoded `0` values.

Sources/ZcashLightClientKit/Block/Scan/BlockScanner.swift
```diff
@@ -50,38 +49,29 @@ extension BlockScannerImpl: BlockScanner {
             try Task.checkCancellation()
 
             let previousScannedHeight = lastScannedHeight
+            let startHeight = previousScannedHeight + 1
 
             // TODO: [#576] remove this arbitrary batch size https://github.com/zcash/ZcashLightClientKit/issues/576
```
nuttycom, line 54: Is this comment still relevant? Or is the batch size now load-bearing given that it's derived from the scan ranges?

Sources/ZcashLightClientKit/Block/Scan/BlockScanner.swift
```diff
@@ -50,38 +49,29 @@ extension BlockScannerImpl: BlockScanner {
...
             }
 
             let scanFinishTime = Date()
 
-            if let lastScannedBlock = try await transactionRepository.lastScannedBlock() {
-                lastScannedHeight = lastScannedBlock.height
-                await latestBlocksDataProvider.updateLatestScannedHeight(lastScannedHeight)
-                await latestBlocksDataProvider.updateLatestScannedTime(TimeInterval(lastScannedBlock.time))
-            }
+            lastScannedHeight = startHeight + Int(batchSize) - 1
```
nuttycom, line 67: This has the potential to be incorrect if `scanBlocks` scans fewer than `batchSize` blocks. We should really have `scanBlocks` return metadata about the scanning that was completed; there are other features that have needed that as well. Not for 2.0.0, this is related to https://github.com/zcash/librustzcash/issues/918 which should be addressed in the post-2.0.0 cleanup phase.

Sources/ZcashLightClientKit/Block/Scan/BlockScanner.swift
```diff
@@ -50,38 +49,29 @@ extension BlockScannerImpl: BlockScanner {
...
-                let progress = BlockProgress(
-                    startHeight: totalProgressRange.lowerBound,
-                    targetHeight: totalProgressRange.upperBound,
-                    progressHeight: lastScannedHeight
-                )
+                try await didScan(lastScannedHeight, batchSize)
 
                 metrics.pushProgressReport(
-                    progress: progress,
+                    progress: 0,
```
nuttycom, line 74: Please document these hardcoded `0`s.

Sources/ZcashLightClientKit/CombineSynchronizer.swift
```diff
@@ -59,8 +58,6 @@ public protocol CombineSynchronizer {
 
     func getRecipients(for transaction: ZcashTransaction.Overview) -> SinglePublisher<[TransactionRecipient], Never>
 
-    func allPendingTransactions() -> SinglePublisher<[ZcashTransaction.Overview], Error>
```
nuttycom, line 62: It might be useful to restore this at some later point, based upon information in the `v_transactions` view.

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -409,35 +409,30 @@ public class Initializer {
...
+        if let seed, try accountRepository.getAll().isEmpty {
+            var chainTip: UInt32?
+            
+            if walletMode == .restoreWallet {
+                chainTip = UInt32(try await lightWalletService.latestBlockHeight())
+            }
+            
+            _ = try await rustBackend.createAccount(
+                seed: seed,
+                treeState: try checkpoint.treeState().serializedData(partial: false).bytes,
```
nuttycom, line 433: Nonblocking: this should pass the `TreeState`, not its bytes. This confusion resulted from a bad comment of mine on an earlier PR.

Sources/ZcashLightClientKit/Metrics/SDKMetrics.swift
```diff
@@ -84,9 +82,7 @@ public class SDKMetrics {
         guard isEnabled else { return }
         
         let blockMetricReport = BlockMetricReport(
-            startHeight: progress.startHeight,
-            progressHeight: progress.progressHeight,
-            targetHeight: progress.targetHeight,
+            progress: 0,
```
nuttycom, line 85: A `progress` value is passed as an argument to this function, why is it not used here?

Sources/ZcashLightClientKit/Repository/TransactionRepository.swift
```diff
@@ -11,10 +11,10 @@ protocol TransactionRepository {
     func closeDBConnection()
     func countAll() async throws -> Int
     func countUnmined() async throws -> Int
-    func blockForHeight(_ height: BlockHeight) async throws -> Block?
-    func lastScannedHeight() async throws -> BlockHeight
-    func lastScannedBlock() async throws -> Block?
-    func firstUnenhancedHeight() throws -> BlockHeight?
+//    func blockForHeight(_ height: BlockHeight) async throws -> Block?
```
nuttycom, line 14: These can be deleted.

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
```diff
@@ -35,9 +35,11 @@ protocol ZcashRustBackendWelding {
     /// have been received by the currently-available account (in order to enable
     /// automated account recovery).
     /// - parameter seed: byte array of the zip32 seed
+    /// - parameter treeState: byte array containing the TreeState Protobuf object for the height prior to the account birthday
+    /// - parameter recoverUntil: the fully-scanned height up to which the account will be treated as "being recovered"
     /// - Returns: The `UnifiedSpendingKey` structs for the number of accounts created
     /// - Throws: `rustCreateAccount`.
-    func createAccount(seed: [UInt8]) async throws -> UnifiedSpendingKey
+    func createAccount(seed: [UInt8], treeState: [UInt8], recoverUntil: UInt32?) async throws -> UnifiedSpendingKey
```
nuttycom, line 42: `treeState` here should be the concrete `TreeState` type.

Tests/DarksideTests/AdvancedReOrgTests.swift
```diff
@@ -359,25 +367,26 @@ class AdvancedReOrgTests: ZcashTestCase {
         
         sleep(2)
         
+        // TODO: [#1247] needs to review this to properly solve, https://github.com/zcash/ZcashLightClientKit/issues/1247
```
nuttycom, line 370: I've reopened #1247 to track the rework that needs to be done here; it's fine if it's resolved by changing the test architecture but I don't want it to get lost.

nuttycom (commented):
I opened this review-only PR, so I cannot approve or request changes. None of my comments are blocking, but it would be helpful to have the questions addressed and where changes are required, please create issues.

str4d (commented):
Reviewed as of 2ac984a0ddafe1afd8bf9ef2b5825a56dbca346b (excluding some files under `Tests/`).

Tests/DarksideTests/PendingTransactionUpdatesTest.swift
```diff
@@ -163,19 +170,20 @@ class PendingTransactionUpdatesTest: ZcashTestCase {
...
+//            return
+//        }
+//
+//        /*
+//        6a. verify that there's a pending transaction with a mined height of sentTxHeight
+//        */
+//        LoggerProxy.info("6a. verify that there's a pending transaction with a mined height of \(sentTxHeight)")
+//        XCTAssertEqual(afterStagePendingTx.minedHeight, sentTxHeight)
+//        XCTAssertNotNil(afterStagePendingTx.minedHeight, "pending transaction shown as unmined when it has been mined")
+//        XCTAssertTrue(afterStagePendingTx.isPending(currentHeight: sentTxHeight))
```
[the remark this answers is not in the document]
nuttycom: I reopened #1247

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -131,9 +120,9 @@ public protocol Synchronizer: AnyObject {
     /// do not already exist). These files can be given a prefix for scenarios where multiple wallets
     ///
     /// - Parameters:
-    ///   - seed: ZIP-32 Seed bytes for the wallet that will be initialized.
-    ///   - viewingKeys: Viewing key derived from seed.
+    ///   - seed: ZIP-32 Seed bytes for the wallet that will be initialized
     ///   - walletBirthday: Birthday of wallet.
+    ///   - walletMode: Set `.newWallet` when preparing synchronizer for a brand new generated wallet, `.restoreWallet` when wallet is about to be restored from a seed and  `.existingWallet` for all other scenarios.
     /// - Throws:
     ///     - `aliasAlreadyInUse` if the Alias used to create this instance is already used by other instance.
```
pacu, line 127: is this still being used? I don't see where the alias can be provided

Tests/DarksideTests/AdvancedReOrgTests.swift
```diff
@@ -441,8 +451,9 @@ class AdvancedReOrgTests: ZcashTestCase {
 
         let expectedVerifiedBalance = initialTotalBalance + pendingTx.value
         let currentVerifiedBalance = try await coordinator.synchronizer.getShieldedVerifiedBalance()
-        let expectedPendingTransactionsCount = await coordinator.synchronizer.pendingTransactions.count
-        XCTAssertEqual(expectedPendingTransactionsCount, 0)
+        // TODO: [#1247] needs to review this to properly solve, https://github.com/zcash/ZcashLightClientKit/issues/1247
+//        let expectedPendingTransactionsCount = await coordinator.synchronizer.pendingTransactions.count
+//        XCTAssertEqual(expectedPendingTransactionsCount, 0)
```
pacu, line 456: How does the SDK track sent and unconfirmed transactions?

Tests/DarksideTests/DarksideSanityCheckTests.swift
```diff
@@ -72,13 +79,14 @@ class DarksideSanityCheckTests: ZcashTestCase {
...
-        
-        let blocksDao = BlockSQLDAO(dbProvider: SimpleConnectionProvider(path: coordinator.databases.dataDB.absoluteString, readonly: false))
-        
-        let firstBlock = try blocksDao.block(at: expectedFirstBlock.height)
-        let lastBlock = try blocksDao.block(at: expectedLastBlock.height)
-        
-        XCTAssertEqual(firstBlock?.hash.toHexStringTxId(), expectedFirstBlock.hash)
-        XCTAssertEqual(lastBlock?.hash.toHexStringTxId(), expectedLastBlock.hash)
+
+        // TODO: [#1247] needs to review this to properly solve, https://github.com/zcash/ZcashLightClientKit/issues/1247
```
pacu, line 83: The idea of this test is just syncing and tracking that you haven't missed a block. it's the most basic darksidewalletd test to have some common ground on saying "ok, simple things work" and take it from there

pacu (commented):
image
can't reject. :/


Reviewed the files and left questions and change request.

Package.resolved
```diff
@@ -1,5 +1,14 @@
 {
   "pins" : [
+    {
```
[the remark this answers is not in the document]
LukasKorba: Since it's happening only to you locally it can't be an issue globally. It's a combination of xcode, caches and derived data. Try to clean derived data and load project from scratch. https://www.swiftanytime.com/blog/how-to-delete-derived-data-in-xcode

Sources/ZcashLightClientKit/Block/Actions/UpdateSubtreeRootsAction.swift
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+extension UpdateSubtreeRootsAction: Action {
+    var removeBlocksCacheWhenFailed: Bool { false }
+
+    func run(with context: ActionContext, didUpdate: @escaping (CompactBlockProcessor.Event) async -> Void) async throws -> ActionContext {
+        var request = GetSubtreeRootsArg()
+        request.shieldedProtocol = .sapling
+        request.maxEntries = await configProvider.config.network.networkType == .testnet ? 65536 : 0
```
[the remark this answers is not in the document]
LukasKorba: Maybe this is no longer needed but at the beginning we were supposed to use 655536 for testnet only so we created this condition. By we I mean both ios and android, I see Honza still has the same condition in his code

```
maxEntries = if (network.isTestnet()) {
                        65536
                    } else {
                        0
                    },
```

Sources/ZcashLightClientKit/Block/Actions/UpdateSubtreeRootsAction.swift
```diff
@@ -0,0 +1,59 @@
...
+            for try await subtreeRoot in stream {
+                roots.append(subtreeRoot)
+            }
+        } catch ZcashError.serviceSubtreeRootsStreamFailed(LightWalletServiceError.timeOut) {
+            throw ZcashError.serviceSubtreeRootsStreamFailed(LightWalletServiceError.timeOut)
+        }
+
+        logger.info("Sapling tree has \(roots.count) subtrees")
+        do {
+            try await rustBackend.putSaplingSubtreeRoots(startIndex: UInt64(request.startIndex), roots: roots)
```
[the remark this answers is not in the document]
LukasKorba: Yes, it's a default

```swift
struct GetSubtreeRootsArg {
  // SwiftProtobuf.Message conformance is added in an extension below. See the
  // `Message` and `Message+*Additions` files in the SwiftProtobuf library for
  // methods supported on all messages.

  /// Index identifying where to start returning subtree roots
  var startIndex: UInt32 = 0

  /// Shielded protocol to return subtree roots for
  var shieldedProtocol: ShieldedProtocol = .sapling

  /// Maximum number of entries to return, or 0 for all entries.
  var maxEntries: UInt32 = 0

  var unknownFields = SwiftProtobuf.UnknownStorage()

  init() {}
}
```

Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
LukasKorba: That is good catch. The method was still "used" but under condition of `ZcashError.rustValidateCombinedChainInvalidChain` and that one is no longer triggered in the whole SDK. So removed related code.

Sources/ZcashLightClientKit/Block/Download/BlockDownloader.swift
LukasKorba: I removed the progress completely, it was supposed to be cleaned up in https://github.com/zcash/ZcashLightClientKit/pull/1254

Sources/ZcashLightClientKit/Block/Enhance/BlockEnhancer.swift
LukasKorba: same answer as above

Sources/ZcashLightClientKit/Block/FetchUnspentTxOutputs/UTXOFetcher.swift
LukasKorba: same answer as above

Sources/ZcashLightClientKit/Block/Scan/BlockScanner.swift
LukasKorba: What do you mean derived from scan ranges?

This commented line does inside

```
        guard network == .mainnet else { return UInt32(config.scanningBatchSize) }

        if height > 1_650_000 {
            // librustzcash thread saturation at a number of blocks
            // that contains 100 * num_cores Sapling outputs.
            return UInt32(max(ProcessInfo().activeProcessorCount, 10))
        }

        return UInt32(config.scanningBatchSize)

```

so it's either 10 or 100 (for config.scanningBatchSize). The condition that decides that is about height being either above or bellow 1_650_000 blocks. If 100 is now valid for any height, let's get rid of this (not my code so I don't know much about it).

Sources/ZcashLightClientKit/Block/Scan/BlockScanner.swift
LukasKorba: Created a followup ticket https://github.com/zcash/ZcashLightClientKit/issues/1259

Sources/ZcashLightClientKit/Block/Scan/BlockScanner.swift
LukasKorba: removed completely

Sources/ZcashLightClientKit/CombineSynchronizer.swift
LukasKorba: followup ticket https://github.com/zcash/ZcashLightClientKit/issues/1260

Sources/ZcashLightClientKit/Initializer.swift
LukasKorba: updated the interface and code

Sources/ZcashLightClientKit/Metrics/SDKMetrics.swift
LukasKorba: removed completely

Sources/ZcashLightClientKit/Repository/TransactionRepository.swift
LukasKorba: cleaned up

Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift
LukasKorba: resolved a few comments above

Sources/ZcashLightClientKit/Block/Actions/DownloadAction.swift
```diff
@@ -30,27 +30,28 @@ extension DownloadAction: Action {
...
 
         let batchRange = batchRangeStart...batchRangeEnd
-        let downloadLimit = batchRange.upperBound + (2 * config.batchSize)
+        let potentialDownloadLimit = batchRange.upperBound + (2 * config.batchSize)
+        let downloadLimit = await context.syncControlData.latestBlockHeight >= potentialDownloadLimit ? potentialDownloadLimit : batchRangeEnd
 
         logger.debug("Starting download with range: \(batchRange.lowerBound)...\(batchRange.upperBound)")
-        await downloader.update(latestDownloadedBlockHeight: batchRange.lowerBound)
+        
+        await downloader.update(latestDownloadedBlockHeight: batchRange.lowerBound, force: true) // SbS
```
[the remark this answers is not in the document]
LukasKorba: Rather `batchRangeStart`, refactored a bit this code + cleaned up // SbS comment

Sources/ZcashLightClientKit/Block/Actions/DownloadAction.swift
```diff
@@ -30,27 +30,28 @@ extension DownloadAction: Action {
     var removeBlocksCacheWhenFailed: Bool { true }
 
     func run(with context: ActionContext, didUpdate: @escaping (CompactBlockProcessor.Event) async -> Void) async throws -> ActionContext {
-        guard let lastScannedHeight = await context.syncControlData.latestScannedHeight else {
+        guard let lastScannedHeight = await context.lastScannedHeight else {
```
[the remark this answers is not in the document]
LukasKorba: All State Machine actions do only their purpose. In case of DownloadAction, that one is purely downloading the given range and that can be called a linear downloading, yes. The scan ranges are decided somewhere else so input for this action can be a range of just 1 block, 5 blocks, 10k blocks or even 400k blocks and this action will simply process what is asked to do.

Sources/ZcashLightClientKit/Resources/checkpoints/mainnet/2092500.json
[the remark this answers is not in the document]
LukasKorba: created a follow up https://github.com/zcash/ZcashLightClientKit/issues/1261

Tests/TestUtils/proto/darkside.pb.swift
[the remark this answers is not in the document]
LukasKorba: In that case also `compact_formats.pb.swift` and `service.pb.swift`

Sources/ZcashLightClientKit/Block/Actions/ProcessSuggestedScanRangesAction.swift
```diff
@@ -0,0 +1,81 @@
...
+extension ProcessSuggestedScanRangesAction: Action {
+    var removeBlocksCacheWhenFailed: Bool { false }
+
+    func run(with context: ActionContext, didUpdate: @escaping (CompactBlockProcessor.Event) async -> Void) async throws -> ActionContext {
+        logger.info("Getting the suggested scan ranges from the wallet database.")
+        let scanRanges = try await rustBackend.suggestScanRanges()
+
+        if let firstRange = scanRanges.first {
+            let lowerBound = firstRange.range.lowerBound - 1
+            let upperBound = firstRange.range.upperBound - 1
```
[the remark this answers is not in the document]
LukasKorba: renamed to `rangeStartExclusive` and `rangeEndInclusive` so it's clear what these two represents.

Sources/ZcashLightClientKit/Block/Actions/ProcessSuggestedScanRangesAction.swift
```diff
@@ -0,0 +1,81 @@
...
+                
+                logger.info("Setting the total range for Spend before Sync to \(minHeight...maxHeight).")
+                await context.update(totalProgressRange: minHeight...maxHeight)
+            }
+
+            // If there is a range of blocks that needs to be verified, it will always
+            // be returned as the first element of the vector of suggested ranges.
+            if firstRange.priority == .verify {
+                await context.update(requestedRewindHeight: lowerBound + 1)
+                await context.update(state: .rewind)
```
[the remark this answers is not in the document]
LukasKorba: Short answer, I've done it the same way as Honza did. When I returned from my vacation he was ahead of me and after meeting him this was the outcome of the call. When `verify` is suggested, we rewind to the start of the range.

Android
```
 // Running synchronization for the [ScanRange.SuggestScanRangePriority.Verify] range
        while (verifyRangeResult is VerifySuggestedScanRange.ShouldVerify) {
            Twig.info { "Starting verification of range: $verifyRangeResult" }

            // Remove existing blocks as they'll be re-downloaded
            downloader.rewindToHeight(verifyRangeResult.scanRange.range.start)
            deleteAllBlockFiles(
                downloader = downloader,
                lastKnownHeight = verifyRangeResult.scanRange.range.start
            )
```

Question now is what is the correct approach. If this solution is "good enough" for the RC, I would leave it and resolved it in a follow up ticket. I think major change now of the State Machine could postpone the RC.

Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
```diff
@@ -616,9 +628,8 @@ extension CompactBlockProcessor {
 
     private func resetContext() async {
         let lastEnhancedheight = await context.lastEnhancedHeight
-        context = ActionContext(state: .idle)
+        context = ActionContextImpl(state: .idle)
         await context.update(lastEnhancedHeight: lastEnhancedheight)
-        await compactBlockProgress.reset()
```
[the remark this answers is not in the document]
LukasKorba: Please note this is not a CompactBlockProcessor but CompactBlockProgress, it got simplified significantly so it does no longer need to reset anything. See https://github.com/zcash/ZcashLightClientKit/pull/1254

Sources/ZcashLightClientKit/Block/Actions/Action.swift
```diff
@@ -7,11 +7,40 @@
...
+    func update(lastChainTipUpdateTime: TimeInterval) async
+    func update(lastScannedHeight: BlockHeight) async
+    func update(lastDownloadedHeight: BlockHeight) async
+    func update(lastEnhancedHeight: BlockHeight?) async
+    func update(requestedRewindHeight: BlockHeight) async
+}
+
+actor ActionContextImpl: ActionContext {
     var state: CBPState
     var prevState: CBPState?
```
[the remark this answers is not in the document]
LukasKorba: First of all it's for an internal use only so clients or public API is no affected. Second of all, the use is simple
```swift
if await context.prevState == .idle {
```

because it's an actor. So noAction is no needed

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -409,35 +409,30 @@ public class Initializer {
     /// - Parameter seed: ZIP-32 Seed bytes for the wallet that will be initialized
     /// - Throws: `InitializerError.dataDbInitFailed` if the creation of the dataDb fails
     /// `InitializerError.accountInitFailed` if the account table can't be initialized. 
-    func initialize(with seed: [UInt8]?, viewingKeys: [UnifiedFullViewingKey], walletBirthday: BlockHeight) async throws -> InitializationResult {
+    func initialize(with seed: [UInt8]?, walletBirthday: BlockHeight, for walletMode: WalletInitMode) async throws -> InitializationResult {
```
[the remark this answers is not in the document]
LukasKorba: Answer bellow

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -409,35 +409,30 @@ public class Initializer {
...
         try await storage.create()
 
         if case .seedRequired = try await rustBackend.initDataDb(seed: seed) {
             return .seedRequired
```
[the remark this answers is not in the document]
LukasKorba: This is interesting point in general. @nuttycom I probably need some help here.

1. I don't know why seed is an optional here, what would be the use case of no seed at all? So instead of guarding the seed, we should probably require the seed. @pacu is there a reason why this has been an option for ages? some use case?

2. I don't see a combination that is invalid in terms of given seed vs. walletMode

Sources/ZcashLightClientKit/Model/ScanProgress.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+
+import Foundation
+
+struct ScanProgress: Equatable {
+    let numerator: UInt64
+    let denominator: UInt64
+    
+    func progress() throws -> Float {
+        let value = Float(numerator) / Float(denominator)
```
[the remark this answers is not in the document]
LukasKorba: This is by purpose. ScanProgress is initialized internally with this code

```swift
        if result.denominator == 0 {
            switch result.numerator {
            case 0:
                return nil
            default:
                throw ZcashError.rustGetScanProgress(lastErrorMessage(fallback: "`getScanProgress` failed with unknown error"))
            }
        } else {
            return ScanProgress(numerator: result.numerator, denominator: result.denominator)
        }

```

so division by 0 is not possible.

Sources/ZcashLightClientKit/Repository/TransactionRepository.swift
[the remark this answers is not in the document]
LukasKorba: already commented by Kris, resolved

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -521,8 +436,136 @@ actor ZcashRustBackend: ZcashRustBackendWelding {
...
+                    completing_block_height: UInt32(root.completingBlockHeight)
+                )
+            )
+        }
+
+        var contiguousFfiRoots = ContiguousArray(ffiSubtreeRootsVec)
+
+        let len = UInt(contiguousFfiRoots.count)
+
+        let rootsPtr = UnsafeMutablePointer<FfiSubtreeRoots>.allocate(capacity: 1)
```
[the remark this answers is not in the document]
LukasKorba: This is very likely true but I see some other allocations that seem to me miss dealloc calls. Created the follow up ticket to put more time into this topic https://github.com/zcash/ZcashLightClientKit/issues/1262

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -242,10 +228,6 @@ public protocol Synchronizer: AnyObject {
     /// - Returns: an array with the given Transactions or an empty array
     func allTransactions(from transaction: ZcashTransaction.Overview, limit: Int) async throws -> [ZcashTransaction.Overview]
 
-    /// Fetch all pending transactions
-    /// - Returns: an array of transactions which are considered pending confirmation. can be empty
-    func allPendingTransactions() async throws -> [ZcashTransaction.Overview]
```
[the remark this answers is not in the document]
LukasKorba: For now `allTransactions` does the job because every transaction has a state

```swift 
        public enum State {
            /// transaction has a `minedHeight` that's greater or equal than
            /// `ZcashSDK.defaultStaleTolerance` confirmations.
            case confirmed
            /// transaction has no `minedHeight` but current known height is less than
            /// `expiryHeight`.
            case pending
            /// transaction has no
            case expired
```

so it's easy to filter it out. Follow up [ticket](https://github.com/zcash/ZcashLightClientKit/issues/1260) could bring allPendingTransactions back.

Sources/ZcashLightClientKit/Synchronizer.swift
LukasKorba: comment updated so it's clear, the alias is `for:` actually

Tests/DarksideTests/AdvancedReOrgTests.swift
LukasKorba: Short answer, this has been removed even from Android's SDK long time ago. May 5th actually, see https://github.com/zcash/zcash-android-wallet-sdk/pull/990/files

Same happened in iOS now, mostly because ZcashTransaction now holds all 3 states, same as on Android

```
- Synchronizer APIs for listing pending transactions have been removed, along with the `PendingTransaction` object.
- `Synchronizer.clearedTransactions` has been renamed to `Synchronizer.transactions` and includes sent, received, and pending transactions.
```

Updating DarkSideTests will take time.

Example/ZcashLightClientSample/ZcashLightClientSample/ViewController.swift
```diff
@@ -82,13 +82,7 @@ class MainTableViewController: UITableViewController {
     
     override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
         if let destination = segue.destination as? TransactionsTableViewController {
-            if let id = segue.identifier, id == "Pending" {
```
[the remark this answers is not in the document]
LukasKorba: It's connected to the synchronizer's allPendingTransactions(). Removed from the UI for now, once https://github.com/zcash/ZcashLightClientKit/issues/1260 is done, we can bring it back

Sources/ZcashLightClientKit/Block/Actions/ProcessSuggestedScanRangesAction.swift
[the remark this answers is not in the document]
LukasKorba: UPD: I think we can just delete this rewind from `processSuggestedScanRanges` action and let ScanAction to handle it with the following code:

```swift
            if isContinuityError(errorMsg) {
                await context.update(requestedRewindHeight: batchRange.lowerBound - 10)
                await context.update(state: .rewind)
                return context
            } else {
                throw ZcashError.rustScanBlocks(errorMsg)
            }
```

Sources/ZcashLightClientKit/Model/ScanRange.swift
```diff
@@ -0,0 +1,27 @@
...
+        case scanned = 10
+        case historic = 20
+        case openAdjacent = 30
+        case foundNote = 40
+        case chainTip = 50
+        case verify = 60
+        
+        init(_ value: UInt8) {
+            self = Priority(rawValue: value) ?? .ignored
+        }
```
str4d, line 22: Hmm, I'm not happy about this resolution (just renaming `.unknown` to `.ignored`, because is `value > 10` then it is an unscanned range and needs to be scanned, not ignored.

The ideal case is that unknown priority values are preserved as they are (so maybe having a `case unknown(UInt8)` or something similar). If Swift can't do that, then we should promote unknown priority values to the nearest priority above them.

Alternatively, given that these numeric values come from the FFI layer, we could treat unknown priority values as a programming error, and do whatever the Swift equivalent of a panic is here.

Sources/ZcashLightClientKit/Block/Actions/DownloadAction.swift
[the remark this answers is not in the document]
str4d: Please open an issue for renaming `context.lastScannedHeight` etc. to better indicate what they are used for, because `lastScannedHeight` is now an inaccurate name in this usage. We should probably do that rename between 2.0.0-rc.1 and 2.0.0, as it may unearth bugs.

Sources/ZcashLightClientKit/Block/Actions/RewindAction.swift
```diff
@@ -0,0 +1,48 @@
...
+
+    init(container: DIContainer) {
+        downloader = container.resolve(BlockDownloader.self)
+        rustBackend = container.resolve(ZcashRustBackendWelding.self)
+        downloaderService = container.resolve(BlockDownloaderService.self)
+        logger = container.resolve(Logger.self)
+    }
+    
+    private func update(context: ActionContext) async -> ActionContext {
+        await context.update(state: .download)
```
str4d, line 24: Rewinding can invalidate / change scan ranges, so this needs to be:
```suggestion
        await context.update(state: .processSuggestedScanRanges)
```

str4d: Or if `.clearCache` is what everything needs to go through to get there:
```suggestion
        await context.update(state: .clearCache)
```

Sources/ZcashLightClientKit/Block/Scan/BlockScanner.swift
nuttycom: Hm, interesting. This might be a workaround for long scan times processing 100-block batches sandblasted blocks; I suspect this is why Pacu asked for https://github.com/zcash/librustzcash/issues/704 so let's leave this as it is for now.

Sources/ZcashLightClientKit/Block/Actions/DownloadAction.swift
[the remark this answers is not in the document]
LukasKorba: https://github.com/zcash/ZcashLightClientKit/issues/1263

Sources/ZcashLightClientKit/Block/Actions/UpdateChainTipAction.swift
```diff
@@ -0,0 +1,52 @@
...
+
+extension UpdateChainTipAction: Action {
+    var removeBlocksCacheWhenFailed: Bool { false }
+
+    func run(with context: ActionContext, didUpdate: @escaping (CompactBlockProcessor.Event) async -> Void) async throws -> ActionContext {
+        let lastChainTipUpdateTime = await context.lastChainTipUpdateTime
+        let now = Date().timeIntervalSince1970
+
+        // Update chain tip can be called from different contexts
+        if await context.prevState == .updateSubtreeRoots || now - lastChainTipUpdateTime > 600 {
```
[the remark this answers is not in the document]
LukasKorba: https://github.com/zcash/ZcashLightClientKit/issues/1264

Sources/ZcashLightClientKit/Block/Actions/ProcessSuggestedScanRangesAction.swift
```diff
@@ -0,0 +1,74 @@
...
+                """)
+
+            await context.update(lastScannedHeight: rangeStartExclusive)
+            await context.update(lastDownloadedHeight: rangeStartExclusive)
+            await context.update(syncControlData: syncControlData)
+
+            // the total progress range is computed only for the first time
+            // as a sum of all ranges
+            let totalProgressRange = await context.totalProgressRange
+            if totalProgressRange.lowerBound == 0 && totalProgressRange.upperBound == 0 {
```
nuttycom, line 53: Now that I have more context, this seems like it could be severely sub-optimal because you could have a very small un-scanned range hundreds of thousands of blocks ago, and a very small un-scanned range at the chain tip, and it would try to download all of the scanned blocks in between these two? Am I misunderstanding here?

nuttycom: Okay, I now understand that the `totalProgressRange` is just for reporting, and is in fact now unused and can be removed.

Sources/ZcashLightClientKit/Model/ScanRange.swift
LukasKorba: Resolved, added a logic that tries to find the matching priorities, otherwise uses the 1 above the value.

LukasKorba: Resolved, added a logic that tries to find the matching priorities, otherwise uses the 1 above the value.

Sources/ZcashLightClientKit/Block/Actions/Action.swift
```diff
@@ -25,18 +52,27 @@ actor ActionContext {
         self.state = state
     }
     func update(syncControlData: SyncControlData) async { self.syncControlData = syncControlData }
-    func update(totalProgressRange: CompactBlockRange) async { self.totalProgressRange = totalProgressRange }
+    func update(totalProgressRange: CompactBlockRange) async {
```
nuttycom, line 55: Can this now be removed, since `totalProgressRange` has been removed elsewhere?

LukasKorba: removed

Sources/ZcashLightClientKit/Model/ScanRange.swift
```diff
@@ -18,7 +18,19 @@ struct ScanRange {
...
+            } else {
+                // no match found, getting the one above the value
+                let valueNeeded = (value / 10) * 10 + 10
+                
+                if let priority = Priority(rawValue: valueNeeded) {
+                    self = priority
+                } else {
+                    fatalError("The value \(value) is out of the range of priorities.")
+                }
+            }
```
nuttycom, line 33: I don't think that coercing here is the right solution; this is an enumeration, so failing to recognize the exact value _should_ just be a fatal error. So the previous implementation was correct, but instead of defaulting to `.ignored` it should raise the error.

LukasKorba: Only the matching priorities are allowed, otherwise fatalError

Sources/ZcashLightClientKit/Model/ScanRange.swift
```diff
@@ -0,0 +1,31 @@
...
+        case openAdjacent = 30
+        case foundNote = 40
+        case chainTip = 50
+        case verify = 60
+        
+        init(_ value: UInt8) {
+            if let priority = Priority(rawValue: value) {
+                self = priority
+            } else {
+                fatalError("The value \(value) is out of the range of priorities.")
```
str4d, line 24: I'm happy with this resolution; it is now a programming error to change `zcash-light-client-ffi` to pass over additional priority values, without also updating this logic.

str4d (approved):
utACK 26fe1f7310817248148bb454de65d085063e21e0

nuttycom:
utACK https://github.com/zcash/ZcashLightClientKit/commit/26fe1f7310817248148bb454de65d085063e21e0

Sources/ZcashLightClientKit/Model/ScanProgress.swift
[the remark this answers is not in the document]
pacu: I imagined it was the case, nonetheless it doesn't mean that the type doesn't have the bug locally. 

Anyone who lacks the context could craft a change that could pass a value that would make the code crash. The code could be refactored and the bug easily introduced unknowingly.

For example:

This PR has 6K LoCs added and 5K LoCs removed across 193 files. A change that triggers the division by zero could be added (knowingly or unknowingly) in a future PR of this size. If the reviewer(s) miss such side-effect, the code PR will be approved and the code will make it to production. Moreover if the code that uses this struct is input from a library (the FFI), it adds a layer of indirection that makes it a great candidate for someone looking vectors to do some damage.

pacu (commented):
changes are non blocking for an RC, but I would pay specific attention to the Division By Zero and the allocations on the RustBackend not leaking

nuttycom:
Closing as complete with the 2.0.0 release.

PR #1275, opened 2023-09-18
Update zcash-light-client-ffi to adopt fix for missing transparent history

nuttycom:

# Author

# Reviewer


Tests/TestUtils/TestDbBuilder.swift
```diff
@@ -54,10 +54,8 @@ enum TestDbBuilder {
         Bundle.module.url(forResource: "darkside_caches", withExtension: "db")
     }
     
-    static func prepopulatedDataDbProvider(rustBackend: ZcashRustBackendWelding) async throws -> ConnectionProvider? {
-        guard let url = prePopulatedMainnetDataDbURL() else { return nil }
-
-        let provider = SimpleConnectionProvider(path: url.absoluteString, readonly: true)
+    static func prepopulatedDataDbProvider(rustBackend: ZcashRustBackend) async throws -> ConnectionProvider? {
+        let provider = SimpleConnectionProvider(path: (await rustBackend.dbData).0, readonly: true)
```
nuttycom, line 58: Something I don't understand here; why does it appear that I have to use `.0` instead of `.absoluteString`?

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
```diff
@@ -73,7 +65,7 @@ class TransactionSQLDAO: TransactionRepository {
 
     func find(offset: Int, limit: Int, kind: TransactionKind) async throws -> [ZcashTransaction.Overview] {
         let query = transactionsView
-            .order((ZcashTransaction.Overview.Column.minedHeight ?? BlockHeight.max).desc, ZcashTransaction.Overview.Column.id.desc)
```
nuttycom, line 76: We could also order by transaction index here, but since that will be null for received transparent transactions I'm not sure how much value it has. If the goal is just ordering stability in the UI, we could have the secondary ordering be by txid, but TBH I don't think it matters much; it's an edge case that the wallet both sends and receives distinct transactions within the same block.

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
```diff
@@ -96,16 +88,17 @@ class TransactionSQLDAO: TransactionRepository {
...
-                (ZcashTransaction.Overview.Column.minedHeight ?? BlockHeight.max).desc, ZcashTransaction.Overview.Column.id.desc
-            )
+            .order((ZcashTransaction.Overview.Column.minedHeight ?? BlockHeight.max).desc)
             .filter(
-                Int64(transactionBlockTime) > ZcashTransaction.Overview.Column.blockTime
-                && transactionIndex > ZcashTransaction.Overview.Column.index
+                transactionBlockHeight > ZcashTransaction.Overview.Column.minedHeight
+                || (
+                    transactionBlockHeight == ZcashTransaction.Overview.Column.minedHeight &&
+                    transactionIndex > (ZcashTransaction.Overview.Column.index ?? -1)
```
nuttycom, line 100: After further consideration, the ordering on `transactionIndex` doesn't make any sense at all here; blocks are internally not ordered by index, and even with transparent funds a transaction spending zero-conf funds could appear before the outputs that it spends within the block. So this is really useless; we should order only by block height.

str4d: > blocks are internally not ordered by index, and even with transparent funds a transaction spending zero-conf funds could appear before the outputs that it spends within the block.

This is false. There is a consensus rule that requires transactions within a block to be ordered such that you never observe children before their parents. More specifically, it's the combination of this check (that a transaction's transparent inputs are available):
https://github.com/zcash/zcash/blob/c908d4bb9ae3be92da14c81ff7a6ddec904c9226/src/main.cpp#L3247-L3249
with this code that adds a transaction's outputs to the view making them available:
https://github.com/zcash/zcash/blob/c908d4bb9ae3be92da14c81ff7a6ddec904c9226/src/main.cpp#L3367
and the fact that we iterate over a block's transactions in index order (making the above deterministic).

nuttycom: Okay. In any case, the filter that was previously on line 108 was a bug that could result in nondeterministically-incomplete transaction history.

nuttycom: We can keep this new implementation for now, but we should revisit whether it actually makes sense.

Sources/ZcashLightClientKit/Entity/TransactionEntity.swift
```diff
@@ -25,9 +25,9 @@ public enum ZcashTransaction {
...
                 currentHeight: BlockHeight,
                 minedHeight: BlockHeight?,
-                expiredUnmined: Bool
+                expiredUnmined: Bool?
             ) {
-                guard !expiredUnmined else {
+                guard let expiredUnmined, !expiredUnmined else {
                     self = .expired
                     return
                 }
```
str4d, line 33: The logic just below here likely has the same off-by-one error as I just fixed in the Android SDK (`currentHeight - minedHeight` is zero-indexed, but confirmations are 1-indexed, so this should be `currentHeight + 1 - minedHeight`).

Tests/OfflineTests/TransactionRepositoryTests.swift
```diff
@@ -14,7 +14,11 @@ class TransactionRepositoryTests: XCTestCase {
 
     override func setUp() async throws {
         try await super.setUp()
-        let rustBackend = ZcashRustBackend.makeForTests(fsBlockDbRoot: Environment.uniqueTestTempDirectory, networkType: .testnet)
+        let rustBackend = ZcashRustBackend.makeForTests(
+            dbData: TestDbBuilder.prePopulatedMainnetDataDbURL()!,
```
str4d, line 18: Does the database file at this URL get copied before being used, or does this result in the actual database file checked into the repository being modified?

Tests/OfflineTests/TransactionRepositoryTests.swift
```diff
@@ -14,7 +14,11 @@ class TransactionRepositoryTests: XCTestCase {
...
+            fsBlockDbRoot: Environment.uniqueTestTempDirectory,
+            networkType: .mainnet
```
str4d, line 20: Huh, how on earth was this set of tests working before...

Tests/OfflineTests/TransactionRepositoryTests.swift
```diff
@@ -189,24 +193,25 @@ class TransactionRepositoryTests: XCTestCase {
     }
     
     func testFindAllFrom() async throws {
-        let transaction = try await self.transactionRepository.find(id: 16)
+        let rawID = Data(fromHexEncodedString: "5d9b91e31a6d3f94844a4c330e727a2d5d0643f6caa6c75573b28aefe859e8d2")!
+        let transaction = try await self.transactionRepository.find(rawID: rawID)
         let transactionsFrom = try await self.transactionRepository.find(from: transaction, limit: Int.max, kind: .all)
 
-        XCTAssertEqual(transactionsFrom.count, 8)
+        XCTAssertEqual(transactionsFrom.count, 15)
```
str4d, line 200: Why does this number increase?

I'm somewhat confused about why it was 8 in the first place; `darkside_data.db` only contains 21 transactions, and the view returns one row per transaction (because we group by `(account, txid)` and the database only has one account), so with this being row 16 (and the row IDs having the same ordering as by block height in this database), I would expect this number to be one of (depending on whether `from` is included in the returned list):
- If this finds transactions _earlier than_ `from`: 15 or 16
- If this finds transactions _later than_ `from`: 5 or 6

The fact that this is being changed to 15 suggests that `transactionRepository.find(from)` is after this PR returning all transactions _earlier than_ `from`, rather than any of the other three options. The tests below this point do seem to indicate that prior to this PR that was the intention (i.e. this is a reverse-chronological API), but was it not doing that?

Tests/OfflineTests/TransactionRepositoryTests.swift
```diff
@@ -189,24 +193,25 @@ class TransactionRepositoryTests: XCTestCase {
...
 
         transactionsFrom.forEach { preceededTransaction in
-            guard let preceededTransactionIndex = preceededTransaction.index, let transactionIndex = transaction.index else {
+            guard let precedingHeight = preceededTransaction.minedHeight, let transactionHeight = transaction.minedHeight else {
                 XCTFail("Transactions are missing indexes.")
```
str4d, line 204:
```suggestion
                XCTFail("Transactions are missing mined heights.")
```

Tests/OfflineTests/TransactionRepositoryTests.swift
```diff
@@ -189,24 +193,25 @@ class TransactionRepositoryTests: XCTestCase {
...
-            guard let preceededTransactionBlockTime = preceededTransaction.blockTime, let transactionBlockTime = transaction.blockTime else {
+            guard let precedingBlockTime = preceededTransaction.blockTime, let transactionBlockTime = transaction.blockTime else {
                 XCTFail("Transactions are missing block time.")
                 return
             }
 
-            XCTAssertLessThan(preceededTransactionIndex, transactionIndex)
-            XCTAssertLessThan(preceededTransactionBlockTime, transactionBlockTime)
+            XCTAssertLessThanOrEqual(precedingHeight, transactionHeight)
+            XCTAssertLessThan(precedingBlockTime, transactionBlockTime)
```
str4d, line 214: The heights cannot be equal if the block times are not allowed to be equal. Moreover, there is nothing that requires block times to be monotonically increasing, so `precedingBlockTime < transactionBlockTime` can't hold generally.

I've checked the `transactions` table of `darkside_data.db`, and if this API were to return transactions _later than_ `from` then there is a pair of transactions in `darkside_data.db` that were mined in the same block. However, per above this API appears to be returning transactions _earlier than_ (and not including) `from`, so it might happen that "less than" holds as necessary (but in that case it should also be "less than" for heights).

Example/ZcashLightClientSample/ZcashLightClientSample/List Transactions/TransactionsDataSource.swift
```diff
@@ -37,8 +37,10 @@ class TransactionsDataSource: NSObject {
         case .cleared:
             let rawTransactions = await synchronizer.transactions
             for transaction in rawTransactions {
-                let memos = try await synchronizer.getMemos(for: transaction)
-                transactions.append(TransactionDetailModel(transaction: transaction, memos: memos))
+                if transaction.minedHeight != nil {
```
str4d, line 40: Does `lightwalletd`'s transaction-fetching API not search the mempool?

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
```diff
@@ -73,7 +65,7 @@ class TransactionSQLDAO: TransactionRepository {
 
     func find(offset: Int, limit: Int, kind: TransactionKind) async throws -> [ZcashTransaction.Overview] {
         let query = transactionsView
-            .order((ZcashTransaction.Overview.Column.minedHeight ?? BlockHeight.max).desc, ZcashTransaction.Overview.Column.id.desc)
+            .order((ZcashTransaction.Overview.Column.minedHeight ?? BlockHeight.max).desc)
```
str4d, line 68: The Android SDK now orders secondarily by transaction index (which happens to be stable even with it being `null` for transparent-only transactions, because we never fetch this information for transparent-only transactions). Non-blocking, it probably doesn't matter that transactions within a single block don't have a stable ordering between Android and iOS UIs (and the old ordering was not stable across recovery-from-seed).

str4d (approved):
utACK 404eff21b104afb602a7bf33b4ac8158f470a292, identified issues are in tests.

Tests/OfflineTests/TransactionRepositoryTests.swift
str4d: Reading the database lookup later on in my review, I confirmed that it is returning reverse-height order, and that it excludes `from`. I also confirmed that the previous filtering mechanic was running afoul of the non-monotonically-increasing nature of block timestamps, which is likely why some transaction history was being omitted. It is correct that windowing should only be done via height, and any date-based ordering should be handled at the UI layer.

Tests/OfflineTests/TransactionRepositoryTests.swift
nuttycom: I have not observed any diff to the checked-in database file. In this test in particular, I believe that the database is used in a read-only fashion.

Tests/OfflineTests/TransactionRepositoryTests.swift
nuttycom: The reason that this changed is the index bug - all transactions later than the height of the first with lower `tx_index` values were being excluded. I verified the correct value by looking at the sqlite database; I suspect that the previous '8' value was derived from the code under test.

Tests/OfflineTests/TransactionRepositoryTests.swift
nuttycom: Apparently none of the migrations were relevant to this test until now.

str4d (approved):
Re-utACK a338e1daf5032fee507c069635bc875de6aada6a

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
daira: Hmm. I would consider stability of ordering in the UI quite important. It's not critical for this release, but please file an issue.

Sources/ZcashLightClientKit/Entity/TransactionEntity.swift
```diff
@@ -144,7 +143,6 @@ extension ZcashTransaction.Output {
 extension ZcashTransaction.Overview {
     enum Column {
         static let accountId = Expression<Int>("account_id")
-        static let id = Expression<Int>("id_tx")
         static let minedHeight = Expression<BlockHeight?>("mined_height")
         static let index = Expression<Int?>("tx_index")
         static let rawID = Expression<Blob>("txid")
```
daira, line 148: The order in librustzcash is `txid` then `tx_index`.

daira (commented):
Post-hoc review: this may be a significant bug. https://github.com/zcash/ZcashLightClientKit/pull/1275/files#r1330793426

Sources/ZcashLightClientKit/Entity/TransactionEntity.swift
nuttycom: The ordering doesn't matter.

daira: Are you sure? This is *exactly* what I'd expect if it did:
> I looked at the offline tests, fixed 2, 2 more commented out for now.. but I think the setup is wrong, I see in the DB that the txids are 1, 2, …21 but tests expect txid to be hash of 32 bytes long… so the test rawIDs don’t match mocked transaction data

daira: Ok I looked at the [implementation of Expression](https://github.com/stephencelis/SQLite.swift/blob/d34f63a90a153a8c54ea743f88978faf83d01a91/Sources/SQLite/Typed/Expression.swift#L56) and it's just a wrapper for a `String`, so I don't see how the order could matter.

Sources/ZcashLightClientKit/DAO/TransactionDao.swift
daira: It seems desirable to me for the order to be stable; it makes comparison an NHS testing easier, for example.

daira: Oh lol this is a much older PR than I thought. At least I'm consistent: https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/pull/1275#discussion_r1330787815

PR #1373, opened 2024-02-09
[#1153] Allow runtime switch of lightwalletd servers

### LINKED ISSUE #1153: Allow runtime switch of lightwalletd servers
Initialization of the SDK is done when an instance of `Initializer` is provided into the init method of the SDKSynchronizer. 

`Initializer` is basically a way how users of the SDK provides app/custom data to the SDK. URLs where the databases are stored as well as an endpoint. The SDK when properly initilalized uses this endpoint during the whole lifetime. That means any user of the SDK can actually initialize it with a different endpoint. X builds of the wallets can be done and every one can use a different endpoint. Let's call it a non-runtime setup. The endpoint is decided before the SDK is prepared and initialized.

The situation is much different when the SDK is initialized and someone would like to switch the endpoint runtime. There's no API at the moment that could offer such feature. It should be possible but it's a complex task that needs to be properly thought through. I see at least these to happen:

- update of the LightWalletGRPCService, new one must be be initialized and registered as a dependency => overriding the previous one
- broadcast somehow that this dependency changed, many parties rely on it and got the reference to the previous LightWalletGRPCService, all parties must throw it away and receive new one
- watch sync is (not) in process, stops the sync, clear out the states, etc.

To test it with valid alternatives, use (mainnet only!)

| Location | Host | Port |
| --- | --- | --- |
| North America  | [na.lightwalletd.com](http://na.lightwalletd.com/) | 443 |
| South America | [sa.lightwalletd.com](http://sa.lightwalletd.com/) | 443 |
| Europe & Africa | [eu.lightwalletd.com](http://eu.lightwalletd.com/) | 443 |
| Asia & Oceania | [ai.lightwalletd.com](http://ai.lightwalletd.com/) | 443 |

LukasKorba:
Closes #1153 

- new public API switchTo(endpoint:) implemented
- API documented
- code cleaned up

# Author

# Reviewer


Sources/ZcashLightClientKit/Error/ZcashError.swift
```diff
@@ -121,6 +121,11 @@ public enum ZcashError: Equatable, Error {
     /// - `rustError` contains error generated by the rust layer.
     /// ZRUST0003
     case rustDecryptAndStoreTransaction(_ rustError: String)
+    /// Error from rust layer when calling ZcashRustBackend.getBalance
+    /// - `account` is account passed to ZcashRustBackend.getBalance.
+    /// - `rustError` contains error generated by the rust layer.
+    /// ZRUST0004
+    case rustGetBalance(_ account: Int, _ rustError: String)
```
HonzaR, line 128: These changes do not seem to be related. Do we need just rebase?

Example/ZcashLightClientSample/ZcashLightClientSample/DemoAppConfig.swift
```diff
@@ -21,15 +21,10 @@ enum DemoAppConfig {
...
 
-//    static let defaultBirthdayHeight: BlockHeight = ZcashSDK.isMainnet ? 935000 : 1386000
-//    static let defaultSeed = try! Mnemonic.deterministicSeedBytes(from: """
-//    live combine flight accident slow soda mind bright absent bid hen shy decade biology amazing mix enlist ensure biology rhythm snap duty soap armor
-//    """)
-
-    static let defaultBirthdayHeight: BlockHeight = ZcashSDK.isMainnet ? 1935000 : 2170000
+    static let defaultBirthdayHeight: BlockHeight = ZcashSDK.isMainnet ? 935000 : 1386000
     static let defaultSeed = try! Mnemonic.deterministicSeedBytes(from: """
-    wish puppy smile loan doll curve hole maze file ginger hair nose key relax knife witness cannon grab despair throw review deal slush frame
```
str4d, line 31: Nit: we probably just want to comment out the other test mnemonic and its birthday, instead of losing it.

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -532,6 +532,90 @@ public class SDKSynchronizer: Synchronizer {
         return subject.eraseToAnyPublisher()
     }
     
+    // MARK: Server switch
+
+    public func switchTo(endpoint: LightWalletEndpoint) async throws {
+        // Stop synchronization
+        let status = await self.status
+        if status != .stopped && status != .disconnected {
+            await blockProcessor.stop()
```
str4d, line 541: I presume that because this is a method on the synchronizer, `self.status` can only be raced here by another thread calling one of the other public synchronizer methods (because as I understand it, `blockProcessor` is the only "background service" in the synchronizer). Where else in the synchronizer can `blockProcessor` be started, or the status be changed between `.unprepared`, `.stopped | .disconnected`, and the other (running) states?

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -532,6 +532,90 @@ public class SDKSynchronizer: Synchronizer {
...
+
+        let validateSever = ValidateServerAction(
+            container: initializer.container,
+            configProvider: CompactBlockProcessor.ConfigProvider(config: await blockProcessor.config)
+        )
+ 
+        do {
+            _ = try await validateSever.run(with: ActionContextImpl(state: .idle)) { _ in }
+        } catch {
+            throw ZcashError.synchronizerServerSwitch
```
str4d, line 564: Should the new server be validated before we stop `blockProcessor` (so in the event of a user mistyping the new server, the SDK continues synchronizing with the existing one)? Or does the `initializer.container.register` call alter a global that `blockProcessor` depends on for correct operation?

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### [#1363] Account balances in the SynchronizerState
 `shieldedBalance: WalletBalance` has been replaced with `accountBalances: AccountBalance`. `AccountBalance` provides the same values as `shieldedBalance` but adds up a pending changes. Under the hood this calls rust's `getWalletSummary` which improved also the syncing initial values of % and balances.
 
+## Added
+
+### [#1153] Allow runtime switch of lightwalletd servers
+New API implemented that allows clients to change the `mainnet` endpoint. Use `func switchTo(endpoint: LightWalletEndpoint) async throws`. 
+Possible errors:
+`ZcashError.synchronizerServerSwitch`: endpoint fails, check the address, port and format address:port,
+`some ZcashError related to synchronizer.Start()`: the switch calls start() at the end and that is the only throwing function except the validation.
```
str4d, line 20:
```suggestion
- `ZcashError.synchronizerServerSwitch`: endpoint fails, check the address, port and format address:port,
- Some `ZcashError` related to `synchronizer.Start()`: the switch calls `start()` at the end and that is the only throwing function except the validation.
```

Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
```diff
@@ -412,6 +412,52 @@ extension CompactBlockProcessor {
     }
 }
 
+// MARK: - Switch server
+
+extension CompactBlockProcessor {
+    func updateService(_ container: DIContainer) {
```
str4d, line 418: Is it possible to use this method during `CompactBlockProcessor` initialization, so that we are exercising this same pathway for normal startup and server-switching usage? If not, I'd like to understand why (maybe because the actions being modified below are instead being constructed directly with the LWD service on initialization?)

str4d (changes requested):
Reviewed 403ae8af3845ce2eb829f163ae51e1c0d7aea129.

Sources/ZcashLightClientKit/Error/ZcashError.swift
LukasKorba: No, @str4d forgot to commit this error so I needed to put it back manually. This is required and expected.

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
LukasKorba: First of all, the method is async one so until it's finished, no other public API can be called and interrupt body of it. Second of all, the state is purposely an Actor, in an async context, it functions like mutex basically. So it can't be changed during read. If your question was about what if this method starts with some state but state changes during the body until it's completed, the answer is no, this can't happen.

Sources/ZcashLightClientKit/Block/CompactBlockProcessor.swift
LukasKorba: When I think about this, the answer is yes. We could use it but compact block processor initialization is a complex several layers and objects involved operation where all actions are basically initialized with the LWD right in the initializer. So the short answer is we could be right now we would only overridden all instances with the same LWD. This could be a follow up thing, to change a few things and initalizers first before we could rely on this updateService path.

Example/ZcashLightClientSample/ZcashLightClientSample/DemoAppConfig.swift
LukasKorba: Returned it back as a comment

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
str4d: That's unfortunate; I presume that it'a a side-effect of reusing `validateServer` here.

str4d (approved):
utACK c7369057f184d1a815057b360a9c8834766b18c0

PR #1382, opened 2024-02-21
Expose APIs for working with transaction proposals

str4d:
Closes Electric-Coin-Company/zcash-swift-wallet-sdk#1343.

# Author

# Reviewer

### DIFF
--- a/Sources/ZcashLightClientKit/Model/Proposal.swift
+++ b/Sources/ZcashLightClientKit/Model/Proposal.swift
@@ -11,6 +11,16 @@ import Foundation
 public struct Proposal: Equatable {
     let inner: FfiProposal
 
+    /// Returns the number of transactions that this proposal will create.
+    ///
+    /// This is equal to the number of `TransactionSubmitResult`s that will be returned
+    /// from `Synchronizer.createProposedTransactions`.
+    ///
+    /// Proposals always create at least one transaction.
+    public func transactionCount() -> Int {
+        inner.steps.count
+    }
+
     /// Returns the total fee to be paid across all proposed transactions, in zatoshis.
     public func totalFeeRequired() -> Zatoshi {
         inner.steps.reduce(Zatoshi.zero) { acc, step in
--- a/Sources/ZcashLightClientKit/Model/Proposal.swift
+++ b/Sources/ZcashLightClientKit/Model/Proposal.swift
@@ -1,0 +1,18 @@
+//
+//  Proposal.swift
+//
+//
+//  Created by Jack Grigg on 20/02/2024.
+//
+
+import Foundation
+
+/// A data structure that describes a series of transactions to be created.
+public struct Proposal: Equatable {
+    let inner: FfiProposal
+
+    /// Returns the total fee to be paid across all proposed transactions, in zatoshis.
+    public func totalFeeRequired() -> Zatoshi {
+        Zatoshi(Int64(inner.balance.feeRequired))
+    }
+}
--- a/Sources/ZcashLightClientKit/Model/Proposal.swift
+++ b/Sources/ZcashLightClientKit/Model/Proposal.swift
@@ -13,7 +13,9 @@ public struct Proposal: Equatable {
 
     /// Returns the total fee to be paid across all proposed transactions, in zatoshis.
     public func totalFeeRequired() -> Zatoshi {
-        Zatoshi(Int64(inner.balance.feeRequired))
+        inner.steps.reduce(Zatoshi.zero) { acc, step in
+            acc + Zatoshi(Int64(step.balance.feeRequired))
+        }
     }
 }
 
--- a/Sources/ZcashLightClientKit/Model/Proposal.swift
+++ b/Sources/ZcashLightClientKit/Model/Proposal.swift
@@ -16,3 +16,18 @@ public struct Proposal: Equatable {
         Zatoshi(Int64(inner.balance.feeRequired))
     }
 }
+
+public extension Proposal {
+    /// IMPORTANT: This function is for testing purposes only. It produces fake invalid
+    /// data that can be used to check UI elements, but will always produce an error when
+    /// passed to `Synchronizer.createProposedTransactions`. It should never be called in
+    /// production code.
+    static func testOnlyFakeProposal(totalFee: UInt64) -> Self {
+        var ffiProposal = FfiProposal()
+        var balance = FfiTransactionBalance()
+
+        balance.feeRequired = totalFee
+
+        return Self(inner: ffiProposal)
+    }
+}
--- a/Sources/ZcashLightClientKit/Synchronizer.swift
+++ b/Sources/ZcashLightClientKit/Synchronizer.swift
@@ -209,6 +209,7 @@ public protocol Synchronizer: AnyObject {
     ///
     /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
     /// `SynchronizerErrors.notPrepared`.
+    @available(*, deprecated, message: "Upcoming SDK 2.1 will create multiple transactions at once for some recipients.")
     func sendToAddress(
         spendingKey: UnifiedSpendingKey,
         zatoshi: Zatoshi,
@@ -222,6 +223,7 @@ public protocol Synchronizer: AnyObject {
     ///
     /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
     /// `SynchronizerErrors.notPrepared`.
+    @available(*, deprecated, message: "Upcoming SDK 2.1 will create multiple transactions at once for some recipients.")
     func shieldFunds(
         spendingKey: UnifiedSpendingKey,
         memo: Memo,
--- a/Sources/ZcashLightClientKit/Synchronizer.swift
+++ b/Sources/ZcashLightClientKit/Synchronizer.swift
@@ -207,7 +207,7 @@ public protocol Synchronizer: AnyObject {
     /// - Parameter toAddress: the recipient's address.
     /// - Parameter memo: an `Optional<Memo>`with the memo to include as part of the transaction. send `nil` when sending to transparent receivers otherwise the function will throw an error
     ///
-    /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
+    /// If `prepare()` hasn't already been called since creation of the synchronizer instance or since the last wipe then this method throws
     /// `SynchronizerErrors.notPrepared`.
     @available(*, deprecated, message: "Upcoming SDK 2.1 will create multiple transactions at once for some recipients.")
     func sendToAddress(
@@ -220,8 +220,9 @@ public protocol Synchronizer: AnyObject {
     /// Shields transparent funds from the given private key into the best shielded pool of the account associated to the given `UnifiedSpendingKey`.
     /// - Parameter spendingKey: the `UnifiedSpendingKey` that allows to spend transparent funds
     /// - Parameter memo: the optional memo to include as part of the transaction.
+    /// - Parameter shieldingThreshold: the minimum transparent balance required before a transaction will be created.
     ///
-    /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
+    /// If `prepare()` hasn't already been called since creation of the synchronizer instance or since the last wipe then this method throws
     /// `SynchronizerErrors.notPrepared`.
     @available(*, deprecated, message: "Upcoming SDK 2.1 will create multiple transactions at once for some recipients.")
     func shieldFunds(
@@ -275,7 +276,7 @@ public protocol Synchronizer: AnyObject {
 
     /// Returns the latests UTXOs for the given address from the specified height on
     ///
-    /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
+    /// If `prepare()` hasn't already been called since creation of the synchronizer instance or since the last wipe then this method throws
     /// `SynchronizerErrors.notPrepared`.
     func refreshUTXOs(address: TransparentAddress, from height: BlockHeight) async throws -> RefreshedUTXOs
 
@@ -299,7 +300,7 @@ public protocol Synchronizer: AnyObject {
     /// `rewind(policy:)` itself doesn't start the sync process when it's done and it doesn't trigger notifications as regorg would. After it is done
     /// you have start the sync process by calling `start()`
     ///
-    /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then returned publisher emits
+    /// If `prepare()` hasn't already been called since creation of the synchronizer instance or since the last wipe then returned publisher emits
     /// `SynchronizerErrors.notPrepared` error.
     ///
     /// - Parameter policy: the rewind policy
--- a/Sources/ZcashLightClientKit/Synchronizer.swift
+++ b/Sources/ZcashLightClientKit/Synchronizer.swift
@@ -176,14 +176,22 @@ public protocol Synchronizer: AnyObject {
     /// - Parameter accountIndex: the account for which to shield funds.
     /// - Parameter shieldingThreshold: the minimum transparent balance required before a proposal will be created.
     /// - Parameter memo: an optional memo to include as part of the proposal's transactions.
+    /// - Parameter transparentReceiver: a specific transparent receiver within the account
+    ///             that should be the source of transparent funds. Default is `nil` which
+    ///             will select whichever of the account's transparent receivers has funds
+    ///             to shield.
+    ///
+    /// Returns the proposal, or `nil` if the transparent balance that would be shielded
+    /// is zero or below `shieldingThreshold`.
     ///
     /// If `prepare()` hasn't already been called since creation of the synchronizer instance or since the last wipe then this method throws
     /// `SynchronizerErrors.notPrepared`.
     func proposeShielding(
         accountIndex: Int,
         shieldingThreshold: Zatoshi,
-        memo: Memo
-    ) async throws -> Proposal
+        memo: Memo,
+        transparentReceiver: TransparentAddress?
+    ) async throws -> Proposal?
 
     /// Creates the transactions in the given proposal.
     ///
--- a/Sources/ZcashLightClientKit/Synchronizer.swift
+++ b/Sources/ZcashLightClientKit/Synchronizer.swift
@@ -154,7 +154,53 @@ public protocol Synchronizer: AnyObject {
     /// - Parameter accountIndex: the optional accountId whose address is of interest. By default, the first account is used.
     /// - Returns the address or nil if account index is incorrect
     func getTransparentAddress(accountIndex: Int) async throws -> TransparentAddress
-    
+
+    /// Creates a proposal for transferring funds to the given recipient.
+    ///
+    /// - Parameter accountIndex: the account from which to transfer funds.
+    /// - Parameter recipient: the recipient's address.
+    /// - Parameter amount: the amount to send in Zatoshi.
+    /// - Parameter memo: an optional memo to include as part of the proposal's transactions. Use `nil` when sending to transparent receivers otherwise the function will throw an error.
+    ///
+    /// If `prepare()` hasn't already been called since creation of the synchronizer instance or since the last wipe then this method throws
+    /// `SynchronizerErrors.notPrepared`.
+    func proposeTransfer(
+        accountIndex: Int,
+        recipient: Recipient,
+        amount: Zatoshi,
+        memo: Memo?
+    ) async throws -> Proposal
+
+    /// Creates a proposal for shielding any transparent funds received by the given account.
+    ///
+    /// - Parameter accountIndex: the account for which to shield funds.
+    /// - Parameter shieldingThreshold: the minimum transparent balance required before a proposal will be created.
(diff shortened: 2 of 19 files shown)


Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -265,6 +265,60 @@ public class SDKSynchronizer: Synchronizer {
...
+
+        let transactions = try await transactionEncoder.createProposedTransactions(
+            proposal: proposal,
+            spendingKey: spendingKey
+        )
+
+        for transaction in transactions {
+            let encodedTransaction = try transaction.encodedTransaction()
+
+            try await transactionEncoder.submit(transaction: encodedTransaction)
```
str4d, line 316: Currently the entire method throws an error if the submission fails (either due to a gRPC error, or due to getting an error in the response from `lightwalletd`). This code should instead collect the first error that occurs, not attempt to submit any subsequent transactions, and return to the caller the three separate categories of transactions (submitted, failed, not attempted). [I do this with a `Flow` in the Android SDK](https://github.com/Electric-Coin-Company/zcash-android-wallet-sdk/pull/1397/files#diff-4b4a0b33c9b47a30dbd2e951c80e04c51b5be81ca83b2a36546c0c4400785b77R579-R595), but I don't know the equivalent right way to do this in Swift. cc @LukasKorba.

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -154,7 +154,52 @@ public protocol Synchronizer: AnyObject {
...
+    ) async throws -> Proposal
+
+    /// Creates a proposal for shielding any transparent funds received by the given account.
+    ///
+    /// - Parameter accountIndex: the account for which to shield funds.
+    /// - Parameter memo: an optional memo to include as part of the proposal's transactions.
+    ///
+    /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
+    /// `SynchronizerErrors.notPrepared`.
+    func proposeShielding(
```
nuttycom, line 181: I'm concerned that as a high-level API, this will not adequately protect the user's privacy because it will potentially link addresses. At the SDK level, do we provide the ability to generate diversified addresses? If so, there might be a preexisting problem here with the existing methods that are exposed.

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -265,6 +265,78 @@ public class SDKSynchronizer: Synchronizer {
...
+            proposal: proposal,
+            spendingKey: spendingKey
+        )
+        var iterator = transactions.makeIterator()
+        var submitFailed = false
+
+        return AsyncThrowingStream() {
+            guard let transaction = iterator.next() else { return nil }
+
+            if (submitFailed) {
```
nuttycom, line 321:
```suggestion
            if submitFailed {
```

Sources/ZcashLightClientKit/Transaction/TransactionEncoder.swift
```diff
@@ -19,44 +19,49 @@ public enum TransactionEncoderError: Error {
...
-    /// - Parameter MemoBytes: string containing the memo (optional)
-    /// - Parameter accountIndex: index of the account that will be used to send the funds
+    /// - Parameter accountIndex: the account from which to transfer funds.
+    /// - Parameter recipient: string containing the recipient's address.
+    /// - Parameter amount: the amount to send in Zatoshi.
+    /// - Parameter memo: an optional  memo to include as part of the proposal's transactions. Use `nil` when sending to transparent receivers otherwise the function will throw an error.
+    ///
+    /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
+    /// `SynchronizerErrors.notPrepared`.
+    func proposeTransfer(
```
nuttycom, line 31: Now that we have this functionality and also https://github.com/Electric-Coin-Company/zcash-light-client-ffi/pull/115, we should be able to expose the latter in the API?

Sources/ZcashLightClientKit/Transaction/WalletTransactionEncoder.swift
```diff
@@ -54,93 +54,52 @@ class WalletTransactionEncoder: TransactionEncoder {
...
-
         let txId = try await rustBackend.createProposedTransaction(
-            proposal: proposal,
+            proposal: proposal.inner,
             usk: spendingKey
         )
 
-        return txId
+        logger.debug("transaction id: \(txId)")
+        return [try await repository.find(rawID: txId)]
```
nuttycom, line 102: Does `repository.find` throw if it cannot find the txid? I would think that ideally that would return an optional result, so that "not found" is distinguished from "error" (though in this particular case you'd want to throw an error anyway, so it's nonblocking for now.)

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
nuttycom, line 1: While you're touching this file, please fix the unterminated `swiftlint:disable`.

nuttycom (approved):
utACK with questions that we should probably create follow-up issues to address.

Sources/ZcashLightClientKit/Transaction/WalletTransactionEncoder.swift
str4d: We shouldn't encounter this case (successful wallet DB lookup that returns no result) because if we reach here, the transaction was successfully created by the Rust backend (and didn't itself result in an exception). Any exception thrown here would instead be due to database access failures, which are indeed exceptions.

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -154,7 +154,52 @@ public protocol Synchronizer: AnyObject {
...
-    
+
+    /// Creates a proposal for transferring funds to the given recipient.
+    ///
+    /// - Parameter accountIndex: the account from which to transfer funds.
+    /// - Parameter recipient: the recipient's address.
+    /// - Parameter amount: the amount to send in Zatoshi.
+    /// - Parameter memo: an optional  memo to include as part of the proposal's transactions. Use `nil` when sending to transparent receivers otherwise the function will throw an error.
+    ///
+    /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
```
daira, line 165:
```suggestion
    /// If `prepare()` hasn't already been called since creation of the synchronizer instance or since the last wipe then this method throws
```

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -154,7 +154,52 @@ public protocol Synchronizer: AnyObject {
...
     /// - Returns the address or nil if account index is incorrect
     func getTransparentAddress(accountIndex: Int) async throws -> TransparentAddress
-    
+
+    /// Creates a proposal for transferring funds to the given recipient.
+    ///
+    /// - Parameter accountIndex: the account from which to transfer funds.
+    /// - Parameter recipient: the recipient's address.
+    /// - Parameter amount: the amount to send in Zatoshi.
+    /// - Parameter memo: an optional  memo to include as part of the proposal's transactions. Use `nil` when sending to transparent receivers otherwise the function will throw an error.
```
daira, line 163:
```suggestion
    /// - Parameter memo: an optional memo to include as part of the proposal's transactions. Use `nil` when sending to transparent receivers otherwise the function will throw an error.
```

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -154,7 +154,52 @@ public protocol Synchronizer: AnyObject {
...
+        accountIndex: Int,
+        recipient: Recipient,
+        amount: Zatoshi,
+        memo: Memo?
+    ) async throws -> Proposal
+
+    /// Creates a proposal for shielding any transparent funds received by the given account.
+    ///
+    /// - Parameter accountIndex: the account for which to shield funds.
+    /// - Parameter memo: an optional memo to include as part of the proposal's transactions.
```
daira, line 177: `shieldingThreshold` needs to be documented both here and on the deprecated version.

Sources/ZcashLightClientKit/Synchronizer.swift
daira: This is not a regression so I'm fine with it for now. The Right Thing here is to do something similar to what we did for the Sprout->Sapling transition, or make the timing of shielding transactions follow a Poisson distribution.

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -154,7 +154,52 @@ public protocol Synchronizer: AnyObject {
...
+    ///
+    /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
+    /// `SynchronizerErrors.notPrepared`.
+    func proposeShielding(
+        accountIndex: Int,
+        shieldingThreshold: Zatoshi,
+        memo: Memo
+    ) async throws -> Proposal
+
+    /// Creates the transactions in the given proposal
```
daira, line 187:
```suggestion
    /// Creates the transactions in the given proposal.
```

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -154,7 +154,52 @@ public protocol Synchronizer: AnyObject {
...
+    ///
+    /// - Parameter proposal: the proposal to create.
```
daira, line 189:
```suggestion
    /// - Parameter proposal: the proposal for which to create transactions.
```

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -154,7 +154,52 @@ public protocol Synchronizer: AnyObject {
...
+    /// - Parameter spendingKey: the `UnifiedSpendingKey` associated with the notes that will be spent.
```
daira, line 190:
```suggestion
    /// - Parameter spendingKey: the `UnifiedSpendingKey` associated with the account for which the proposal was created.
```

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -176,6 +222,7 @@ public protocol Synchronizer: AnyObject {
     ///
```
daira, line 222: Document `shieldingThreshold`.

Sources/ZcashLightClientKit/Transaction/TransactionEncoder.swift
```diff
@@ -19,44 +19,49 @@ public enum TransactionEncoderError: Error {
...
-    /// - Parameters:
-    /// - Parameter spendingKey: a `UnifiedSpendingKey` containing the spending key
-    /// - Parameter zatoshi: the amount to send in `Zatoshi`
-    /// - Parameter to: string containing the recipient address
-    /// - Parameter MemoBytes: string containing the memo (optional)
-    /// - Parameter accountIndex: index of the account that will be used to send the funds
+    /// - Parameter accountIndex: the account from which to transfer funds.
+    /// - Parameter recipient: string containing the recipient's address.
+    /// - Parameter amount: the amount to send in Zatoshi.
+    /// - Parameter memo: an optional  memo to include as part of the proposal's transactions. Use `nil` when sending to transparent receivers otherwise the function will throw an error.
```
daira, line 27: Revert back to `memoBytes`.

Sources/ZcashLightClientKit/Transaction/TransactionEncoder.swift
```diff
@@ -19,44 +19,49 @@ public enum TransactionEncoderError: Error {
...
+        accountIndex: Int,
+        recipient: String,
+        amount: Zatoshi,
+        memoBytes: MemoBytes?
+    ) async throws -> Proposal
+
+    /// Creates a proposal for shielding any transparent funds received by the given account.
+    ///
+    /// - Parameter accountIndex: the account for which to shield funds.
+    /// - Parameter memo: an optional memo to include as part of the proposal's transactions.
```
daira, line 41: Document `shieldingThreshold`.

Sources/ZcashLightClientKit/Transaction/TransactionEncoder.swift
```diff
@@ -19,44 +19,49 @@ public enum TransactionEncoderError: Error {
...
+    ///
+    /// If `prepare()` hasn't already been called since creating of synchronizer instance or since the last wipe then this method throws
```
daira, line 43:
```suggestion
    /// If `prepare()` hasn't already been called since creation of the synchronizer instance or since the last wipe then this method throws
```

Sources/ZcashLightClientKit/Transaction/TransactionEncoder.swift
```diff
@@ -19,44 +19,49 @@ public enum TransactionEncoderError: Error {
...
+    /// `SynchronizerErrors.notPrepared`.
+    func proposeShielding(
+        accountIndex: Int,
+        shieldingThreshold: Zatoshi,
+        memoBytes: MemoBytes?
+    ) async throws -> Proposal
+
+    /// Creates the transactions in the given proposal
```
daira, line 51:
```suggestion
    /// Creates the transactions in the given proposal.
```

Sources/ZcashLightClientKit/Transaction/TransactionEncoder.swift
```diff
@@ -19,44 +19,49 @@ public enum TransactionEncoderError: Error {
...
+    ///
+    /// - Parameter proposal: the proposal to create.
```
daira, line 53: Same change as above.

Sources/ZcashLightClientKit/Transaction/TransactionEncoder.swift
str4d: Yes; @pacu is working on that in #1380.

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
str4d: I don't see any.

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/compare/4d4d7492cb7916ded6ab2a1b7a0894847a470690..cf6d4a115b85d801f886714d7927833db3919c40) to address review comments.

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
str4d: Oh, you mean the "Blanket Disable Command Violation"? IDK what the intention is there, that's a question for @LukasKorba (and should probably be addressed in a separate PR).

Sources/ZcashLightClientKit/Synchronizer.swift
nuttycom: I'm not talking about timing here, although that is also a concern; with this API as written, it provides no mechanism for wallets that have given out UAs to different parties to select which transparent receivers the funds should be spent from in shielding. So any concurrent transparent receives to distinct Unified Addresses will be linked. The Rust API takes a vector of transparent addresses to shield from, so that wallets can choose to link addresses if they want but aren't compelled to: https://github.com/zcash/librustzcash/blob/main/zcash_client_backend/src/data_api/wallet.rs#L1195

str4d: In a pairing today, @nuttycom and I decided that we will change `Synchronizer.proposeShielding` in two ways:
- It will take an optional transparent receiver, corresponding to a specific diversified UA (or the default transparent address) that should have its transparent funds shielded.
  - If the transparent receiver is not part of the specified account, an error will be returned.
  - If no receiver is provided, the current behaviour of "shield transparent funds in whichever transparent receiver under the account has them" will be used.
- It will return an error if there are transparent funds in more than one transparent receiver in the given account.
  - This fixes an address linkability problem that currently exists with `Synchronizer.shieldFunds`, wherein a transaction spending funds from multiple transparent receivers in the same account can be created which indelibly links them on-chain.
  - We are not fixing the problem in `Synchronizer.shieldFunds` because working around it would require changing the `shieldFunds` API, but we are deprecating and soon removing the method instead.

The transparent receiver argument serves two functions:
- It is a (more complicated but safer) escape hatch to use when the new "multiple transparent receivers" error is encountered.
- It enables SDK users to implement their own mecahnism for avoiding address linkability when shielding funds, in the interim while we do not have our own solution (https://github.com/zcash/librustzcash/issues/1201).

str4d: The API changes for this have now been implemented, but the functional changes require FFI changes so will need to go in a separate feature branch.

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
LukasKorba: I'll resolve it as a follow up for the whole SDK, there are more instances unrelated to this issue.

Sources/ZcashLightClientKit/Model/Proposal.swift
```diff
@@ -8,11 +8,26 @@
...
     let inner: FfiProposal
 
     /// Returns the total fee to be paid across all proposed transactions, in zatoshis.
     public func totalFeeRequired() -> Zatoshi {
-        return Zatoshi(Int64(inner.balance.feeRequired))
+        Zatoshi(Int64(inner.balance.feeRequired))
+    }
+}
+
+public extension Proposal {
```
pacu, line 20: I would advise not to expose this publicly.

pacu (changes requested):

Sources/ZcashLightClientKit/Model/Proposal.swift
str4d: If you have a suggestion for how to avoid exposing this publicly, I'm all ears. The constraint here is Zashi iOS using TCA, which needs to be able to construct mock data to be returned from its mock implementation of the `Synchronizer` protocol for use in e.g. preview of UIs (so we can't just return something that would be an error case).

pacu: NOTE: I don't know which things you've tried and which ones you haven't, but there are a few things I can thing from the top of my head. I promise I'll look this deeper shortly. I might be still lacking context and this could be somewhat inaccurate or insufficient. 


### A. "Uncle bob's Clean code" approach
Treat the whole SDK as an external dependency under the principle of Bob Martin's clean code architecture. Look at the problem as you were Unstoppable or EDGE and you don't control the dependency. Wrap around it and have a shim that you Do control and use that instead. 

### B. Pointfree.co / TCA approach. 
Make proposal on the SDK Protocol. The SDK returns an implementation of it. Then use a protocol witness or a mock implementation on Zashi to provide the mocked version that you need. 

### A and B
Make a protocol on Zashi  for what the app expects of a proposal. This propocol contains a throwing `toProposal()` function. 
Make an Adapter of the SDK's Proposal type that conforms to Zashi's. 
Make a Mock implementation for testing on Zashi that throws whenever `toProposal()` is invoked.

Sources/ZcashLightClientKit/Transaction/TransactionEncoder.swift
pacu: Rebased!

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/compare/67d564fcdc553daaa440f652f5a12f6eda44a92b..f617f17e2ab8109dddbb53cfd2050d9c8a012e85) to clean up the commits.

LukasKorba:
I tested all 5 APIs while working on adoption of this PR in Zashi. I can confirm
- deprecated Send API still works ok
- deprecated shileding API still works ok
- new proposeTransfer API works ok
- new proposeShielding API works ok
- createProposedTransactions API works ok

str4d:
Before we merge this, I'm going to add a helper method to `Proposal` to expose the number of steps, so that callers can anticipate how many events they will get from the async stream returned from `createProposedTransactions`.

nuttycom (approved):
cursory utACK, modulo @pacu's comments

nuttycom (approved):
re-utACK 72acbc6700df62e76bdb1deb016610b445c828a7

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -265,6 +265,84 @@ public class SDKSynchronizer: Synchronizer {
...
+        )
+
+        let transactions = try await transactionEncoder.createProposedTransactions(
+            proposal: proposal,
+            spendingKey: spendingKey
+        )
+        var iterator = transactions.makeIterator()
+        var submitFailed = false
+
+        return AsyncThrowingStream() {
```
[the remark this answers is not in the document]
LukasKorba: It's async stream and there's very specific way how to work with it. The stream emits values until it finishes (nil value). And a client awaits it until the nil is emitted (or an error in case of AsyncThrowingStream - throwing means this stream can emit an error, another stream is AsyncStream which ends only when nil is emitted).

str4d:
[Force-pushed](https://github.com/Electric-Coin-Company/zcash-swift-wallet-sdk/compare/72acbc6700df62e76bdb1deb016610b445c828a7..129ac4398e86a62abd73056d55c26ba947dd5ef6) to remove the 2.0.11 changelog bump, as @LukasKorba will make a few other release updates before the actual release commit.

PR #1569, opened 2025-06-12
[#1568] Postponed initialization of Tor

### LINKED ISSUE #1568: Tor blocking main thread
- initialization of TorClient done with DI is blocking main thread
- refresh USD rate is blocking main thread

Fix is about postponing `zcashlc_create_tor_runtime` call. `TorClient` instance is created when DI is established but rust's side is done later when TorClient is actually needed.

Fix for the rate is about putting it on the thread, it can do its work and when done, it simply notifies the subject - reactive approach.

LukasKorba:
Closes #1568 

- Tor is initialized outside dependency injection instance creation
- Refresh USD rate is encapsulated with Task

# Author

# Reviewer

### DIFF
--- a/Sources/ZcashLightClientKit/Tor/TorClient.swift
+++ b/Sources/ZcashLightClientKit/Tor/TorClient.swift
@@ -9,10 +9,13 @@ import Foundation
 import libzcashlc
 
 public class TorClient {
-    private let runtime: OpaquePointer
+    private var underlyingRuntime: OpaquePointer?
+    private var torDir: URL?
+
     public var cachedFiatCurrencyResult: FiatCurrencyResult?
 
     init(torDir: URL) throws {
+        self.torDir = torDir
         // Ensure that the directory exists.
         let fileManager = FileManager()
         if !fileManager.fileExists(atPath: torDir.path) {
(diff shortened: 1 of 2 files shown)


Sources/ZcashLightClientKit/Tor/TorClient.swift
```diff
@@ -22,26 +25,42 @@ public class TorClient {
...
         zcashlc_free_tor_runtime(runtime)
     }
+    
+    private func resolveRuntime() throws -> OpaquePointer {
+        if let runtime = underlyingRuntime {
+            return runtime
+        }
+
+        guard let torDir else {
+            throw ZcashError.rustTorClientInit(lastErrorMessage(fallback: "`TorClient` init failed with torDir nil"))
```
str4d, line 46: How can this error occur? `torDir` is provided during `init`, so surely it can be stored as a `URL` instead of a `URL?` and guaranteed to exist?

Sources/ZcashLightClientKit/Tor/TorClient.swift
```diff
@@ -22,26 +25,42 @@ public class TorClient {
...
+        }
+
+        let rawDir = torDir.osPathStr()
+        let runtimePtr = zcashlc_create_tor_runtime(rawDir.0, rawDir.1)
+        
+        guard let runtimePtr else {
+            throw ZcashError.rustTorClientInit(lastErrorMessage(fallback: "`TorClient` init failed with unknown error"))
+        }
+        
+        underlyingRuntime = runtimePtr
```
str4d, line 56: Bug: now that initialization of `underlyingRuntime` is lazy, it needs to be protected with a lock. Otherwise two `TorClient` calls could race, both observe that `underlyingRuntime` is unset, create two Tor clients running from the same Tor dir (and interfering with each other), and then only save the pointer to one of them (causing the other to never be freed).

Sources/ZcashLightClientKit/Tor/TorClient.swift
```diff
@@ -22,26 +25,42 @@ public class TorClient {
...
-        runtime = runtimePtr
     }
 
     private init(runtimePtr: OpaquePointer) {
-        runtime = runtimePtr
+        underlyingRuntime = runtimePtr
     }
 
     deinit {
+        guard let runtime = underlyingRuntime else { return }
```
str4d, line 35: I presume that this access does not need to be guarded by the lock, because I presume that `deinit` cannot run while any other `TorClient` method call is running?

Sources/ZcashLightClientKit/Tor/TorClient.swift
```diff
@@ -22,26 +25,42 @@ public class TorClient {
...
+        
         zcashlc_free_tor_runtime(runtime)
     }
+    
+    private func resolveRuntime() throws -> OpaquePointer {
```
str4d, line 40: Postponing `zcashlc_create_tor_runtime` will lead to worse first-use response time. Starting up a Tor node takes several seconds, potentially tens of seconds if it is the first time the wallet has run. By deferring its startup until it is first used, that delay might be user-observable (depending on which call causes it).

Have you checked what pathway tends to start `TorClient` now?

str4d (changes requested):
Reviewed 18d27030fedf1429f3b1887ac3558fd7ba56d32e

Sources/ZcashLightClientKit/Tor/TorClient.swift
LukasKorba: It's a tradeoff because we also have 
```swift
private init(runtimePtr: OpaquePointer) {
        underlyingRuntime = runtimePtr
    }
```
and that doesn't have torDir but requires to resolve it. A or B solution have downsides but I changed it for now to be torDir: URL and I extended this private initializer to input it.

PR #1636, opened 2026-02-17
Integrate the FFI sources with the SDK, with Github releases for build artifacts.

nuttycom:
The zcash-light-client-ffi repository being separate from the SDK has been a longstanding source of friction in SDK development, and the practice of committing release binaries to the FFI repository has bloated the repository to the point that it is no longer usable. This update reintroduces the FFI code to the SDK, including all of the git history that existed in the zcash-light-client-ffi repository. It then adds a release process that relies on github release artifacts for library dependencies instead of committing build binaries to git.

# Author

# Reviewer


rust/build.rs
```diff
@@ -0,0 +1,67 @@
+extern crate cbindgen;
+
+use std::{env, path::PathBuf};
+
+fn main() {
+    println!("cargo:rerun-if-changed=rust/src/lib.rs");
+    println!("cargo:rerun-if-changed=rust/wrapper.c");
+    println!("cargo:rerun-if-changed=rust/wrapper.h");
```
str4d: I verified that the only changes to this file (relative to the aforementioned revision) are to adjust the paths to reflect the `Cargo.toml` being in the root.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,176 @@
...
+# Initialize local FFI (builds from source)
+./Scripts/init-local-ffi.sh
+
+# Or use --cached to download pre-built release (faster, but may not match local source)
+./Scripts/init-local-ffi.sh --cached
+```
+
+Then add the local package to Xcode:
+1. Open your Xcode project/workspace
+2. File → Add Package Dependencies → Add Local...
```
str4d: Bug: this does not exist in Xcode 26.1.1.

There is `File → Add Package Dependencies...` which opens a window, but I cannot find anywhere in that window to add a local package.

Scripts/init-local-ffi.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+# Initialize local FFI development environment
+# Usage: ./Scripts/init-local-ffi.sh [--cached]
+#   --cached   Download pre-built release instead of building from source
+
+set -e
+cd "$(dirname "$0")/.."
+
+USE_CACHED=false
```
str4d: I verified that the non-cached pathway builds libzcashlc.xcframework locally. I did not test `--cached`.

I noted that it didn't work for me in the default macOS Terminal, because for some reason zsh has a screwed-up PATH for me and forcibly places Homebrew's Rust installation ahead of my rustup installation (causing the targets to both not be visible if already installed, and not be installable, because all of that relies on rustup). Using Ghostty somehow fixes my PATH issue even though it is also using zsh internally.

.github/workflows/build-ffi.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Create Draft Release
+        uses: softprops/action-gh-release@v1
+        with:
+          tag_name: ${{ github.event.inputs.version }}
+          name: ${{ github.event.inputs.version }}
+          draft: true
+          files: |
+            BuildSupport/products/libzcashlc.xcframework.zip
+            BuildSupport/products/checksum.txt
```
str4d: Was the `checksum.txt` file added intentionally, or did the AI generate it on a whim? If the download URL for the XCFramework could be intercepted somewhere to swap in a malicious archive, then that same intercept can swap the checksum file to one matching the malicious archive. Checksums are only useful when served via a divergent pathway; this serves it from the exact same CDN as the file it is checking.

.github/workflows/build-ffi.yml
```diff
@@ -0,0 +1,77 @@
...
+        required: true
+
+env:
+  RUST_TOOLCHAIN: stable
+
+jobs:
+  build-xcframework:
+    runs-on: macos-15
+    steps:
+      - uses: actions/checkout@v4
```
str4d: What revision is this going to check out? It seems weird that there is no apparent way to control which commit / branch / tag is used to build the release.

Scripts/release.sh
```diff
@@ -0,0 +1,171 @@
...
+#   <version>  The version to release (e.g., '2.5.0')
+#
+# Prerequisites:
+#   - gh CLI installed and authenticated
+#   - Rust toolchain with all Apple targets
+#   - GPG key configured for signing tags
+#   - Clean working directory (no uncommitted changes)
+#
+# For security releases where you need more control over timing,
+# use prepare-release.sh instead and perform steps manually.
```
str4d: Maintenance hazard: the AI has generated three different release scripts that all contain basically the same logic:
- `build-ffi.yml` workflow, which is only triggerable via `workflow_dispatch` and thus is intended to be triggered manually.
- `prepare-release.sh`, which does exactly the same thing as the workflow, but run from the CLI instead of GitHub (so it runs on your local hardware instead of a GH runner, and can be run against a completely arbitrary and unverifiable local repo state).
- `release.sh`, which includes a chunk of logic that is identical to `prepare-release.sh`.

We should instead have a single way to prepare a draft release (building the XCFramework binaries, creating the draft release, and uploading the framework binaries), and then two ways to use it (either directly for security releases, or as part of a full release script). I don't particularly mind whether a Bash script or a GitHub workflow is the canonical way (it should be possible to trigger the workflow from the CLI via `gh api`, which the scripts assume is installed; meanwhile a workflow could itself just call the Bash script).

Scripts/rebuild-local-ffi.sh
```diff
@@ -0,0 +1,109 @@
...
+# Regenerate headers if build.rs produced new ones
+if [[ -d "target/Headers" ]]; then
+    cp -R "target/Headers" "$FRAMEWORK_DIR/"
+fi
+
+echo ""
+echo "Rebuilt $TARGET in $XCFRAMEWORK_DIR"
+echo ""
+echo "Note: This is a single-architecture build for local development only."
+echo "      The xcframework now contains only $RUST_TARGET for the $XCFRAMEWORK_SLICE slice."
```
str4d: Bug: the AI claims that its rebuild script results in a single-architecture build, but all this does is copy the single-build binary over the top of whatever happens to be there. In local testing, I confirmed that the other binaries produced by `Scripts/init-local-ffi.sh` are all still present, meaning that the xcframework is now in a broken state for any target other than whatever this rebuild script was run with.

In particular, if someone runs this script with no arguments (rebuilding just for the simulator) and then plugs in an iPhone and deploys to it, Xcode will AFAICT happily use the un-updated binaries. In the best case that results in a compilation failure due to incompatible header changes; in the worst case the deployed binary silently corrupts data across the FFI, and the deployed binary is used with real funds resulting in user loss-of-funds.

AFAICT if someone is using `ZcashSDK.xcworkspace` (i.e. if `FFIBuilder` is part of the build tree) then it *should* trigger a rebuild when the user changes the deployment target, so I think they would not be affected. But IDK if Xcode always triggers that rebuild, or if it would see the binaries on-disk and assume it can use them. So I'm not 100% confident it wouldn't be an issue here as well.

Instead, the existing XCFramework should be entirely replaced. This should be done atomically, in a way that ensures Xcode doesn't observe and react to a half-broken local framework.

BuildSupport/FFIBuilder.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,236 @@
...
+			inputPaths = (
+			);
+			name = "Build FFI";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/bash;
+			shellScript = "# Build FFI for the current platform\n# This script is called by Xcode when building the FFIBuilder target\n\nset -e\ncd \"$SRCROOT/..\"\n\n# Determine target based on Xcode's destination\ncase \"$PLATFORM_NAME\" in\n    macosx)\n        TARGET=\"macos\"\n        ;;\n    iphonesimulator)\n        TARGET=\"ios-sim\"\n        ;;\n    iphoneos)\n        TARGET=\"ios-device\"\n        ;;\n    *)\n        echo \"warning: Unknown platform $PLATFORM_NAME, defaulting to ios-sim\"\n        TARGET=\"ios-sim\"\n        ;;\nesac\n\necho \"Building FFI for $TARGET...\"\n./Scripts/rebuild-local-ffi.sh \"$TARGET\"\n";
```
str4d: This is the only part of this file I reviewed. It looks correct, but I could not test it locally because of package resolution failing due to the missing remote binary.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,176 @@
...
+
+```bash
+open ZcashSDK.xcworkspace
+```
+
+This workspace includes:
+- The SDK package (for editing Swift code)
+- `FFIBuilder` target with a Run Script build phase
+
+**How it works:** When you build the FFIBuilder target, it runs `rebuild-local-ffi.sh` with the appropriate platform based on your selected destination. You can add FFIBuilder as a dependency to ensure FFI rebuilds happen before SDK builds.
```
str4d: > You can add FFIBuilder as a dependency

So the existing workspace does not have this automatically? Why can't it?

How does the user do this themselves? The documentation needs to say.

docs/LOCAL_DEVELOPMENT.md
Chlup: I don't think `File → Add Package Dependencies → Add Local` can be used here. This thing in Xcode is used when adding/managing depedencies on project level. Because when this UI is used Xcode stores dependencies info into project file. It doesn't modify Package.swift file (because ordinary projects don't have it). For example `ZcashLightClientSample` project from this repo has depedencies manages exactly in this way. But in this case Package.swift file from SDK itself must be modified. 

After I did steps described in this MD file and compiled ffi I modified SDK Package.swift like this:
```
dependencies: [
        .package(url: "https://github.com/grpc/grpc-swift.git", from: "1.24.2"),
        .package(url: "https://github.com/stephencelis/SQLite.swift.git", from: "0.15.3"),
        .package(path: "/Users/chlup/Developing/zcash/zcash/ZcashLightClientKit/LocalPackages")
    ],
    targets: [
        // Binary target for the Rust FFI library
        // Updated by Scripts/release.sh during the release process
//        .binaryTarget(
//            name: "libzcashlc",
//            url: "https://github.com/zcash/zcash-swift-wallet-sdk/releases/download/0.0.1/libzcashlc.xcframework.zip",
//            checksum: "0000000000000000000000000000000000000000000000000000000000000000"
//        ),
        .target(
            name: "ZcashLightClientKit",
            dependencies: [
                .product(name: "SQLite", package: "SQLite.swift"),
                .product(name: "GRPC", package: "grpc-swift"),
                .product(name: "libzcashlc", package: "LocalPackages")
//                "libzcashlc"
            ],
```

- Local ffi package is added: `.package(path: "/Users/chlup/Developing/zcash/zcash/ZcashLightClientKit/LocalPackages")`
- Use of framework from release must be commented otherwise resolution fails in Xcode.
- Updated how libzcashlc is added to target depedencies. 

After these modifications I can compile SDK with local ffi package. 

Documentation should say that developer needs to have either own fork or SDK or developer must use it as local package in his/her app. In both cases developer can modify SDK Package.swift file. And then these required changes should be described.

Scripts/prepare-release.sh
```diff
@@ -0,0 +1,131 @@
...
+# Prepare FFI artifacts for an SDK release
+# Usage: ./Scripts/prepare-release.sh [--force-overwrite-existing-release] <version>
+#
+# This script:
+#   1. Builds the full xcframework (all architectures)
+#   2. Creates a zip archive with checksum
+#   3. Uploads to GitHub as a DRAFT release
+#   4. Outputs the values needed for Package.swift
+#
+# After running this script:
```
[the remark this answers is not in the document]
nuttycom: This is what `release.sh` does.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,219 @@
+# Local FFI Development
+
+This guide explains how to work on the Rust FFI code alongside the Swift SDK.
+
+## Overview
+
+The SDK uses a pre-built XCFramework (`libzcashlc`) for the Rust FFI layer. For most SDK development, you don't need to rebuild the FFI — SPM automatically downloads the pre-built binary from GitHub Releases.
```
str4d: I see the latest AI model changed all these hyphens to em dashes.

docs/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Local Package Override
+
+The `LocalPackages` directory contains a Swift package named `libzcashlc` with the same product name as the binary target in `Package.swift`. When `Package.swift` detects that `LocalPackages/Package.swift` exists, it adds `LocalPackages` as a path dependency and uses it instead of the `.binaryTarget` declaration. This switching is automatic — no manual edits to `Package.swift` are needed.
+
+## Automatic FFI Rebuilds
+
+The shared `ZcashLightClientKit` scheme in `ZcashSDK.xcworkspace` includes `FFIBuilder` as a build dependency. FFIBuilder runs `rebuild-local-ffi.sh` with the appropriate platform based on your selected destination, so Rust code is automatically recompiled when you build in Xcode.
+
+**Note:** The FFIBuilder target requires `init-local-ffi.sh` to have been run first — it calls `rebuild-local-ffi.sh`, which expects `LocalPackages/` to exist.
```
str4d: When I first opened `ZcashSDK.xcworkspace` in Xcode after running the init script, it resolved the package graph fine, and then `FFIBuilder` immediately failed with the error "Command PhaseScriptExecution failed with a nonzero exit code". When I then manually ran "Product -> Build For --> Testing", it completed the build without error.

ValarDragon:
Dumb question, if big binaries made the repo hard to use, couldn't we use git-LFS ?

Having committed binaries is really nice for incremental updates

PR #1639, opened 2026-02-17
feat: EIP-681 transaction request parsing COR-1007

schell:
These changes add EIP-681 parsing, validation and generation to the Swift SDK.

[COR-1007](https://linear.app/zodl/issue/COR-1007/integrate-eip-681-api-into-ios-sdk)

# Author

- [ ] Self-review: Did you review your own code in GitHub's web interface? Code often looks different when reviewing the diff in a browser, making it easier to spot potential bugs.
- [ ] Automated tests: Did you add appropriate automated tests for any code changes?
- [ ] Code coverage: Did you check the code coverage report for the automated tests?  While we are not looking for perfect coverage, the tool can point out potential cases that have been missed.
- [ ] Documentation: Did you update Docs as appropiate? (E.g [README.md](../blob/main/README.md), etc.)
- [ ] Run the app: Did you run the app and try the changes? 
- [ ] Did you provide Screenshots of what the App looks like before and after your changes as part of the description of this PR? (only applicable to UI Changes)
- [ ] Rebase and squash: Did you pull in the latest changes from the main branch and squash your commits before assigning a reviewer? Having your code up to date and squashed will make it easier for others to review. Use best judgement when squashing commits, as some changes (such as refactoring) might be easier to review as a separate commit.

# Reviewer

- [ ] Checklist review: Did you go through the code with the [Code Review Guidelines](../blob/main/CODE_REVIEW_GUIDELINES.md) checklist?
- [ ] Ad hoc review: Did you perform an ad hoc review?  _In addition to a first pass using the code review guidelines, do a second pass using your best judgement and experience which may identify additional questions or comments. Research shows that code review is most effective when done in multiple passes, where reviewers look for different things through each pass._
- [ ] Automated tests: Did you review the automated tests?
- [ ] Manual tests: Did you review the manual tests?_You will find manual testing guidelines under our [manual testing section](../blob/mater/docs/testing/manual_testing)_
- [ ] How is Code Coverage affected by this PR? _We encourage you to compare coverage befor and after your changes and when possible, leave it in a better place. [Learn More...](../blob/master/docs/testing/local_coverage.md)_
- [ ] Documentation: Did you review Docs, [README.md](../blob/master/README.md), [LICENSE.md](../blob/master/LICENSE.md), and [Architecture.md](../blob/master/docs/Architecture.md) as appropriate?
[...]

PR #1684, opened 2026-04-15
ci: Replace CodeQL default setup with advanced setup for Swift FFI builds

schell:
## Summary

- Replace CodeQL default setup with a custom advanced setup workflow that builds the Rust FFI before Swift analysis
- Uses `build-mode: manual` for Swift, replicating the Rust FFI build and `LocalPackages` setup from `swift.yml`
- Uses `build-mode: none` for `actions`, `c-cpp`, and `rust` (matching current default setup behavior)

## Motivation

CodeQL's default setup uses `autobuild` for Swift, which runs `swift build` without first compiling the Rust FFI layer. When `swift build` runs, `Package.swift` detects no `LocalPackages/` directory and falls back to the remote binary target — a pre-built xcframework from the latest GitHub Release. If any new `#[no_mangle] extern "C"` functions have been added to the Rust code since that release, the downloaded xcframework's header won't declare them, and `swift build` fails with "cannot find in scope" errors.

This is currently blocking Swift CodeQL analysis on [#1639](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1639) (EIP-681 support), which adds new FFI functions that don't exist in the 2.4.6 release binary. The regular build check (`swift.yml`) passes because it builds the Rust FFI from source.

Swift does not support `build-mode: none` — only `autobuild` or `manual` — so the only way to inject the Rust build step is via a custom workflow with `build-mode: manual`.

## Admin action required after merge

After this PR lands on `main`, a repository admin needs to **disable CodeQL default setup** in:

> Settings > Code Security > Code scanning

Otherwise both the default setup and this custom workflow will run simultaneously, producing duplicate results. The custom workflow will handle all the same languages (`actions`, `c-cpp`, `rust`, `swift`).

## After disabling default setup

Once default setup is disabled and this workflow is active, [#1639](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1639) should be rebased onto `main` so the new CodeQL workflow runs for it. The Swift analysis will then succeed because the workflow builds the FFI from source before running `swift build`.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -134,9 +134,13 @@ jobs:
         cp -R BuildSupport/products/libzcashlc.xcframework LocalPackages/
         cp BuildSupport/LocalPackages-Package.swift LocalPackages/Package.swift
 
+    - name: Resolve Swift dependencies
+      if: matrix.language == 'swift'
+      run: swift package resolve
+
     - name: Build Swift package
       if: matrix.language == 'swift'
-      timeout-minutes: 15
+      timeout-minutes: 30
       run: swift build
 
     # --- End Swift manual build steps ---
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -16,18 +16,14 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
-  schedule:
-    - cron: '44 7 * * 1'
+
+permissions:
+  contents: read
 
 jobs:
   analyze:
     name: Analyze (${{ matrix.language }})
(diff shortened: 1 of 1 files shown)


schell:
@nuttycom - to see this PR pass we need to disable the default CodeQL setup first, since this provides our own. The default is essentially blocking the SARIF upload from succeeding.

nuttycom (dismissed):
Looks good to me!

nuttycom (dismissed):
utACK 9ab0838b8fbd3843575ce79b59b17793cd0738b0

nuttycom (approved):
utACK 2bb9cddbe196529d05963f088079b0c42706eb65

PR #1689, opened 2026-04-27
Expose fullyScannedHeight on SynchronizerState

greg0x:
## Summary

Adds `fullyScannedHeight` to `SynchronizerState` so callers can gate on contiguous-from-birthday scan progress, not chain tip or head-first scan progress.

## Why

Callers that need to read authoritative wallet state at a specific height — for example, a snapshot height for voting power lookup, or any anchor-height balance read — currently have no good signal:

- `latestBlockHeight` comes from lightwalletd and reflects the chain tip, not local state.
- `maxScannedHeight` reflects head-first scan progress, which can skip ahead of birthday-first scanning under Spend-before-Sync.

Only `fullyScannedHeight` — already tracked in `LatestBlocksDataProvider` — is safe to compare against a fixed target height. This PR just surfaces it on `SynchronizerState` so downstream consumers get it through the existing `stateStream`/`latestState` plumbing without any new API surface.

## Source compatibility

The new initializer parameter has a default of `.zero`, so existing construction call sites (darkside tests etc.) keep compiling.

## Test plan

- [ ] CI green
- [ ] Existing offline/darkside tests pass (no behavior change for callers that don't read the new field)

### DIFF
--- a/Sources/ZcashLightClientKit/Synchronizer.swift
+++ b/Sources/ZcashLightClientKit/Synchronizer.swift
@@ -42,6 +42,13 @@ public struct SynchronizerState: Equatable {
     public var syncStatus: SyncStatus
     /// height of the latest block on the blockchain known to this synchronizer.
     public var latestBlockHeight: BlockHeight
+    /// Height below which every block has been scanned contiguously from the wallet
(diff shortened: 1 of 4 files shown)


greg0x:
closing in favor of #1696

PR #1696, opened 2026-05-04
[#1694] Add Synchronizer.getTreeState(height:) for app-layer consumers

### LINKED ISSUE #1694: Add Synchronizer.getTreeState(height:) API for app-layer commitment tree retrieval
## Is your feature request related to a problem? Please describe.

App-layer code that links `ZcashLightClientKit` sometimes needs to fetch a commitment tree state for a specific block height to hand to an external component — for example, witness generation in a Rust FFI consumed from app code. The first concrete consumer is the upcoming shielded-voting witness generation anchored at a poll snapshot height (see https://github.com/zcash/zcash-swift-wallet-sdk/pull/1687).

The only paths currently available to such a caller are either reaching into `Initializer.lightWalletService` (private SDK plumbing — `LightWalletService` is `internal` by design, and `Initializer.lightWalletService` is a `var` re-assigned on endpoint changes), or promoting the entire 17-method, protobuf-typed transport protocol to public. Both are bigger architectural changes than the need warrants.

## Describe the solution you'd like

Add a single typed accessor to the `Synchronizer` protocol that returns the protobuf-serialized `TreeState` bytes:

```swift
public protocol Synchronizer: AnyObject {
    /// Fetch the commitment tree state at the given block height from lightwalletd,
    /// returned as protobuf-serialized bytes suitable for witness generation.
    func getTreeState(height: UInt64) async throws -> Data
}

### PR DESCRIPTION
**Pull request by p0mvn**
Closes #1694

This PR is part of the Shielded Voting merge sequence, split for ease of review. Please view [this document](https://hackmd.io/@TJGUVkqNQYieiMqJQQBDeA/B1nzBrBA-g/edit) for details.

Adds a single typed accessor on the `Synchronizer` protocol so app-layer code that links `ZcashLightClientKit` can fetch a commitment tree state for a given height without reaching into the SDK's internal `LightWalletService`. Returns the protobuf-serialized `TreeState` bytes (`Data`), suitable for direct consumption by an external FFI (e.g. witness generation in the Rust voting stack).

```swift
public protocol Synchronizer: AnyObject {
    /// Fetch the commitment tree state at the given block height from lightwalletd,
    /// returned as protobuf-serialized bytes suitable for witness generation.
    ///
    /// Tor posture: when Tor is enabled on the Synchronizer, this uses a unique,
    /// one-shot Tor circuit per call (`.uniqueTor`), matching the policy of
    /// other public transport calls on this protocol. When Tor is disabled,
    /// this uses a direct gRPC connection.
    func getTreeState(height: UInt64) async throws -> Data
}
```

## Why this rather than wrap the underlying call at the call site

The full motivation lives in the commit message; the short version is that the SDK's transport (`LightWalletService`) is `internal` by design, and the call sites that need a tree state are app-layer (outside the SDK module). Without a public entry point, the only options were "reach into private SDK plumbing" or "promote the entire transport protocol to public." Both are bigger architectural changes than this need justifies. See _Alternatives considered_ below.

## ServiceMode: Tor posture

The `ServiceMode` argument to the underlying `LightWalletService.getTreeState` call is `await sdkFlags.ifTor(.uniqueTor)`, matching the policy of every other public `Synchronizer` → `lightWalletService` call (`fetchUTXOsBy`, `checkSingleUseTransparentAddresses`, `updateTransparentAddressTransactions`, the `importAccount` latest-height probe).

When Tor is disabled (the default), `ifTor(...)` collapses to `.direct` — so this costs exactly the same as a plain gRPC call for users who haven't opted into Tor.

When Tor is enabled, a fresh one-shot circuit per call keeps the tree-state fetch unlinkable from other SDK traffic — in particular, unlinkable from the subsequent `.txIdGroup`-scoped submission of any transaction anchored at the same height. For the primary consumer of this API (shielded-vote witness generation anchored at a poll snapshot), that decorrelation is the whole point: a `.direct` tree-state fetch followed by a Tor'd vote-tx submission would leak a "wallet at IP X interested in height H" signal right before lightwalletd sees "Tor'd tx anchored at H", linking the wallet to the vote.

## Source compatibility

The new requirement on the `Synchronizer` protocol ships with a throwing default in a `public extension`, so out-of-repo conformers (downstream test doubles, alternate transports, apps that implement the protocol directly) keep compiling on upgrade. `SDKSynchronizer` overrides the default with the real lightwalletd call.

## Tests

Two offline tests added in `SynchronizerOfflineTests`:
- `testSynchronizerGetTreeStatePassesHeightAndReturnsConfiguredData` — verifies the height argument flows through and the returned `Data` is propagated.
- `testSynchronizerGetTreeStatePropagatesError` — verifies thrown errors propagate to the caller.

`Tests/TestUtils/Sourcery/GeneratedMocks/AutoMockable.generated.swift` is updated with the new mock entry for `getTreeState(height:)`.

### 1. Expose `LightWalletService` (or an `Initializer.lightWalletService` getter) publicly

That's a huge public-API expansion for a one-method need.

The pattern in the existing protocol points the other way: `submitRawTransaction`, `proposeTransfer`, `getUnifiedAddress`, the resubmission machinery — every transport-backed capability on `Synchronizer` is already wrapped this way. This change follows that established pattern.

### 2. Have the consumer reach into `initializer.lightWalletService` directly

- The `Initializer` API is not a supported public surface — its `lightWalletService` is a `var`, re-assigned on endpoint changes (e.g. server switch), so any cached reference goes stale.
- It bypasses the SDK's encapsulation by design — a future change that, say, makes `LightWalletService` non-stored, lazily-resolved, or pool-managed would silently break the consumer.
- It requires the consumer to import / depend on protobuf-generated types (`BlockID`, `TreeState`).

### 3. Introduce a smaller, voting-specific protocol (e.g. `VotingDataSource`) instead of growing `Synchronizer`

```swift
public protocol VotingDataSource {
    func getTreeState(height: UInt64) async throws -> Data
}
extension SDKSynchronizer: VotingDataSource { /* same impl */ }
```

This keeps the public surface narrow and makes the dependency from the consumer explicit ("this needs a `VotingDataSource`, not a whole `Synchronizer`"). It's worth doing once there are 2–3 such transport-layer needs, but it's overkill for one method, and naming it now ("`VotingDataSource`?") would prematurely couple a public protocol name to a specific feature that the SDK shouldn't otherwise be aware of. Easy to refactor into later if the surface grows.

### 4. Keep the protocol requirement, drop the throwing default

Without the default, every out-of-repo `Synchronizer` conformer (downstream test doubles, alternate transports, apps that implement the protocol directly) would fail to compile on upgrade. That converts a strictly additive change into a source-breaking one. The default is a 6-line cost for full source compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Added
 - `SDKSynchronizer.rescanFrom(height:)`: Rescans the chain from the given BlockHeight.
 - `SynchronizerState.fullyScannedHeight`: Contiguous-from-birthday scan high-water mark published on `stateStream`/`latestState`. Callers that need an authoritative view of the wallet's note and nullifier state at a specific height (for example, balance anchored at a poll snapshot) should gate on this rather than `latestBlockHeight` (chain tip) or `maxScannedHeight` (head-first scan progress, which can race ahead under Spend-before-Sync).
+- `Synchronizer.getTreeState(height:)`: Fetches the commitment tree state at the given block height from lightwalletd and returns the protobuf-serialized `TreeState` bytes, for app-layer consumers that need to hand tree state to an external component (for example, witness generation via FFI). A throwing default implementation keeps the addition source-compatible for downstream `Synchronizer` conformers.
 
 ## Changed
 - Bumped Rust dependencies to current crates.io releases (`zcash_address` 0.10→0.11, `zcash_client_backend` 0.21→0.22, `zcash_client_sqlite` 0.19→0.20, `zcash_primitives`/`zcash_proofs` 0.26→0.27, `zcash_protocol` 0.7→0.8, `zcash_transparent` 0.6→0.7, `sapling-crypto` 0.6→0.7, `orchard` 0.12→0.13, `pczt` 0.5→0.6) and removed the `[patch.crates-io]` git-rev overrides. No public Swift API changes.
--- a/Sources/ZcashLightClientKit/Synchronizer.swift
+++ b/Sources/ZcashLightClientKit/Synchronizer.swift
@@ -494,6 +494,19 @@ public protocol Synchronizer: AnyObject {
     ///   hex-encoded bytes otherwise.
     func debugDatabase(sql: String) -> String
 
+    /// Fetch the commitment tree state at the given block height from lightwalletd,
+    /// returned as protobuf-serialized bytes suitable for witness generation.
+    ///
+    /// Tor posture: when Tor is enabled on the Synchronizer, this uses a unique,
+    /// one-shot Tor circuit per call (`.uniqueTor`), matching the policy of
+    /// other public transport calls on this protocol (`fetchUTXOsBy`,
+    /// `checkSingleUseTransparentAddresses`, `updateTransparentAddressTransactions`).
+    /// A fresh circuit keeps each fetch unlinkable from other SDK traffic — in
+    /// particular from later `.txIdGroup`-scoped submission of a transaction
+    /// anchored at the same height. When Tor is disabled, this uses a direct
+    /// gRPC connection.
+    func getTreeState(height: UInt64) async throws -> Data
+
     /// Get an ephemeral single use transparent address
     /// - Parameter accountUUID: The account for which the single use transparent address is going to be created.
     /// - Returns The struct with an ephemeral transparent address and gap limit info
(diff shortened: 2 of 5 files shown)


nuttycom (dismissed):
utACK

nuttycom:
Conflicts will need resolution

p0mvn:
On it

p0mvn:
@nuttycom conflict resolved. Thank you. Ready for final ACK.

Also granted access to our fork for any future convenience.

nuttycom (approved):
utACK

PR #1687, opened 2026-04-20
Shielded vote FFI integration

greg0x:
## Overview

See [Shielded Voting — Changes vs Upstream](https://hackmd.io/@greg-nagy/zodl-shielded-vote-integration) for the full cross-repo picture (librustzcash PRs, zcash_voting, this SDK layer, and the zodl-ios application layer).

## Summary

Adds the iOS SDK integration layer for shielded voting: a hand-rolled C FFI for `zcash_voting`, Swift wrappers, and the wallet↔voting data plumbing needed to drive the delegation → vote → share-submission lifecycle from a Swift client.

- **58 `extern "C"` FFI functions** in `rust/src/voting.rs` covering round management, wallet note queries, hotkey generation, delegation proof construction, vote commitment, share encryption, submission, crash-recovery state persistence, and commitment-tree sync
- **Swift wrappers** (`VotingRustBackend.swift`, `VotingTypes.swift`) with thread-safe handle management, `Codable` boundary types, and secret stripping at the FFI boundary
- **New SDK API** `Synchronizer.getTreeState(height:)` for commitment-tree witness generation, with a throwing default impl in a protocol extension so out-of-repo conformers remain source-compatible
- **`fullyScannedHeight` exposed on `SynchronizerState`** — callers that need authoritative wallet state at a snapshot height (voting power lookup) gate on this rather than the chain tip or head-first scan progress
- **Build integration** — `build.rs` watches all Rust source files for `cbindgen` rebuild; `SystemConfiguration` framework linked for macOS `reqwest` proxy detection

## Architecture

The SDK is the glue between `librustzcash` (wallet) and `zcash_voting` (voting):

zodl-ios (Swift/TCA)
  └─ zcash-swift-wallet-sdk (SPM)
       ├─ VotingRustBackend.swift (wraps C FFI)
       └─ libzcashlc.a (Rust staticlib)
            ├─ librustzcash ← wallet DB queries
            └─ zcash_voting ← voting DB, ZKP proofs, encryption

`zcash_voting` never touches the wallet DB. The SDK FFI queries notes/witnesses via `librustzcash` and passes them to `zcash_voting` as arguments. A small number of FFI functions do this wallet↔voting plumbing; the rest are thin pass-throughs to `zcash_voting`.

Complex types cross the FFI boundary via JSON serde (`serde_json` ↔ `Codable`); simple types use `#[repr(C)]` structs. Encrypted share secrets are stripped at the boundary — only ciphertext components are exposed to Swift.

## Voting workflow

[...]

Issue #1661, opened 2026-03-24
Integrate and expose v1 implementation of shielded voting


str4d:
This will enable wallets that use this SDK to interact with the Orchard-based Shielded Voting Protocol.

It is expected that this first version will have various limitations, such as requiring wallets to manually bake in details of each vote and release a corresponding update.

PR #1699, opened 2026-05-04
[#1661] Add zcash_voting dependency foundation

p0mvn:
Tracks #1661.

This PR is part of the Shielded Voting merge sequence (#1687), split for ease of review.

Brings the SDK Rust crate onto the released `zcash_voting 0.4` dependency line and the matching pinned `orchard 0.13.1` (with the `unstable-voting-circuits` feature), and lands a single pure-function FFI symbol that proves end-to-end linkage. The follow-up PRs in this sequence will then layer wallet-to-voting plumbing, the broader FFI surface, and the Swift wrappers on top of this foundation.

## Why this rather than landing more of the FFI now

Reviewing #1687 as a single 1571-line Swift wrapper plus a 3000-line `voting.rs` is hard. The FFI surface there mixes four concerns that a reviewer needs to evaluate independently:

|---|---|---|
| Does the SDK build against released `zcash_voting` and `orchard 0.13.1`? | Cargo graph resolves to single versions, FFI links | **this PR** |
| Wallet-to-voting plumbing (note snapshot, witness generation) | Wallet DB queries at historical heights are correct | follow-up |
| Voting DB lifecycle, JSON serde across FFI, ProgressContext lifetimes | Memory ownership and FFI safety | follow-up |
| Swift `VotingRustBackend` API surface | Public Swift contract, locking, error mapping | follow-up |

This PR isolates the first row. The diff is intentionally narrow and the new symbol is provably outside the wallet path.

## Version changes

| Crate          | Before | After                                                  |
| -------------- | ------ | ------------------------------------------------------ |
| orchard        | 0.13   | `=0.13.1`, `features = ["unstable-voting-circuits"]`   |
| zcash_voting   | new    | `0.4`, `default-features = false`                      |

Both are released on crates.io; no `[patch.crates-io]` entries are introduced and no fork URLs are pinned. The `orchard` `unstable-voting-circuits` feature is required transitively by `zcash_voting`.

## What the new FFI does — and what it deliberately does not do

`rust/src/voting/share_tracking.rs` exposes one symbol:

```rust
#[unsafe(no_mangle)]
pub unsafe extern "C" fn zcashlc_voting_compute_share_nullifier(
    vote_commitment: *const u8, // 32 bytes
    primary_blind:   *const u8, // 32 bytes
    share_index:     u32,
) -> *mut c_char // 64-char hex nullifier; caller frees via zcashlc_string_free
```

It forwards to `zcash_voting::share_tracking::compute_share_nullifier` and returns the 32-byte share-reveal nullifier as a 64-character hex C-string. On error or panic it returns `NULL`. The returned pointer must be freed by the caller via the SDK's existing `zcashlc_string_free`.

It intentionally stays outside the wallet-to-voting data path:

- No wallet DB open / read
- No voting DB open / read
- No network I/O (no PIR, no lightwalletd)
- No proving-key cache warm-up or other process-lifetime side effects
- No randomness
- No secret material in or out

The function is a deterministic computation over caller-supplied bytes. Its only purpose in this PR is to prove that the SDK Rust crate links and dispatches against `zcash_voting`.

## Source compatibility

No public Swift API changes. No new Swift wrapper. No `Synchronizer` protocol additions or default implementations. Existing constructor call sites and downstream conformers continue to compile unchanged.

## Out of scope (intentionally deferred to follow-ups)

- Swift `VotingRustBackend` and `VotingTypes` (the 1571 + 456 line files in #1687)
- Wallet note snapshot lookup, witness generation, voting DB lifecycle, PIR snapshot resolver
- Any `Package.swift` linker or binary target changes
- Any FFI symbol that takes a `wallet_db_path` or opens `WalletDb`

These will land as separate, individually reviewable PRs once this foundation is in.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1201,7 +1201,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dd5f2b7218a51c827a11d22d1439b598121fac94bf9b99452e4afffe512d78c9"
 dependencies = [
  "heck",
- "indexmap 2.14.0",
+ "indexmap 1.9.3",
  "itertools 0.14.0",
  "proc-macro-crate",
  "proc-macro2",
@@ -2228,7 +2228,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.62.2",
+ "windows-core",
 ]
 
 [[package]]
@@ -2530,6 +2530,7 @@ dependencies = [
  "zcash_protocol",
  "zcash_script",
  "zcash_transparent",
+ "zcash_voting",
  "zip32",
 ]
 
@@ -6049,6 +6050,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6a02e4885ed3bc0f2de90ea6dd45ebcbb66dacffe03547fadbb0eeae2770887d"
 
 [[package]]
+name = "voting-circuits"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2ba6635f8e207689c290634f6e34f7eddaf1b254aa0f9b0ef65418d89d1d1ae7"
+dependencies = [
+ "blake2b_simd",
+ "ff",
+ "group",
+ "halo2_gadgets",
+ "halo2_poseidon",
+ "halo2_proofs",
+ "incrementalmerkletree",
+ "lazy_static",
+ "orchard",
+ "pasta_curves",
+ "rand 0.8.6",
+ "sinsemilla",
+]
+
+[[package]]
 name = "walkdir"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6252,7 +6273,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9babd3a767a4c1aef6900409f85f5d53ce2544ccdfaa86dad48c91782c6d6893"
 dependencies = [
  "windows-collections",
- "windows-core 0.61.2",
+ "windows-core",
  "windows-future",
  "windows-link 0.1.3",
  "windows-numerics",
@@ -6264,7 +6285,7 @@ version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3beeceb5e5cfd9eb1d76b381630e82c4241ccd0d27f1a39ed41b2760b255c5e8"
 dependencies = [
- "windows-core 0.61.2",
+ "windows-core",
 ]
 
 [[package]]
@@ -6276,21 +6297,8 @@ dependencies = [
  "windows-implement",
  "windows-interface",
  "windows-link 0.1.3",
- "windows-result 0.3.4",
- "windows-strings 0.4.2",
-]
-
-[[package]]
-name = "windows-core"
-version = "0.62.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b8e83a14d34d0623b51dce9581199302a221863196a1dde71a7663a4c2be9deb"
-dependencies = [
- "windows-implement",
- "windows-interface",
- "windows-link 0.2.1",
- "windows-result 0.4.1",
- "windows-strings 0.5.1",
+ "windows-result",
+ "windows-strings",
 ]
 
 [[package]]
@@ -6299,7 +6307,7 @@ version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc6a41e98427b19fe4b73c550f060b59fa592d7d686537eebf9385621bfbad8e"
 dependencies = [
- "windows-core 0.61.2",
+ "windows-core",
  "windows-link 0.1.3",
  "windows-threading",
 ]
@@ -6344,7 +6352,7 @@ version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9150af68066c4c5c07ddc0ce30421554771e528bde427614c61038bc2c92c2b1"
 dependencies = [
- "windows-core 0.61.2",
+ "windows-core",
  "windows-link 0.1.3",
 ]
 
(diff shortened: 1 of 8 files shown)


rust/src/voting/helpers.rs
```diff
@@ -0,0 +1,14 @@
...
+
+/// Parse a byte slice from raw pointer + length.
+///
+/// # Safety
+///
+/// - `ptr` must be non-null and valid for reads for `len` bytes.
+/// - The memory referenced by `ptr` must not be mutated for the duration of the call.
+pub(super) unsafe fn bytes_from_ptr<'a>(ptr: *const u8, len: usize) -> &'a [u8] {
+    unsafe { std::slice::from_raw_parts(ptr, len) }
+}
```
p0mvn: Note to reviewer: there will be more helpers in this file and more files in the `voting` sub-module.

I am POCing the layout structure. Please feel free to peek into https://github.com/zcash/zcash-swift-wallet-sdk/pull/1687 for a view of the full integration

rust/src/voting/share_tracking.rs
```diff
@@ -0,0 +1,38 @@
...
+/// Compute the share reveal nullifier from client-known inputs.
+///
+/// Returns the 32-byte nullifier as a hex string (64 chars), or null on error.
+///
+/// # Safety
+///
+/// - `vote_commitment` must point to exactly 32 bytes.
+/// - `primary_blind` must point to exactly 32 bytes.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_voting_compute_share_nullifier(
```
p0mvn, line 20: Note to reviewer:

Please take a look at the full implementation [here](https://github.com/valargroup/zcash_voting/blob/0035bc0669f8f84fdfb6b46b76898c8e2253e1a6/zcash_voting/src/share_tracking.rs#L13-L73)

Context: the proposal's voting amount is split into 16 shares. The full amount is split into 16 shares. Each share has its own nullifier.

This is the function that enables the nullifier derivation for each share.

For more context, see [here](https://valargroup.gitbook.io/shielded-vote-docs/wallet-integration/swift-sdk#id-5.-share-delegation-and-status-tracking)

Cargo.lock
```diff
@@ -1201,7 +1201,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dd5f2b7218a51c827a11d22d1439b598121fac94bf9b99452e4afffe512d78c9"
 dependencies = [
  "heck",
- "indexmap 2.14.0",
+ "indexmap 1.9.3",
```
p0mvn, line 1204: Note to reviewer:

derive-deftly-macros 1.3.0 declares its indexmap requirement as:
```
indexmap = ">=1.8, <3"
```

That's a deliberately wide range that spans both major 1 and major 2. The maintainer accepts either; cargo gets to pick.

Both indexmap 1.9.3 and indexmap 2.14.0 are still in the lockfile.

Cargo.toml
```diff
@@ -78,6 +78,9 @@ rust_decimal = { version = "1", default-features = false, features = ["c-repr"]
 # - The "static" feature is required for the "compression" default feature of arti-client.
 xz2 = { version = "0.1", features = ["static"] }
 
+# Voting
+zcash_voting = { version = "0.4", default-features = false }
```
p0mvn, line 82: Note to reviewer: In a future PR, I will enable the custom set of features that this dependency will need.

This is an incremental, minimal scope import for ease of review

rust/src/voting/helpers.rs
nuttycom: Why is this function needed? Is it essentially just doing `std::slice::from_raw_parts` with type ascription?

p0mvn: Thanks for noting. That's correct. Mostly, ergonomics. We had this called from several places.

Image

However, I can see that the benefit that it brings is limited. To avoid blocking on this helper. I am going to remove it for now and avoid abstracting prematurely. We can revisit if this is helpful in future PRs, as we keep merging more work.

p0mvn: @nuttycom  Force pushed with an amendment

rust/CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Added
+- `zcashlc_voting_compute_share_nullifier`: Compute the 32-byte share-reveal
+  nullifier from a vote commitment, primary blind, and share index. Returns
+  the nullifier as a 64-character hex C-string; the caller must free the
+  returned pointer via `zcashlc_string_free`. Returns `NULL` on error or
+  panic. Pure-function FFI: no wallet DB, voting DB, network, randomness,
+  or secret material involved.
+
+### Changed
+- Added `zcash_voting 0.4` (`default-features = false`) as a Rust dependency.
```
nullcopy: (nit) IMO this should be under 'Added' not 'Changed'

Cargo.lock
nullcopy: I'd rather not see a downgrade like this if it didn't have a purpose, but I don't have a strong argument against, given your rationale 🤷

rust/src/voting/share_tracking.rs
```diff
@@ -0,0 +1,36 @@
...
+    share_index: u32,
+) -> *mut c_char {
+    let res = catch_panic(|| {
+        let vc = unsafe { std::slice::from_raw_parts(vote_commitment, 32) };
+        let blind = unsafe { std::slice::from_raw_parts(primary_blind, 32) };
+
+        let nullifier = voting::share_tracking::compute_share_nullifier(vc, share_index, blind)
+            .map_err(|e| anyhow!("compute_share_nullifier failed: {}", e))?;
+
+        let hex_str: String = nullifier.iter().map(|b| format!("{:02x}", b)).collect();
```
nullcopy: (non-blocking) rather than temporarily allocate 32 small strings, I'd prefer to preallocate the string (e.g. with `String::with_capacity(64)`) and then write the bytes into the string in a loop. Or just pull in the `hex` crate.

Not a huge deal, and definitely not blocking.

nullcopy (dismissed):
utACK 6a62dce0db47d92489693956122964f3f867f9d2

Cargo.lock
p0mvn: I'd rather not as well. But trying to force this dependency not to downgrade causes a messier blow-up of other dependencies. Based on what I'm seeing locally when trying to switch back to 2.14.0, the cleanest solution is keeping it as it is right now.

rust/src/voting/share_tracking.rs
p0mvn: Done and force pushed [092c381](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1699/commits/092c3817e28ee02febbba5a2d3bac0613010fb85)[092c381](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1699/commits/092c3817e28ee02febbba5a2d3bac0613010fb85)

rust/CHANGELOG.md
p0mvn: Done and force pushed [092c381](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1699/commits/092c3817e28ee02febbba5a2d3bac0613010fb85)[092c381](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1699/commits/092c3817e28ee02febbba5a2d3bac0613010fb85)

nullcopy (approved):
utACK 092c3817e28ee02febbba5a2d3bac0613010fb85

PR #1702, opened 2026-05-05
[zcash#1661] Add PIR proof validation FFI (zcashlc_voting_validate_pir_proof)

p0mvn:
Tracks #1661.

This PR is part of the Shielded Voting merge sequence (#1687), split for ease of review. It is the next incremental step after #1699: the foundation PR landed `zcash_voting` as a dependency and proved end-to-end FFI linkage with a single pure-function symbol. This PR is the first user of `zcash_voting`'s `client-pir` feature, which brings `pir-client` and its supporting crates into the SDK Cargo graph. Subsequent PRs in this sequence will layer richer voting functionality on top of that graph without re-litigating dependency churn in every review.

## Why this rather than landing more of the FFI now

#1699 brought in `zcash_voting` itself with no features enabled. Most of the Shielded Voting FFI surface needs `client-pir`, which transitively pulls in `pir-client`, `imt-tree`, `voting-circuits`, and `pir-types`. Reviewing those Cargo graph additions alongside DB lifecycle, `ProgressContext` lifetimes, network I/O, and Swift wrappers in a single PR (as in #1687) makes it very hard to evaluate them independently.

| --- | --- | --- |
| Does the SDK build with `zcash_voting`/`client-pir` and the released `pir-client` ecosystem? | Cargo graph resolves to single versions, FFI links | **this PR** |
| Wallet-to-voting plumbing (note snapshot, witness generation) | Wallet DB queries at historical heights are correct | follow-up |
| Voting DB lifecycle, JSON serde across FFI, `ProgressContext` lifetimes | Memory ownership and FFI safety | follow-up |
| Stateful PIR delegation / network I/O | Tor + lightwalletd integration | follow-up |
| Swift `VotingRustBackend` API surface | Public Swift contract, locking, error mapping | follow-up |

This PR isolates the first row. The diff is intentionally narrow and the new symbol is provably outside the wallet path and the network path.

## Version changes

| Crate | Before | After |
| --- | --- | --- |
| `zcash_voting` | 0.4 (no features) | 0.5.2 (`default-features = false`, `client-pir`) |
| `pir-client` | absent | 0.2 (transitive) |
| `imt-tree` | absent | 0.1 (transitive) |
| `voting-circuits` | absent | 0.4 (transitive) |
| `pir-types` | absent | 0.1 (transitive) |

`zcash_voting 0.5` introduced an upstream feature split that separates PIR (`client-pir`) from vote-commitment-tree sync (`client-tree-sync`). Only `client-pir` is enabled here. `vote-commitment-tree` and `vote-commitment-tree-client` stay excluded from the SDK lockfile until the tree-sync FFI lands in a later PR.

`pasta_curves` and `ff` are added as direct entries in `Cargo.toml`. Both are already in the resolved graph through `orchard` / `zcash_voting`; declaring them only grants this crate visibility for `pallas::Base` and `PrimeField::from_repr`. They do not introduce new versions.

All crates resolve from crates.io. No `[patch.crates-io]` entries are introduced and no fork URLs are pinned.

## What the new FFI does — and what it deliberately does not do

`rust/src/voting/delegation.rs` exposes one symbol:

```rust
#[unsafe(no_mangle)]
pub unsafe extern \"C\" fn zcashlc_voting_validate_pir_proof(
    root:          *const u8, // 32 bytes  (LE pallas::Base)
    nf_bounds:     *const u8, // 96 bytes  (3 × 32 LE pallas::Base)
    leaf_pos:      u32,
    path:          *const u8, // 928 bytes (29 × 32 LE pallas::Base)
    nullifier:     *const u8, // 32 bytes
    expected_root: *const u8, // 32 bytes
) -> i32
```

It parses the wire format of `zcash_voting::ImtProofData` and forwards to `zcash_voting::zkp1::validate_and_convert_pir_proof`, returning:

- `0` — proof is well-formed and authenticates the nullifier under the expected root,
- `1` — proof is well-formed but invalid (path/root mismatch, wrong leaf, etc.),
- `-1` — inputs are malformed (non-canonical field encoding) or the call panicked.

It intentionally stays outside the wallet-to-voting data path:

- No wallet DB open / read
- No voting DB open / read
- No network I/O (no PIR fetch, no lightwalletd)
- No proving-key cache warm-up or other process-lifetime side effects
- No randomness
- No secret material in or out

The function is a deterministic computation over caller-supplied bytes. Its purpose in this PR is to bring the `client-pir` dependency graph into the SDK and prove that `zcash_voting`'s PIR-validation entry point links and dispatches.

## Tests

Four `#[cfg(test)]` tests in `rust/src/voting/delegation.rs` cover:

- accepts a valid proof,
- rejects a root mismatch,
- rejects a corrupted path,
- rejects non-canonical field encoding.

The fixture is a precomputed proof generated from `zcash_voting`'s test-only `TestImt` helper. Storing it as bytes lets the SDK keep `[dev-dependencies]` empty: no direct test-only dependency on `halo2_gadgets` or `voting-circuits` is needed to construct a valid proof. Production coverage of the construction path lives in `zcash_voting` itself.

Verified locally with:

\`\`\`
cargo test -p libzcashlc voting::delegation
cargo metadata --locked --format-version 1
\`\`\`

## Source compatibility

No public Swift API changes. No new Swift wrapper. No `Synchronizer` protocol additions or default implementations. Existing constructor call sites and downstream conformers continue to compile unchanged.


rust/src/voting/delegation.rs
```diff
@@ -0,0 +1,213 @@
...
+                parse_base(&nf_bounds_bytes[64..96], "nf_bounds[2]")?,
+            ],
+            leaf_pos,
+            path: parse_path(path_bytes)?,
+        };
+
+        let nullifier = parse_base(&nullifier_bytes, "nullifier")?;
+        let expected_root = parse_base(&expected_root_bytes, "expected_root")?;
+
+        match zkp1::validate_and_convert_pir_proof(proof, nullifier, expected_root) {
```
p0mvn, line 75: Note to reviewer: [this](https://github.com/valargroup/zcash_voting/blob/9cdad63b7a676c68c7931318980317bfd47e5212/zcash_voting/src/zkp1.rs#L119-L128) is the entrypoint

[The core](https://github.com/valargroup/zcash_voting/blob/9cdad63b7a676c68c7931318980317bfd47e5212/vote-commitment-tree/src/path.rs#L61-L64) of the logic: compute merkle root from proof + nullifier and verify they match.

This is used for verifying the authentication path (nullifier non-membership) at a snapshot after receiving the response from the PIR server.

The logic here does not actually trigger PIR paths, only leverages its library's type APIs.

rust/src/voting/delegation.rs
```diff
@@ -0,0 +1,213 @@
...
...f6e35450e42c71abba35f1d0853a5c4faaba2861384d1538c031808c91140538602e8454283e9c5cfd47564c267f0815aae2d1dac4842d52f55784572f5a4ddbde392035bbe5619a86ec2db7dbca75ee6081b6dd6ab726f8254dd893ec76266b8b7dc66c70011f958767558a461a6143f0eb100693423819863eeddc19d02343311d5073ce3e931fdb19b745755a7e925818201c6346015827f3a7c07a65bd137df252fbd4379f6ef59601cd19d9c7d89d85634263cc04689b97d136dfa9f2457502788d5407d53d9a04c6d8d8732e7283f9f7b0a3531a728584a001839fc736a82d711de75d4d97b60a1432aa06873dbfada599a73a027fd25eefa6e305a354af3002c07fd283b5bdf1dc00502f0957ef3150ce9e020dfade15e2bfe919f9867c69c69b17c3aae833bf3f71fa6044748daab6779b020535813867047cdf120108e15fcc1257e42709fe6bfdcba82cb43c7be467562211564f02b6c295c7ee794a223f832c9aea620c634cd447c91a102497d1cf7b8a31e97207509990c7253c37300480fd747489cf99cf23d5ab7c7991d1a725714a092f0f453af80b9d7d6828742f9fad934eefea1cd3a281b396e40b3b804e27de3b6fc3b07e82930f463606951a5b0ddcf8b63e4cebf88387f4be2cca1446dd7f3715076183e96f5e260b2008e52fa71f57dbfb958ceaf42d99d54fdf6da7bd343b7ff965aeb8d2753f923ea9d1bf6df39de61763c145550f3748f049ba5a1bb42160420d736b7e3a9f172e40d98e3decff6a759ca472043254f7c639b9fcae001353d53603c500bc474aef03cde95a101a7dde4fccadb8379407e3f479044ef316";
```
p0mvn, line 110: Note to reviewer: an alternative required bringing in dev-dependencies.

I chose not; to keep the dependency graph cleaner by leveraging fixtures and linking to the generation source

rust/CHANGELOG.md
```diff
@@ -13,7 +13,10 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   returned pointer via `zcashlc_string_free`. Returns `NULL` on error or
   panic. Pure-function FFI: no wallet DB, voting DB, network, randomness,
   or secret material involved.
-- Added `zcash_voting 0.4` (`default-features = false`) as a Rust dependency.
+- `zcashlc_voting_validate_pir_proof`: Validate a PIR-fetched IMT
+  non-membership proof against an expected root.
+- Added `zcash_voting 0.5.2` (`default-features = false`, `client-pir`) as a
```
p0mvn, line 18: Note to reviewer: I bumped 0.5.2 as I have been improving the dependency graph to minimize what we're importing.

Started tracking CHANGELOG [here](https://github.com/valargroup/zcash_voting/blob/main/CHANGELOG.md) in case you'd like to review.

Cargo.lock
```diff
@@ -2183,6 +2183,22 @@ dependencies = [
  "want",
 ]
 
+[[package]]
+name = "hyper-rustls"
```
p0mvn, line 2187: Note to reviewer: this and all other `hyper` dependencies are needed for the HTTP client used by the PIR server and, in future PRs, for the vote-commitment tree HTTP client.

I have searched in `librustzcash` and have not found a working alternative. Previously, we were depending on `reqwest`, which is a more modern alternative, but pulls in a lot more noise into the dependency graph.

As a result, I replaced `reqwest` with `hyper` to keep a narrower deps focus.

Cargo.lock
```diff
@@ -2252,6 +2268,20 @@ version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9e0384b61958566e926dc50660321d12159025e767c18e043daf26b70104c39"
 
+[[package]]
+name = "imt-tree"
```
p0mvn, line 2272: Note: this is an imt-tree implementation we own [here](https://github.com/valargroup/vote-nullifier-pir/tree/main/imt-tree)

Key details:
- depth 29
- Pallas base field elements
- Poseidon parent hash `Poseidon(left, right)`
- leaf hash `Poseidon(nf_lo, nf_mid, nf_hi)`
- each leaf stores a punctured range `[nf_lo, nf_mid, nf_hi]`
- proof shows `nf_lo < nullifier < nf_hi` and `nullifier != nf_mid`

More on how it is used:
https://valargroup.gitbook.io/shielded-vote-docs/circuit-components/imt-non-membership

Cargo.lock
```diff
@@ -3182,6 +3215,38 @@ version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a89322df9ebe1c1578d689c92318e070967d1042b512afbe49518723f4e6d5cd"
 
+[[package]]
+name = "pir-client"
```
p0mvn, line 3219: Note to reviewer: this is the PIR-client located [here](https://github.com/valargroup/vote-nullifier-pir/tree/main/pir/client)

Context: https://valargroup.gitbook.io/shielded-vote-docs/appendices/pir-efficient-merkle-path-retrieval

Cargo.lock
```diff
@@ -6014,6 +6079,38 @@ dependencies = [
  "wasm-bindgen",
 ]
 
+[[package]]
+name = "valar-spiral-rs"
```
p0mvn, line 6083: Note: this is a dependency of the PIR library.

The repository is [here](https://github.com/valargroup/spiral-rs/commits/valar/avoid-avx512/).

It has been audited by Zellic.

See the report [here](https://github.com/valargroup/ypir/tree/valar/artifact/audits)

Cargo.lock
```diff
@@ -6014,6 +6079,38 @@ dependencies = [
...
+ "getrandom 0.2.17",
+ "rand 0.8.6",
+ "rand_chacha 0.3.1",
+ "serde_json",
+ "sha2 0.10.9",
+ "subtle",
+]
+
+[[package]]
+name = "valar-ypir"
```
p0mvn, line 6098: Note to reviewer: this is the primary PIR package.

Please see the audit report and code [here](https://github.com/valargroup/ypir/tree/valar/artifact/audits)

Cargo.toml
```diff
@@ -16,6 +16,8 @@ build = "rust/build.rs"
 
 # Zcash
 orchard = { version = "=0.13.1", features = ["unstable-voting-circuits"] }
+pasta_curves = "0.5"
+ff = "0.13"
```
p0mvn, line 20: Note to reviewer: these used to be imported transitively. Now, used directly for type conversions.

rust/src/voting/delegation.rs
```diff
@@ -0,0 +1,213 @@
...
+    path: *const u8,
+    nullifier: *const u8,
+    expected_root: *const u8,
+) -> i32 {
+    let res = catch_panic(|| {
+        let root_bytes: [u8; 32] = unsafe { std::slice::from_raw_parts(root, 32) }
+            .try_into()
+            .map_err(|_| anyhow!("root must be exactly 32 bytes"))?;
+        let nf_bounds_bytes = unsafe { std::slice::from_raw_parts(nf_bounds, 96) };
+        let path_bytes = unsafe { std::slice::from_raw_parts(path, 928) };
```
nullcopy: Is there a reason not to size-check `nf_bounds` and `path_bytes`, same as you do for the others?

rust/src/voting/delegation.rs
```diff
@@ -0,0 +1,213 @@
...
+
+fn parse_base(bytes: &[u8], label: &str) -> anyhow::Result<pallas::Base> {
+    let bytes: [u8; 32] = bytes
+        .try_into()
+        .map_err(|_| anyhow!("{label} must be exactly 32 bytes"))?;
+    Option::from(pallas::Base::from_repr(bytes))
+        .ok_or_else(|| anyhow!("{label} is not a canonical pallas::Base encoding"))
+}
+
+fn parse_path(bytes: &[u8]) -> anyhow::Result<[pallas::Base; 29]> {
```
nullcopy: Could you use named constants instead of these magic numbers. I get where they all come from (and I'd probably let the multiples of 32 slide, since they're fairly obvious), but this (and 968 above) is an example where the constant is protocol parameter that would greatly benefit from a name (maybe `NUM_PATH_ELEMENTS`?).

nullcopy (changes requested):
Looks good overall, but I have a few minor asks.

btw, I appreciate the notes you leave on the PR!

rust/src/voting/delegation.rs
p0mvn: Applied the same recommendation to `share_tracking.rs`
[e07db5f](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1702/commits/e07db5f391061460a10f0e2e7c16306c37d74e18)

It would be useful to add a linter and a formatter to the repo. Will look into it

rust/src/voting/delegation.rs
p0mvn: Addressed [03c3235](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1702/commits/03c3235dc571d440fb4b51a28f990ac84a119303)

nullcopy (approved):
utACK 03c3235dc571d440fb4b51a28f990ac84a119303

PR #1704, opened 2026-05-06
[#1661] Add PirSnapshotResolver for snapshot-equal PIR endpoint selection

p0mvn:
Tracks #1661.

This PR is part of the Shielded Voting merge sequence (#1687), split for ease of review. It adds a Swift-only PIR endpoint selection helper with no Rust, Cargo, FFI, wallet DB, or voting DB changes.

Wallet clients receive a list of PIR endpoints and an expected snapshot height for the active voting round. The resolver assumes those endpoints expose the [`vote-nullifier-pir`](https://github.com/valargroup/vote-nullifier-pir) `/root` API, where the response reports the served snapshot height. A delegation proof is bound to the round's exact snapshot, so using a server that is behind or ahead of the round can produce a nullifier non-membership proof for the wrong tree. This PR adds a small resolver that refuses to proceed unless an endpoint reports the exact expected snapshot height.

## Why this rather than landing more of the Swift voting surface now

The full Swift voting surface in #1687 includes wrappers over dozens of Rust FFI symbols, DB handle lifecycle, JSON models for proofs and commitments, progress callbacks, and wallet-to-voting call orchestration. That is too much to review as one Swift change.

| --- | --- | --- |
| PIR endpoint snapshot selection | `/root` probing, strict height equality, diagnostics, deterministic endpoint priority | **this PR** |
| Rust PIR proof validation | Cargo graph, FFI parsing, proof validation return codes | #1702 |
| Voting DB lifecycle and delegation PIR precompute | DB ownership, JSON serde across FFI, network-going Rust PIR path | valargroup/zcash-swift-wallet-sdk#52 |
| Swift voting wrappers and consumer wiring | Public Swift contract, locking, error mapping, note flow into Rust | follow-up |
| Progress callbacks and proving UX | C-callback lifetime, progress reporting, cancellation behavior | follow-up |

This PR isolates the first row. The resolver is fully testable without the Rust voting stack and can be reviewed as ordinary Swift networking/selection logic.

## What the resolver does — and what it deliberately does not do

`Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift` adds:

* `PirSnapshotResolver`
* `PirSnapshotResolverError`
* `PirSnapshotProbeOutcome`
* `PirSnapshotProbing`
* `HTTPPirSnapshotProbe`

The resolver probes all configured endpoints in parallel, decodes each endpoint's [`vote-nullifier-pir`](https://github.com/valargroup/vote-nullifier-pir) `GET /root` response, and returns the first endpoint in config order whose `height` is exactly equal to `expectedSnapshotHeight`.

Strict equality is intentional. A PIR server ahead of the round is just as wrong as one behind it because the proof is tied to the round's specific snapshot.

It intentionally does not add:

* No Rust or Cargo changes.
* No `zcash_voting` dependency changes.
* No FFI symbols.
* No wallet DB or voting DB access.
* No `Synchronizer` / `SDKSynchronizer` changes.
* No in-tree consumer yet. The consumer lands with the Swift voting wrapper split.

## Changelog

Adds an `Unreleased` changelog entry for the new public SDK types: `PirSnapshotResolver`, `PirSnapshotResolverError`, `PirSnapshotProbeOutcome`, `PirSnapshotProbing`, and `HTTPPirSnapshotProbe`.

## Tests

`Tests/OfflineTests/PirSnapshotResolverTests.swift` covers the resolver with an injected `StubProbe`, so the tests do not open the network:

* empty endpoint config throws `noEndpointsConfigured`,
* first matching endpoint in config order is selected,
* matching endpoint after a mismatch is selected,
* exact height match is accepted,
* height above expected is rejected,
* all mismatched endpoints throw with ordered diagnostics,
* missing height is excluded,
* unreachable endpoint is excluded,
* single matching endpoint is returned,
* `errorDescription` includes expected height and per-endpoint diagnostics.

Verified the resolver logic locally in a standalone SwiftPM harness containing only `PirSnapshotResolver.swift` and `PirSnapshotResolverTests.swift`:

```
swift test --filter PirSnapshotResolverTests
```

Result: 11 tests passed, 0 failures.

## Source compatibility

Strictly additive. No existing public API is changed, no protocol requirements are added, and no mock generator output needs to change. The new public types are isolated in their own Swift file and are not wired into existing SDK flows in this PR.


Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift
```diff
@@ -0,0 +1,244 @@
...
+    ///   - session: Optional `URLSession` to reuse. When `nil`, a new session
+    ///     is created with `timeout` applied to both request and (2×) resource
+    ///     timeouts. Pass a custom session in tests; in that case `timeout` is
+    ///     ignored because the session already carries its own configuration.
+    ///   - timeout: Per-request timeout in seconds for the default session.
+    public init(session: URLSession? = nil, timeout: TimeInterval = 5) {
+        if let session {
+            self.session = session
+        } else {
+            let config = URLSessionConfiguration.default
```
czarcas7ic: Minor point, I believe this uses default cache behavior, and I think in this particular instance of probing for fresh height, we should not use cache at all.

czarcas7ic (commented):
I think we need to leave changelog entries as well, specifically since we introduce new public SDK types (`PirSnapshotResolver`, `PirSnapshotResolverError`, `PirSnapshotProbeOutcome`, `PirSnapshotProbing`, `HTTPPirSnapshotProbe`)

Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift
p0mvn: Addressed in c46d963e: the default session now disables URL caching and uses `.reloadIgnoringLocalCacheData`, and each `/root` probe is sent as a `URLRequest` with `.reloadIgnoringLocalCacheData`. Added `HTTPPirSnapshotProbeTests.testProbeBypassesLocalCache` to lock this in.

p0mvn:
> I think we need to leave changelog entries as well, specifically since we introduce new public SDK types (`PirSnapshotResolver`, `PirSnapshotResolverError`, `PirSnapshotProbeOutcome`, `PirSnapshotProbing`, `HTTPPirSnapshotProbe`)

Addressed

Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift
```diff
@@ -0,0 +1,250 @@
...
+    }
+}
+
+/// Probes a single PIR endpoint's `/root` and reports its snapshot status.
+///
+/// Returning a `PirSnapshotProbeOutcome` (rather than throwing) lets the
+/// resolver collect per-endpoint diagnostics across the whole list before
+/// deciding to fail.
+public protocol PirSnapshotProbing: Sendable {
+    func probe(url: String, expectedSnapshotHeight: UInt64) async -> PirSnapshotProbeOutcome
```
nullcopy: Please use `BlockHeight` instead of `UInt64` (throughout this file)

Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift
```diff
@@ -0,0 +1,250 @@
...
+    public init(session: URLSession? = nil, timeout: TimeInterval = 5) {
+        if let session {
+            self.session = session
+        } else {
+            let config = URLSessionConfiguration.default
+            config.requestCachePolicy = .reloadIgnoringLocalCacheData
+            config.timeoutIntervalForRequest = timeout
+            config.timeoutIntervalForResource = timeout * 2
+            config.urlCache = nil
+            self.session = URLSession(configuration: config)
```
nullcopy: `URLSession` is constructed here, and stored in the struct. I don't see where this would get freed/released when the struct drops. Possible memory leak?

Tests/OfflineTests/PirSnapshotResolverTests.swift
```diff
@@ -0,0 +1,489 @@
...
+    func testProbeReturnsMismatchedWhenHeightBelowExpected() async {
+        StubURLProtocol.handler = { _ in (200, self.rootInfoJSON(height: 99)) }
+
+        let outcome = await makeProbe().probe(url: "https://pir.test", expectedSnapshotHeight: 100)
+
+        XCTAssertEqual(outcome.status, .mismatched(height: 99))
+    }
+
+    /// Strict-equality regression — `height > expected` must ALSO classify as
+    /// `.mismatched`. Under the previous `>=` policy this case would have been
```
nullcopy: > Under the previous `>=` policy

I'm not sure what previous policy this is referring to. Possible Claude-thoughts getting left in the comments?

Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift
```diff
@@ -0,0 +1,250 @@
...
+            for await item in group {
+                collected.append(item)
+            }
+            collected.sort { $0.0 < $1.0 }
+            return collected.map(\.1)
+        }
+
+        // All matching endpoints share the same height (= expected) by definition,
+        // so we just pick the first one in config order.
+        let chosen = outcomes.first { outcome in
```
nullcopy: Hmm, this sort makes the protocol easier to attack. A malicious/compromised PR server could return a "matching" result quickly, to always get selected by this probe, only to fail later and block the vote.

I would rather random selection here at minimum

nullcopy (changes requested):
Minor changes asked of c46d963e82c2888988c43c956adf667939ab6417

Tests/OfflineTests/PirSnapshotResolverTests.swift
p0mvn: Improved comment to:
```
Strict-equality: `height > expected` must classify as `.mismatched`
because proofs are bound to the round's exact snapshot.
```

Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift
p0mvn: Addressed

Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift
p0mvn: Proceeded with the recommendation

However, I don't think the original state was a meaningful risk. All servers return a constant. So there is not much that could be maliciously sped up.

Since clients query servers directly, attacking a specific client would require trying to geographically collocate. In this theory, it would be possible to DoS a particular client but not an entire vote.

Additionally, if someone is malicious, their PIR server could be socially removed from the config. The configuration is open source:
https://github.com/valargroup/token-holder-voting-config/blob/52a1d2cabfe4f9666242588a6aa91e1bb191d924/dynamic-voting-config.json#L13

Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift
p0mvn: If the session is owned, we now [free it](https://github.com/valargroup/zcash-swift-wallet-sdk/blob/3181f1e504f05bd05534c3e683262409b42825d8/Sources/ZcashLightClientKit/Rust/Voting/PirSnapshotResolver.swift#L190-L192)

We did not have a test for when a session is created/owned. Added [here](https://github.com/zcash/zcash-swift-wallet-sdk/blob/ac34343d7d710edc634183fcc59dffea890357ee/Tests/OfflineTests/PirSnapshotResolverTests.swift#L291-L309)

nullcopy (dismissed):
utACK ac34343d7d710edc634183fcc59dffea890357ee pending CI

nullcopy (approved):
utACK a51a921ed1937893f2ec460e9bae0190f7cc0b4b

PR #1711, opened 2026-05-07
[#1710] Root makefile for developer convenience

### LINKED ISSUE #1710: Root makefile for developer convenience
We have a Makefile in BuildSupport that has a build process more tailored to release.

For day-to-day developer convenience, it would be  helpful to have the Makefile be in the root while binding into the full init and rebuild scripts (ref: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1709)

p0mvn:
Adds a root `Makefile` with common developer targets (FFI init/rebuild/reset via existing scripts, `swift build`, and offline tests) so contributors can seamlessly run development convenience commands from the root for fast rebuild and verification.

This is especially useful when frequently switching/validating many branches.

Closes #1710

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,39 @@
+# Convenience development targets for this repository. Run from the repo root.
+# See BuildSupport/Makefile for the full FFI build graph.
+
+.DEFAULT_GOAL := help
+
+SCRIPTS := Scripts
+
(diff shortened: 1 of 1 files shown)


nullcopy (approved):
utACK 5f47b63df3cd9ba4142c753ce5fed470b66db147

PR #1714, opened 2026-05-07
[#1661] Add voting wallet-notes and Orchard FVK extraction FFI

p0mvn:
## Summary

Adds two FFI symbols on the path toward exposing shielded voting in the Swift layer. The Swift wrappers and end-to-end wiring will follow in subsequent PRs that consume these symbols.

- **`zcashlc_voting_get_wallet_notes`** — loads unspent Orchard notes for a wallet account at a snapshot height
- **`zcashlc_voting_extract_orchard_fvk_from_ufvk`** — decodes a UFVK string and returns the raw 96-byte Orchard FVK 

### Supporting changes

- Adds `zcash_keys 0.13` (`orchard` feature) as a Rust dependency, used by the new FFI to decode UFVKs and derive Orchard FVKs.
- Adds `received_note_to_note_info` and `open_wallet_db` helpers in `voting/helpers.rs` for the wallet-notes FFI.
- Adds `From<voting::NoteInfo> for JsonNoteInfo` so the Rust-side `voting::NoteInfo` round-trips into the existing JSON shape.
- Hoists `NETWORK_ID_TESTNET` / `NETWORK_ID_MAINNET` constants next to `parse_network` in `lib.rs` so FFI tests stop hardcoding 0/1.
- Adds a `#[cfg(test)] BoxedSlice::as_slice` accessor for crate-internal test inspection of FFI payloads.

Refs #1661.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7122,9 +7122,9 @@ dependencies = [
 
 [[package]]
 name = "zcash_voting"
-version = "0.5.2"
+version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d33a1676b7c56bc58724b5537d77c78b24b3eed39aecfcc79323809fe66cfc7f"
+checksum = "90abca13341b344d82315895bf50f26e93a66b5c4d2fbd8591728af604d4d4db"
 dependencies = [
  "anyhow",
  "blake2b_simd",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2558,6 +2558,7 @@ dependencies = [
  "zcash_address",
  "zcash_client_backend",
  "zcash_client_sqlite",
+ "zcash_keys",
  "zcash_note_encryption",
  "zcash_primitives",
  "zcash_proofs",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ rust_decimal = { version = "1", default-features = false, features = ["c-repr"] 
 xz2 = { version = "0.1", features = ["static"] }
 
 # Voting
-zcash_voting = { version = "0.5.2", default-features = false, features = [
+zcash_voting = { version = "0.5.3", default-features = false, features = [
     "client-pir",
     "client-tree-sync",
 ] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,6 +87,7 @@ zcash_voting = { version = "0.5.2", default-features = false, features = [
     "client-pir",
     "client-tree-sync",
 ] }
+zcash_keys = { version = "0.13", features = ["orchard"] }
 
 [build-dependencies]
 bindgen = "0.72"
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -42,7 +42,7 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   return the raw 96-byte Orchard full viewing key in a
   `*mut FfiBoxedSlice`. Returns `NULL` on missing Orchard component,
   malformed UFVK, or invalid `network_id`.
-- Added `zcash_voting 0.5.2` (`default-features = false`, `client-pir`,
+- Added `zcash_voting 0.5.3` (`default-features = false`, `client-pir`,
   `client-tree-sync`) as a Rust dependency.
 - Added `zcash_keys 0.13` (`orchard` feature) as a Rust dependency, used by
   the new wallet-notes and key-utility FFI for voting to decode UFVKs and derive
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -24,7 +24,7 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `zcashlc_voting_sync_vote_tree`: Sync the vote commitment tree for a round
   from a chain node URL, returning the latest synced block height (>= 0) on
   success, or -1 on error.
-- `zcashlc_voting_generate_van_witness`: Generate a vote authroity note Merkle witness for
+- `zcashlc_voting_generate_van_witness`: Generate a vote authority note Merkle witness for
   the second voting ZKP and return it as a JSON-encoded `VanWitness`
   (`auth_path`, `position`, `anchor_height`) in a `*mut FfiBoxedSlice`.
 - `zcashlc_voting_reset_tree_client`: Drop the in-memory tree client for a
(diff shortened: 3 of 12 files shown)


rust/src/voting/helpers.rs
```diff
@@ -44,6 +51,65 @@ pub(super) fn json_to_boxed_slice<T: Serialize>(
...
+        scope,
+        ufvk_str,
+    })
+}
+
+/// Open the wallet database.
+pub(super) fn open_wallet_db(
+    wallet_db_path: &str,
+    network_id: u32,
+) -> anyhow::Result<
```
p0mvn, line 105: Note to reviewer: the helper is small but it is used extensively in other functions so I kept it

rust/src/lib.rs
```diff
@@ -4137,13 +4137,23 @@ pub unsafe extern "C" fn zcashlc_tor_lwd_conn_check_single_use_taddr(
 // Utility functions
 //
 
+/// `network_id` value for Testnet, accepted by [`parse_network`] and every
+/// `zcashlc_*` FFI that takes a `network_id` parameter.
+pub(crate) const NETWORK_ID_TESTNET: u32 = 0;
+
+/// `network_id` value for Mainnet, accepted by [`parse_network`] and every
+/// `zcashlc_*` FFI that takes a `network_id` parameter.
+pub(crate) const NETWORK_ID_MAINNET: u32 = 1;
```
p0mvn, line 4146: Note to reviewer: had to use mainnet value in tests several times, so I extracted constants

Cargo.toml
```diff
@@ -84,6 +84,7 @@ xz2 = { version = "0.1", features = ["static"] }
 
 # Voting
 zcash_voting = { version = "0.5.2", default-features = false, features = ["client-pir"] }
+zcash_keys = { version = "0.13", features = ["orchard"] }
```
p0mvn, line 90: Note to reviewer: it was already in the dep tree. Now, we are just using keys directly. As a result, this addition was prompted

rust/src/voting/notes.rs
```diff
@@ -0,0 +1,213 @@
...
+        // Determine the target account ID.
+        let target_id = if !account_uuid.is_null() && account_uuid_len == 16 {
+            // Caller provided a 16-byte account UUID, use it directly.
+            let uuid_bytes = unsafe { bytes_from_ptr(account_uuid, account_uuid_len) }?;
+            let arr: [u8; 16] = uuid_bytes
+                .try_into()
+                .map_err(|_| anyhow!("account_uuid must be 16 bytes"))?;
+            zcash_client_sqlite::AccountUuid::from_uuid(uuid::Uuid::from_bytes(arr))
+        } else {
+            // If no account UUID is provided, use the first account in the wallet.
```
czarcas7ic: Minor point, I think we should fail closed for malformed UUID input. If the caller passes a UUID but the length is not 16 bytes, falling back to the first account can hide a caller bug and select the wrong account. The first account fallback seems appropriate only when no UUID is supplied.

rust/src/voting/helpers.rs
```diff
@@ -44,6 +51,65 @@ pub(super) fn json_to_boxed_slice<T: Serialize>(
...
+    let diversifier = orchard_note.recipient().diversifier().as_array().to_vec();
+    let value = orchard_note.value().inner();
+    let rho = orchard_note.rho().to_bytes().to_vec();
+    let rseed = orchard_note.rseed().as_bytes().to_vec();
+    let position = u64::from(note.note_commitment_tree_position());
+    let scope = match note.spending_key_scope() {
+        Scope::External => 0u32,
+        Scope::Internal => 1u32,
+    };
+    let ufvk_str = ufvk.encode(network);
```
czarcas7ic, line 86: I traced the network handling for the `get_wallet_notes -> precompute_delegation_pir` path.

The SDK uses `network_id` as:

- `0 = testnet`
- `1 = mainnet`

See: https://github.com/zcash/zcash-swift-wallet-sdk/blob/b15ceff2d3e9764dde0bc7b777a50bb57150f5c1/rust/src/lib.rs#L4140-L4146

`get_wallet_notes` uses that SDK network to open the wallet DB and serialize `NoteInfo.ufvk_str`.

But `zcash_voting 0.5.2` currently interprets it the other way around when deriving padded nullifiers:

- `0 = mainnet`
- `1 = testnet`

See: https://github.com/valargroup/zcash_voting/blob/329403fb6a097a5b7c37d59a91cfbc2d11e72846/zcash_voting/src/storage/operations.rs#L115-L117

This means the `NoteInfo` JSON can be produced with one network convention and then consumed by `zcash_voting` with the opposite convention.

Since we own `zcash_voting`, we should probably just align `zcash_voting` with the SDK convention (`0 = testnet`, `1 = mainnet`) everywhere it accepts `network_id`, then release/bump `zcash_voting` in this SDK PR.

rust/src/voting/notes.rs
p0mvn: Addressed. Was thinking the same. Traced to confirm that the app does not seem to rely on that path.

rust/src/voting/helpers.rs
p0mvn: Addressed:

https://github.com/valargroup/zcash_voting/pull/40

https://github.com/valargroup/zcash_voting/releases/tag/zcash_voting-v0.5.3

Cargo.lock
```diff
@@ -7121,9 +7122,9 @@ dependencies = [
 
 [[package]]
 name = "zcash_voting"
-version = "0.5.2"
+version = "0.5.3"
```
p0mvn, line 7125: Note to reviewer: please see [this thread](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1714#discussion_r3203303033) for motivation behind this bump

rust/src/voting/helpers.rs
p0mvn: [72b0ab9](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1714/commits/72b0ab94f853bf1c7df4f4c8d608ac98ff144141)

rust/src/voting/helpers.rs
```diff
@@ -44,6 +51,65 @@ pub(super) fn json_to_boxed_slice<T: Serialize>(
     Ok(crate::ffi::BoxedSlice::some(json))
 }
 
+/// Convert a librustzcash ReceivedNote (orchard) into zcash_voting's NoteInfo.
+///
+/// Requires the account's UFVK and network to compute the nullifier and
+/// encode the UFVK string.
+pub(super) fn received_note_to_note_info<P: consensus::Parameters>(
```
nullcopy, line 58: Is there a reason this isn't implemented as a constructor on `voting::NoteInfo`? Something like `NoteInfo::from_orchard_note(...)`?

This feels like the sort of base level functionality that should be provided by the crate, and not pushed into the FFI, but that's just my opinion.

nullcopy: Not a blocking comment, but I wanted to call it out

rust/CHANGELOG.md
```diff
@@ -24,16 +24,29 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
   `zcash_voting::tree_sync::VoteTreeSync`, constructed in
   `zcashlc_voting_db_open` and consumed by the tree-sync FFI above.
-- Added `zcash_voting 0.5.2` (`default-features = false`, `client-pir`) as a
-  Rust dependency.
+- `zcashlc_voting_get_wallet_notes`: Load unspent Orchard notes for a wallet
+  account at a snapshot height and return them as JSON-encoded
+  `Vec<NoteInfo>` in a `*mut FfiBoxedSlice`. `account_uuid` must be a non-null
+  pointer to exactly 16 bytes (binary account UUID). Returns `NULL` on error
+  or panic. Output is suitable as the `notes_json` input to
+  `zcashlc_voting_precompute_delegation_pir`.
```
nullcopy, line 40: FYI, prefer not to have a full contractual description of each method, and instead just give the highlevel changes meaningful to the SDK consumer.

nullcopy: not blocking

nullcopy (approved):
utACK 13e48643da54b14b728722df289202b9b6533db7

PR #1716, opened 2026-05-07
[#1661] Add voting FFI foundation helpers

czarcas7ic:
## Summary

Adds the shared Rust voting FFI foundation layer: JSON bridge types, C-compatible return structs/free functions, standalone utility FFI calls, and helper additions for wallet/key/voting conversions.

This intentionally leaves the direct incremental Merkle tree dependency and note-witness generation wiring for a follow-up PR, so this diff stays focused on the boilerplate foundation so the dependency add can be merged with no conflicts.

## Test plan

- `cargo check --locked`
- `AR=/usr/bin/ar cargo test --locked from_ptr --lib`
- `AR=/usr/bin/ar cargo clippy --all-targets -- -D warnings`


rust/src/voting/helpers.rs
```diff
@@ -1,13 +1,29 @@
...
 /// Borrow a byte slice from a raw `(ptr, len)` pair.
 ///
 /// When `len == 0`, returns an empty slice without reading `ptr`, so `ptr` may be null.
 ///
-/// Centralizing the null + length check here lets every voting FFI byte input — strings,
-/// JSON payloads, anything else — share one boundary contract instead of open-coding it
-/// per call site. `str_from_ptr` now delegates to this helper.
+/// Centralizing the null + length check here lets every voting FFI byte input -- strings,
+/// JSON payloads, anything else -- share one boundary contract instead of open-coding it
+/// per call site. `str_from_ptr` delegates to this helper.
```
p0mvn:
```suggestion
/// Centralizing the null + length check here lets every voting FFI byte input - strings,
/// JSON payloads, anything else - share one boundary contract instead of open-coding it
/// per call site. `str_from_ptr` delegates to this helper.
```

rust/src/voting/helpers.rs
```diff
@@ -44,9 +60,183 @@ pub(super) fn json_to_boxed_slice<T: Serialize>(
...
+    zcash_client_sqlite::WalletDb<rusqlite::Connection, Network, SystemClock, rand::rngs::OsRng>,
+> {
+    zcash_client_sqlite::WalletDb::for_path(
+        wallet_db_path,
+        Network::MainNetwork,
+        SystemClock,
+        rand::rngs::OsRng,
+    )
+    .map_err(|e| anyhow!("failed to open wallet DB for tree operations: {}", e))
+}
```
p0mvn:
```suggestion
```

Non-blocking: propose to omit this helper and use the logic directly. The abstraction is confusing

rust/src/voting/json.rs
```diff
@@ -47,14 +177,177 @@ impl From<voting::DelegationPirPrecomputeResult> for JsonDelegationPirPrecompute
...
+    pub hotkey_raw_address: Vec<u8>,
+    pub hotkey_public_key: Vec<u8>,
+    pub hotkey_address: String,
+    pub seed_fingerprint: Vec<u8>,
+}
+
 /// JSON-serializable VanWitness.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct JsonVanWitness {
-    /// The authentication path for the witness.
```
p0mvn, line 69: Not sure if intentional removal so noting

rust/src/voting/util.rs
```diff
@@ -0,0 +1,459 @@
...
+    hotkey_seed: *const u8,
+    hotkey_seed_len: usize,
+    network_id: u32,
+    account_index: u32,
+) -> *mut crate::ffi::BoxedSlice {
+    let res = catch_panic(|| {
+        let sender = unsafe { bytes_from_ptr(sender_seed, sender_seed_len) }?;
+        let hotkey = unsafe { bytes_from_ptr(hotkey_seed, hotkey_seed_len) }?;
+
+        if sender.len() < 32 {
```
p0mvn: nit: We should define a reusable constant. It was suggested in prior reviews

p0mvn (dismissed):

rust/src/voting/ffi_types.rs
```diff
@@ -0,0 +1,173 @@
...
+///
+/// - `ptr` must be non-null and must point to a struct returned by
+///   `zcashlc_voting_list_rounds`.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_voting_free_round_summaries(ptr: *mut FfiRoundSummaries) {
+    if !ptr.is_null() {
+        let s: Box<FfiRoundSummaries> = unsafe { Box::from_raw(ptr) };
+        if !s.ptr.is_null() {
+            let summaries =
+                unsafe { Box::from_raw(std::ptr::slice_from_raw_parts_mut(s.ptr, s.len)) };
```
nullcopy: Reconstructing ownership via Box::from_raw(slice_from_raw_parts_mut(s.ptr, s.len)), means producers MUST allocate via Box<[T]>::into_raw (e.g. vec.into_boxed_slice() then Box::into_raw). Otherwise, this could cause undefined behavior. e.g. if `FfiRoundSummaries` was constructed with `Vec::into_raw_parts()`, free would be undefined, because this function assumes the allocated size is strictly a multiple of the size of the entries, when in reality its common for vectors to overallocate.

This is not a problem if you construct via `Vec::into_boxed_slice()` then `Box::into_raw()`, or something along those lines, but I have no way of knowing that atm.

My suggestion: go ahead and pull in the constructors for these types, so we can be sure we are deallocating appropriately.

rust/src/voting/ffi_types.rs
```diff
@@ -0,0 +1,173 @@
...
+///
+/// - `ptr` must be non-null and must point to a struct returned by
+///   `zcashlc_voting_get_votes`.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_voting_free_vote_records(ptr: *mut FfiVoteRecords) {
+    if !ptr.is_null() {
+        let s: Box<FfiVoteRecords> = unsafe { Box::from_raw(ptr) };
+        if !s.ptr.is_null() {
+            let records =
+                unsafe { Box::from_raw(std::ptr::slice_from_raw_parts_mut(s.ptr, s.len)) };
```
nullcopy: Same comment as above, in `zcashlc_voting_free_round_summaries`

rust/src/voting/helpers.rs
```diff
@@ -1,20 +1,29 @@
+#![allow(dead_code)]
```
nullcopy: I would prefer not to add a module-wide `#![allow(dead_code)]`. For now, please allow each function specifically, and they can each be removed when their corresponding call-sites land in a future PR.

rust/src/voting/json.rs
```diff
@@ -1,7 +1,13 @@
+#![allow(dead_code)]
```
nullcopy: Same comment about allowing dead code

rust/src/voting/helpers.rs
```diff
@@ -110,9 +119,99 @@ pub(super) fn open_wallet_db(
...
+        VoteReady => 4,
+    }
+}
+
+pub(super) fn usk_from_seed(
+    network_id: u32,
+    seed: &[u8],
+    account: AccountId,
+) -> anyhow::Result<UnifiedSpendingKey> {
+    let network = crate::parse_network(network_id)?;
```
nullcopy, line 148: Is there reason not to have a min seed len check here? I assume there will never be a case where we are ok with the caller constructing a seed from less than `MIN_SEED_LEN` bytes, right?

rust/src/voting/util.rs
```diff
@@ -40,10 +252,62 @@ pub unsafe extern "C" fn zcashlc_voting_extract_orchard_fvk_from_ufvk(
...
+///
+/// `witness_json` is a JSON-encoded `WitnessData`.
+///
+/// Returns 1 if valid, 0 if invalid, -1 on error.
+///
+/// # Safety
+///
+/// - `witness_json` must be valid for reads of `witness_json_len` bytes.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_voting_verify_witness(
```
nullcopy, line 267: Missing test coverage

rust/src/voting/util.rs
```diff
@@ -40,10 +252,62 @@ pub unsafe extern "C" fn zcashlc_voting_extract_orchard_fvk_from_ufvk(
...
 
+/// Extract the Orchard note commitment tree root from a protobuf-encoded TreeState.
+///
+/// Returns the 32-byte nc_root as `*mut FfiBoxedSlice`, or null on error.
+///
+/// # Safety
+///
+/// - `tree_state_bytes` must be valid for reads of `tree_state_bytes_len` bytes.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_voting_extract_nc_root(
```
nullcopy, line 240: Missing test coverage

rust/src/voting/util.rs
```diff
@@ -1,15 +1,227 @@
...
+
+/// Extract a spend auth signature from a signed PCZT.
+///
+/// Returns the signature bytes as `*mut FfiBoxedSlice`, or null on error.
+///
+/// # Safety
+///
+/// - `signed_pczt_bytes` must be valid for reads of `signed_pczt_bytes_len` bytes.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_voting_extract_spend_auth_sig(
```
nullcopy, line 188: Missing test coverage

rust/src/voting/util.rs
```diff
@@ -1,15 +1,227 @@
...
+
+/// Extract the ZIP-244 shielded sighash from finalized PCZT bytes.
+///
+/// Returns the 32-byte sighash as `*mut FfiBoxedSlice`, or null on error.
+///
+/// # Safety
+///
+/// - `pczt_bytes` must be valid for reads of `pczt_bytes_len` bytes.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_voting_extract_pczt_sighash(
```
nullcopy, line 167: Missing test coverage

nullcopy (changes requested):
Changes requested as of 8d60be1d42107662e4a9696c370233ecac94843f

rust/src/voting/ffi_types.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1716/commits/feb5d7a948bfe725437c7feac750791b3a766ec1

Makes sense. Added constructors so allocation is explicit, and updated the free functions to use the matching repo cleanup helpers. Also applied the same fix to `FfiVoteRecords`.

rust/src/voting/ffi_types.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1716/commits/feb5d7a948bfe725437c7feac750791b3a766ec1

rust/src/voting/helpers.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1716/commits/1869e0645f39e9c253990e184cf8c40f08829dce

Moved the dead-code allowance from the module level to the two helper functions that currently need it.

rust/src/voting/json.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1716/commits/92c4dfaf64390b86bfb7a0e42a9c7df6709d0886

Same done here

rust/src/voting/helpers.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1716/commits/5e854fa2b8afb4488c4f4f82c9f32e2d5b71fff2

Moved the minimum seed length check into `usk_from_seed`, so every caller gets the same validation before deriving a USK.

rust/src/voting/util.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1716/commits/d6065e1af8b8ebac6879cefbec542e0b28d87e44

Added here

rust/src/voting/util.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1716/commits/d6065e1af8b8ebac6879cefbec542e0b28d87e44

Added here

rust/src/voting/util.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1716/commits/d6065e1af8b8ebac6879cefbec542e0b28d87e44

Added here

rust/src/voting/util.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1716/commits/d6065e1af8b8ebac6879cefbec542e0b28d87e44

Added here

nullcopy (approved):
utACK d6065e1af8b8ebac6879cefbec542e0b28d87e44

PR #1717, opened 2026-05-07
[#1661] Add voting generate-note-witnesses FFI

p0mvn:
Tracks zcash/zcash-swift-wallet-sdk#1661

## Summary

Adds `zcashlc_voting_generate_note_witnesses`, the next FFI step in the shielded voting wiring (#1661). It generates Orchard Merkle inclusion witnesses for the notes in a voting bundle, anchored at the round's snapshot height, caches them in the voting DB, and returns them JSON-encoded as `Vec<WitnessData>`.

## Dependencies

Adds `incrementalmerkletree 0.8` (`default-features = false`) as a direct Rust dependency, used for `Position` and the `MerklePath` returned by the wallet DB. The crate was already in the transitive graph via `zcash_client_sqlite`/`zcash_voting`; this just makes the use site explicit.

## API surface

- New: `zcashlc_voting_generate_note_witnesses` (Rust FFI only).
- New JSON shape `JsonWitnessData` and `From<voting::WitnessData> for JsonWitnessData` impl in `voting/json.rs`.
- No Swift API surface added in this step.


Cargo.lock
```diff
@@ -453,7 +453,7 @@ version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
 dependencies = [
- "bitflags 2.11.1",
+ "bitflags",
```
p0mvn, line 456: Note to reviewer: the diff in `Cargo.lock` is an outcome of `cargo check`.

This stems from the fact that we added `incrementalmerkletree`. The new dependency has resulted in updating the graph.

To validate, switch to main, add the diff in `Cargo.toml`. Run `cargo check`. The diff may be slightly different, depending on the versions in your local environment. But you should see that related deps are changed. 

We use `incrementalmerkletree` to generate witnesses for a balance proof. So that, the Zcash mainnet note value can be proven for use in voting on a separate L2.

rust/src/voting/delegation.rs
```diff
@@ -4,14 +4,163 @@ use std::sync::Arc;
...
+        let orchard_ct = tree_state
+            .orchard_tree()
+            .map_err(|e| anyhow!("failed to parse orchard tree from TreeState: {}", e))?;
+        let frontier_root = orchard_ct.root();
+        let frontier = orchard_ct.to_frontier();
+        let nonempty_frontier = frontier.take().ok_or_else(|| {
+            anyhow!("empty orchard frontier — no orchard activity at snapshot height")
+        })?;
+
+        // Open the wallet DB
```
czarcas7ic: We should be checking if the decoded tree state is the tree state for the round, something like:

```
if tree_state.height != params.snapshot_height {
    return Err(anyhow!(
        "cached TreeState height {} does not match round snapshot_height {}",
        tree_state.height,
        params.snapshot_height
    ));
}

if frontier_root.to_bytes().as_ref() != params.nc_root.as_slice() {
    return Err(anyhow!("cached TreeState orchard root does not match round nc_root"));
}
```

Otherwise a stale cached `TreeState` can produce witnesses that verify against their own embedded root and get cached, but are not anchored to the round’s actual `snapshot_height` / `nc_root`. `store_witnesses` verifies the path, not that the root belongs to this round.

rust/src/voting/delegation.rs
```diff
@@ -4,14 +4,163 @@ use std::sync::Arc;
...
+        let snapshot_height = u32::try_from(params.snapshot_height).map_err(|_| {
+            anyhow!(
+                "snapshot_height {} does not fit in u32",
+                params.snapshot_height
+            )
+        })?;
+        let checkpoint_height = zcash_protocol::consensus::BlockHeight::from_u32(snapshot_height);
+
+        // Generate witnesses from wallet DB shard data + frontier
+        let merkle_paths = wallet_db
```
czarcas7ic, line 141: Can we add a real success-path test for this call? The current tests only check that bad inputs fail AFAIU

rust/src/voting/delegation.rs
p0mvn: Addressed

rust/src/voting/delegation.rs
p0mvn: Addressed

p0mvn:
Resolving conflict

rust/src/voting/delegation.rs
```diff
@@ -4,14 +4,194 @@ use std::sync::Arc;
...
+                    "generate_orchard_witnesses_at_historical_height failed: {}",
+                    e
+                )
+            })?;
+
+        // Convert MerklePaths to WitnessData
+        let root_bytes = frontier_root_bytes.to_vec();
+        let witnesses: Vec<voting::WitnessData> = merkle_paths
+            .into_iter()
+            .zip(core_notes.iter())
```
nullcopy, line 166: Should check `core_notes.len() == merkle_paths.len()` before zipping their iterators

rust/src/voting/delegation.rs
```diff
@@ -4,14 +4,194 @@ use std::sync::Arc;
...
+        let frontier_root = orchard_ct.root();
+        let frontier_root_bytes = frontier_root.to_bytes();
+        validate_cached_tree_state_for_round(&tree_state, &frontier_root_bytes[..], &params)?;
+        let frontier = orchard_ct.to_frontier();
+        let nonempty_frontier = frontier.take().ok_or_else(|| {
+            anyhow!("empty orchard frontier — no orchard activity at snapshot height")
+        })?;
+
+        // Open the wallet DB
+        let wallet_db = zcash_client_sqlite::WalletDb::for_path(
```
nullcopy: Any reason not to use `super::helpers::open_wallet_db`? That would also wrap some of the checks you performed above (e.g. the network check)

rust/src/voting/delegation.rs
```diff
@@ -4,14 +4,194 @@ use std::sync::Arc;
...
+            .map_err(|e| anyhow!("store_witnesses failed: {}", e))?;
+
+        let json_witnesses: Vec<JsonWitnessData> = witnesses.into_iter().map(Into::into).collect();
+        json_to_boxed_slice(&json_witnesses)
+    });
+    unwrap_exc_or_null(res)
+}
+
+// =============================================================================
+// VotingDatabase methods — Delegation proof
```
nullcopy: (nit) Shouldn't `zcashlc_voting_generate_note_witnesses` live in this section?

rust/src/voting/delegation.rs
```diff
@@ -4,14 +4,194 @@ use std::sync::Arc;
...
+        )
+        .map_err(|e| anyhow!("failed to open wallet DB for tree operations: {}", e))?;
+
+        // Convert note positions to Merkle positions
+        let positions: Vec<Position> = core_notes
+            .iter()
+            .map(|n| Position::from(n.position))
+            .collect();
+
+        // Generate witnesses from wallet DB shard data + frontier
```
nullcopy: Copy pasta from L146?

nullcopy (changes requested):
In addition to the comments below, I think it'd be good to tighten up test coverage. Notable gaps:
* no test for multi-note bundles
* end-to-end stale-TreeState rejection through the FFI
* empty-frontier rejection
* notes_json_len == 0 happy path against a fully-set-up DB.

p0mvn:
Thanks, addressing

rust/src/voting/delegation.rs
p0mvn: Done [a13062b](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1717/commits/a13062b71f3486397d4f15c3a57f5a0e5bcd4dac)

rust/src/voting/delegation.rs
p0mvn: I believe we introduced this later by a concurrent PR. But good idea, switched to this

[3868349](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1717/commits/3868349d1e007abfc256672ca35a0a0d37c283e1)

rust/src/voting/delegation.rs
p0mvn: [3868349](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1717/commits/3868349d1e007abfc256672ca35a0a0d37c283e1)

rust/src/voting/delegation.rs
p0mvn: [31ab9d0](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1717/commits/31ab9d0aae461d66fe0f30aea136232cbe3d453b)

nullcopy (approved):
utACK 31ab9d0aae461d66fe0f30aea136232cbe3d453b

PR #1719, opened 2026-05-07
[#1661] Voting FFI in VotingRustBackend.swift (database, PR, VAN witness gen)

p0mvn:
Tracks #1661.

This PR lands the next batch of `VotingRustBackend` Swift wrappers.

It also lays the structural groundwork (handle lifecycle + private helpers) that the remaining splits from [#1715](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1715) will build on incrementally, one Rust submodule at a time.

## Summary

- Convert `VotingRustBackend` from a caseless namespace `enum` to a `public final class` that owns an opaque `VotingDatabaseHandle`.
- Add `init()` / `open(path:)` / `close()` (idempotent) / `deinit` over `zcashlc_voting_db_open` and `zcashlc_voting_db_free`. Handle access is serialized by an `NSLock`.
- Wrap the FFI that already lives on `main` but had no Swift entry point:
  - `setWalletId(_:)` → `zcashlc_voting_set_wallet_id`
  - `precomputeDelegationPir(...)` → `zcashlc_voting_precompute_delegation_pir`, with `PirSnapshotResolver`-driven endpoint selection
  - `syncVoteTree(roundId:nodeUrl:)` → `zcashlc_voting_sync_vote_tree`
  - `generateVanWitness(roundId:bundleIndex:anchorHeight:)` → `zcashlc_voting_generate_van_witness`
  - `resetTreeClient(roundId:)` → `zcashlc_voting_reset_tree_client`
- Add `VotingRustBackendError.databaseAlreadyOpen` / `.databaseNotOpen` for handle-state errors. Existing `.rustError` and `.invalidData` are unchanged.
- Add private helpers (`requireHandle()`, `decodeJSON(from:)`, `lastErrorMessage(fallback:)`) that subsequent voting wrappers will reuse unchanged.

No Swift API breakage at any existing call site.

## Tests

`Tests/OfflineTests/VotingRustBackendTests.swift` keeps the existing 2 nullifier tests and adds 11 new ones, all running fully offline.

```bash
swift test --filter OfflineTests.VotingRustBackendTests
```

13 tests, 0 failures locally on macOS arm64.

### DIFF
--- a/Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
+++ b/Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
@@ -35,10 +35,225 @@ public enum VotingRustBackendError: LocalizedError, Equatable {
 
 // MARK: - VotingRustBackend
 
-/// Swift wrapper for the voting `libzcashlc` surface.
-public enum VotingRustBackend {}
+/// Wraps the voting `libzcashlc` C FFI surface.
+///
+/// Manages an opaque `VotingDatabaseHandle` pointer for the database-bound
+/// methods. Stateless / static FFI (e.g. `computeShareNullifier`) is exposed
+/// as type methods so callers do not need to open a database.
+///
+/// Thread safety: handle access is serialized by an `NSLock`. The lock guards
+/// the handle slot only — it is released before the FFI call runs, so callers
+/// must not race `close()` against in-flight operations.
+public final class VotingRustBackend: @unchecked Sendable {
+    private let lock = NSLock()
+    private var handle: OpaquePointer?
 
-// MARK: - Share tracking
+    public init() {}
+
+    deinit {
+        if let handle {
+            zcashlc_voting_db_free(handle)
+        }
+    }
+
+    // MARK: - Database lifecycle
+
+    /// Open the voting database at `path`.
+    ///
+    /// Throws `VotingRustBackendError.databaseAlreadyOpen` if the backend
+    /// already holds an open handle.
+    public func open(path: String) throws {
+        lock.lock()
+        defer { lock.unlock() }
+
+        guard handle == nil else {
+            throw VotingRustBackendError.databaseAlreadyOpen
+        }
+
+        let pathBytes = [UInt8](path.utf8)
+        guard let ptr = pathBytes.withUnsafeBufferPointer({ buf in
+            zcashlc_voting_db_open(buf.baseAddress, UInt(buf.count))
+        }) else {
+            throw VotingRustBackendError.rustError(
+                Self.staticLastErrorMessage(fallback: "`voting_db_open` failed")
+            )
+        }
+        handle = ptr
+    }
+
+    /// Close the voting database, freeing the underlying handle.
+    ///
+    /// Idempotent: calling `close()` on an already-closed backend is a no-op.
+    public func close() {
+        lock.lock()
+        defer { lock.unlock() }
+
+        if let dbh = handle {
+            zcashlc_voting_db_free(dbh)
+            handle = nil
+        }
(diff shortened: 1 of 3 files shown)


Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
```diff
@@ -91,6 +306,28 @@ extension VotingRustBackend {
 // MARK: - Private helpers
 
 private extension VotingRustBackend {
+    /// Snapshot the current handle under the lock, throwing if the database
+    /// is not open. The pointer is read-only after `close()` resets it, so the
+    /// snapshot is valid for the duration of the FFI call as long as callers
+    /// do not race `close()` against other operations.
+    func requireHandle() throws -> OpaquePointer {
```
czarcas7ic: I think there’s still a handle lifetime race here. `requireHandle()` reads the raw Rust handle under the lock, but releases the lock before the FFI call uses it. Another thread can call `close()` in between, free the handle, and leave the first call using a freed pointer.

So the lock protects the Swift `handle` slot, but not the lifetime of the Rust handle during an operation. Since this class is `@unchecked Sendable`, I think `close()` needs to be prevented from freeing the handle while any FFI call is in flight.

Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
```diff
@@ -35,10 +35,225 @@ public enum VotingRustBackendError: LocalizedError, Equatable {
...
+        notes: [VotingNoteInfo],
+        pirEndpoints: [String],
+        expectedSnapshotHeight: UInt64,
+        networkId: UInt32,
+        pirResolver: PirSnapshotResolver = PirSnapshotResolver()
+    ) async throws -> VotingDelegationPirPrecomputeResult {
+        // PirSnapshotResolver expects `BlockHeight` (Int); voting snapshot
+        // heights are `UInt64` everywhere else in the voting types, so convert
+        // at the boundary. Snapshot heights well within Int.max in practice.
+        let pirServerUrl = try await pirResolver.resolve(
```
czarcas7ic, line 146: Can we check that the DB is open before doing the PIR resolver work here? Right now this can still probe endpoints even if the backend is closed, and in some cases it will return a resolver error instead of “database not open.”

It seems cleaner if a closed backend fails immediately, like the other instance methods do.

p0mvn: Done

Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
p0mvn: [c36b98d](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1719/commits/c36b98d7c5e1b69a303caae040332f5777cf8ee5)

Added test

Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
```diff
@@ -108,3 +362,13 @@ private extension VotingRustBackend {
...
+
+#if DEBUG
+extension VotingRustBackend {
+    func withLockedHandleForTesting(_ operation: () -> Void) throws {
+        try withHandle { _ in
+            operation()
+        }
+    }
+}
+#endif
```
p0mvn, line 374: Note to reviewer: this is to simulate a locked handle for a concurrency test. Only used in tests and surrounded by the DEBUG flag for safety.

LukasKorba (approved):
utACK

PR #1722, opened 2026-05-07
[#1661] Add voting round recovery FFI

czarcas7ic:
Adds Rust-only voting FFI for round lifecycle/state inspection, recovery metadata persistence, and share-delegation tracking.

Validated with `cargo check --locked` and `cargo clippy --all-targets -- -D warnings`.


rust/src/voting/recovery.rs
```diff
@@ -0,0 +1,498 @@
...
+    sighash_len: usize,
+    rk: *const u8,
+    rk_len: usize,
+) -> i32 {
+    let db = AssertUnwindSafe(db);
+    let res = catch_panic(|| {
+        let handle =
+            unsafe { db.as_ref() }.ok_or_else(|| anyhow!("VotingDatabaseHandle is null"))?;
+        let round_id_str = unsafe { str_from_ptr(round_id, round_id_len) }?;
+        if sig_len != 64 {
```
p0mvn: Consider defining constants for numbers

rust/src/voting/recovery.rs
```diff
@@ -0,0 +1,498 @@
...
+            .get_keystone_signatures(&round_id_str)
+            .map_err(|e| anyhow!("get_keystone_signatures failed: {}", e))?;
+
+        #[derive(serde::Serialize)]
+        struct SigOut {
+            bundle_index: u32,
+            sig: Vec<u8>,
+            sighash: Vec<u8>,
+            rk: Vec<u8>,
+        }
```
p0mvn: Would double-check style if the rest of the code defines types within functions as opposed to in the global space

rust/src/voting/recovery.rs
```diff
@@ -0,0 +1,498 @@
...
+    fn open_memory_db() -> *mut VotingDatabaseHandle {
+        let path = b":memory:";
+        let db = unsafe { zcashlc_voting_db_open(path.as_ptr(), path.len()) };
+        assert!(!db.is_null());
+
+        let wallet = b"wallet";
+        let code = unsafe { zcashlc_voting_set_wallet_id(db, wallet.as_ptr(), wallet.len()) };
+        assert_eq!(code, 0);
+
+        db
```
p0mvn: Just a thought, we should do a pass after everything is merged to ensure that we do not redefine the same helpers across every file

rust/src/voting/recovery.rs
```diff
@@ -0,0 +1,498 @@
...
+                42,
+            )
+        };
+
+        assert_eq!(code, -1);
+        unsafe { zcashlc_voting_db_free(db) };
+    }
+
+    #[test]
+    fn store_keystone_signature_rejects_invalid_lengths() {
```
p0mvn, line 506: Should there be a happy path case? Not sure I saw one

rust/src/voting/rounds.rs
```diff
@@ -0,0 +1,380 @@
...
+        let path = b":memory:";
+        let db = unsafe { zcashlc_voting_db_open(path.as_ptr(), path.len()) };
+        assert!(!db.is_null());
+
+        let wallet = b"wallet";
+        let code = unsafe { zcashlc_voting_set_wallet_id(db, wallet.as_ptr(), wallet.len()) };
+        assert_eq!(code, 0);
+
+        db
+    }
```
p0mvn: Would it make sense to define a test file and consolidate this helper in a single spot?

p0mvn (commented):
Consider verifying that the happy path test cases are present and consolidating test helpers, if possible.

No blocking comments overall

rust/src/voting/recovery.rs
czarcas7ic: yes, discussed

rust/src/voting/recovery.rs
```diff
@@ -0,0 +1,549 @@
...
+                &round_id_str,
+                bundle_index,
+                proposal_id,
+                &json_str,
+                vc_tree_position,
+            )
+            .map_err(|e| anyhow!("store_commitment_bundle failed: {}", e))?;
+        let persisted = handle
+            .db
+            .get_commitment_bundle(&round_id_str, bundle_index, proposal_id)
```
nullcopy: IMO, each of these readback & verify steps are unnecessary. This is defending against a possible bug in `VotingDb`. I'm not sure it makes sense to pay the performance penalty here, when that should be covered by tests in `VotingDb`

Unless, ofc, I am missing some reason why this is a critical bug to defend against. And if so... then probably also need to add the defense to `zcashlc_voting_store_keystone_signature`

rust/src/voting/share_tracking.rs
```diff
@@ -43,3 +80,230 @@ pub unsafe extern "C" fn zcashlc_voting_compute_share_nullifier(
...
+        let urls_bytes = unsafe { bytes_from_ptr(sent_to_urls_json, sent_to_urls_json_len) }?;
+        let sent_to_urls: Vec<String> = serde_json::from_slice(urls_bytes)?;
+        if nullifier_len != SHARE_NULLIFIER_LEN {
+            return Err(anyhow!(
+                "nullifier must be {} bytes, got {}",
+                SHARE_NULLIFIER_LEN,
+                nullifier_len
+            ));
+        }
+        let nullifier = unsafe { bytes_from_ptr(nullifier_ptr, nullifier_len) }?;
```
nullcopy: Here, the nullifier is encoded as a byte array, but in `zcashlc_voting_compute_share_nullifier` nullifiers are encoded as C strings. We should be consistent in the representations

rust/src/voting/test_helpers.rs
```diff
@@ -0,0 +1,33 @@
...
+        ea_pk: vec![7u8; 32],
+        nc_root: vec![8u8; 32],
+        nullifier_imt_root: vec![9u8; 32],
+    };
+    handle.db.init_round(&params, None).expect("insert round");
+
+    let wallet_id = handle.db.wallet_id();
+    let conn = handle.db.conn();
+    conn.execute(
+        "INSERT INTO bundles (round_id, wallet_id, bundle_index) VALUES (?1, ?2, ?3)",
```
nullcopy: The FFI shouldn't have to know about the DB schema/internals. Additionally, this will break silently if the schema changes without a corresponding update here in the FFI. Please expose a test fixture for `VotingDb` to provide whatever access you need here instead

rust/src/voting/recovery.rs
```diff
@@ -0,0 +1,549 @@
...
+}
+
+/// Remove all recovery-state rows for a round.
+///
+/// # Safety
+///
+/// - `db` must be a valid, non-null `VotingDatabaseHandle` pointer.
+/// - `round_id` must be a valid UTF-8 pointer with its stated length.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_voting_clear_recovery_state(
```
nullcopy, line 323: Missing test coverage

nullcopy (changes requested):
Changes requested as of fa08a0d375cf01a7cea0cc3d6bdb193edd829cef

rust/src/voting/recovery.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1722/commits/d84e760300def1a11850a8c0484b72c2645a6f7d

No, your point makes sense, this was done at the wrong layer of abstraction, removed them here.

rust/src/voting/share_tracking.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1722/commits/b766ff96f0d95f8552e3caa4c2a6b256a2cb5b6b

Agreed on consistency here. I updated so share nullifiers are hex strings going in and coming back out.

rust/src/voting/test_helpers.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1722/commits/00d038c2af406c225646184238c6413174cf0ce5

Bumped zcash voting with the test fixure and used here

rust/src/voting/recovery.rs
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1722/commits/d4a7f72604afca407a303d6b700dadc29b5b1cc7

Added coverage for the missing paths: delegation tx hash round-trip, vote tx hash round-trip, commitment bundle round-trip, and `clear_recovery_state` clearing the stored recovery data.

nullcopy (approved):
utACK 295b02472e164b7f4517968070ee33025a55d725

PR #1723, opened 2026-05-08
[#1661] Add voting delegation workflow FFI

czarcas7ic:
Adds the Rust libzcashlc FFI surface for the delegation workflow: hotkey generation, bundle setup/count, PCZT construction, tree-state persistence, proof generation with progress callbacks, delegation submission payloads, Keystone signature submission, and VAN position persistence.

Validation:
- git diff --check refs/prsplit/1722...HEAD
- cargo check --locked
- cargo clippy --all-targets -- -D warnings
- cargo test --locked voting::delegation


Cargo.toml
```diff
@@ -83,7 +83,7 @@ rust_decimal = { version = "1", default-features = false, features = ["c-repr"]
 xz2 = { version = "0.1", features = ["static"] }
 
 # Voting
-zcash_voting = { version = "0.5.3", default-features = false, features = [
+zcash_voting = { version = "0.5.5", default-features = false, features = [
```
czarcas7ic: Note to reviewer:

This bump includes the non zero account index delegation signing fix from 0.5.4, plus the Keystone sighash validation fix from 0.5.5, so Keystone submissions now have the same stored PCZT sighash check as the regular flow.

p0mvn (dismissed):

rust/src/voting/progress.rs
```diff
@@ -0,0 +1,22 @@
...
+    unsafe extern "C" fn(progress: f64, context: *mut std::ffi::c_void);
+
+/// Bridges a C function pointer to the `ProofProgressReporter` trait.
+pub(super) struct ProgressBridge {
+    pub(super) callback: VotingProgressCallback,
+    pub(super) context: *mut std::ffi::c_void,
+}
+
+// SAFETY: The caller guarantees the context pointer is valid for the duration
+// of the proof operation and that the callback is thread-safe.
```
nullcopy, line 16: The public doc on zcashlc_voting_build_and_prove_delegation only says the callback "must be a valid function pointer, or null to skip progress." Please move these threading/lifetime/reentrancy expectations into the # Safety section of the public function so the iOS side knows it must not capture progress_context past the await, and that the callback may fire from a Rayon/proving thread.

rust/src/voting/helpers.rs
```diff
@@ -136,6 +136,14 @@ pub(super) fn usk_from_seed(
     seed: &[u8],
     account: AccountId,
 ) -> anyhow::Result<UnifiedSpendingKey> {
+    if seed.len() < MIN_SEED_LEN {
```
nullcopy, line 139: Didn't this come in another PR already?

nullcopy (changes requested):
@czarcas7ic did you mean to target this PR to main?

czarcas7ic:
@nullcopy we have been stacking PRs to prevent conflicts as much as possible. I would review https://github.com/zcash/zcash-swift-wallet-sdk/pull/1722 first (I am doing a quick conflict res now) and then after that, this one.

rust/src/voting/helpers.rs
czarcas7ic: I think this was from the conflict after merge

rust/src/voting/progress.rs
czarcas7ic: ACK, moved them here https://github.com/zcash/zcash-swift-wallet-sdk/pull/1723/commits/c2f0858242b9520f717fdcfad720f94594ea4ded

rust/src/voting/delegation.rs
```diff
@@ -13,8 +13,17 @@ use zcash_voting::{self as voting, zkp1};
...
+use super::helpers::{
+    bytes_from_ptr, json_to_boxed_slice, open_wallet_db, str_from_ptr, voting_hotkey_to_ffi,
+};
+use super::json::{
+    JsonDelegationPirPrecomputeResult, JsonDelegationProofResult, JsonDelegationSubmission,
+    JsonNoteInfo, JsonVotingPczt, JsonWitnessData,
+};
+use super::progress::ProgressBridge;
+
+const SEED_FINGERPRINT_LEN: usize = 32;
```
nullcopy: We should collect these globals somewhere they can be used across the FFI. I remember other places where seed_fps where length checked against magic numbers. We should make a pass over the FFI and remove those occurance

nullcopy: `zcashlc_voting_generate_delegation_inputs_with_fvk` is an example of a function that could use this const

rust/src/voting/progress.rs
```diff
@@ -0,0 +1,22 @@
...
+unsafe impl Send for ProgressBridge {}
+unsafe impl Sync for ProgressBridge {}
+
+impl voting::ProofProgressReporter for ProgressBridge {
+    fn on_progress(&self, progress: f64) {
+        unsafe { (self.callback)(progress, self.context) }
```
nullcopy: Need `catch_unwind` here?

rust/src/voting/delegation.rs
```diff
@@ -49,6 +58,212 @@ fn validate_cached_tree_state_for_round(
...
+/// Generate a voting hotkey for a round.
+///
+/// Returns a pointer to `FfiVotingHotkey` on success, or null on error.
+/// Call `zcashlc_voting_free_hotkey` to free the returned pointer.
+///
+/// # Safety
+///
+/// - `db` must be a valid, non-null `VotingDatabaseHandle` pointer.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_voting_generate_hotkey(
```
nullcopy, line 69: Just noticing the corresponding `zcashlc_voting_free_hotkey` method doesn't zeroize the data... not strictly related to this PR, but want to make sure that gets fixed

nullcopy (changes requested):
Reviewed at fd990e26c3a9ac9984a96b09ad25093d265dba77

rust/src/voting/delegation.rs
p0mvn: [7ac1420](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1723/commits/7ac1420befa71f03b087805bbe951693f012356a)

rust/src/voting/progress.rs
p0mvn: [277cf53](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1723/commits/277cf5350af057b57d8291805a8f0df27b357081)

rust/src/voting/delegation.rs
p0mvn: [33a53c3](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1723/commits/33a53c35d1948f2ebbc9d231120052969c7360b5)

rust/src/voting/constants.rs
```diff
@@ -0,0 +1,2 @@
+/// Length of a Pallas seed fingerprint in bytes.
+pub(super) const SEED_FINGERPRINT_LEN: usize = 32;
```
nullcopy, line 5: Well now, in light of the `VotingByteCount` enum which came in another recent PR (line 11 in Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift), it looks like global constants like this are being managed in different ways. I don't care too much which way, but I think it should be consistent.

nullcopy: e.g. this could be removed in favor of `VotingByteCount.minSeed`, or vice versa.

czarcas7ic: I think there are other occurrences like this that slipped through from all the splits we did, which is why we will need to do another round of clean up after the full feature is in. That being said, I will address this one now

czarcas7ic: Okay, we communicated and the approach will be to remove the enum approach done in 1724 (the next PR in the series) in favor of the dedicated constants.

nullcopy (approved):
utACK e2ab3d27b60199432dcc7415badeefc9500bd5f9

PR #1728, opened 2026-05-08
[#1661] Add Swift voting cast-vote wrappers

p0mvn:
## Summary
- Expose Swift wrappers for vote witness generation and cast-vote FFI helpers.
- Add offline coverage for handle gating, Rust error propagation, payload construction, signing validation, and vote submission marking.
- Update the unreleased changelog entry for the new voting APIs.

## Note on Testing

Swift-side tests don’t exercise a full successful `buildVoteCommitment` or `generateNoteWitnesses` path with realistic fixture data. That's by choice as those are already covered more deeply in Rust.

## Test plan
- `swift test --filter VotingRustBackendTests`

### DIFF
--- a/Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
+++ b/Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
@@ -344,8 +344,11 @@ extension VotingRustBackend {
     /// Build a vote commitment proof for a proposal.
     ///
     /// The proof callback may be invoked from Rust worker threads. Do not call
-    /// back into this backend from `progress`, because the database handle lock
-    /// is held while the FFI call is active.
+    /// back into this backend from `progress`: the database handle lock is held
+    /// while the FFI call is active, so re-entering the backend will deadlock.
+    ///
+    /// Safety: keep `progress` thread-safe, non-blocking, and limited to
+    /// reporting state outside this backend.
     // swiftlint:disable:next function_parameter_count
     public func buildVoteCommitment(
         roundId: String,
--- a/Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
+++ b/Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
@@ -262,6 +262,232 @@ extension VotingRustBackend {
     }
 }
 
+// MARK: - Delegation witnesses
+
+extension VotingRustBackend {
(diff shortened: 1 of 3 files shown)


Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
```diff
@@ -262,6 +262,232 @@ extension VotingRustBackend {
...
+        }
+        defer { zcashlc_free_boxed_slice(ptr) }
+        return try decodeJSON(from: ptr)
+    }
+
+    /// Build a vote commitment proof for a proposal.
+    ///
+    /// The proof callback may be invoked from Rust worker threads. Do not call
+    /// back into this backend from `progress`, because the database handle lock
+    /// is held while the FFI call is active.
```
nullcopy: I would strengthen this warning with "calling back into this backend from progress will deadlock"

nullcopy: Claude also suggested using `NSRecursiveLock` instead of `NSLock`, to avoid this problem altogether. I'm not a swift dev, so I can't weigh in on the merits of either approach

Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
nullcopy:
```suggestion
    ///  elsewhere.
```

nullcopy (changes requested):
Reviewed 513e7aad13cfac9df64ce7f5627a740beb8ff6b1

Sources/ZcashLightClientKit/Rust/Voting/VotingRustBackend.swift
p0mvn: I'm reading that `NSRecursiveLock` only helps if the callback re-enters from the exact same thread, but Rust explicitly says progress may come from worker threads.

I've strengthened the warning: [982d2a9](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1728/commits/982d2a92fcc707ce076ac1dd0a89043c3c793ff1)

We've tested the current implementation and haven't observed issues if the contract is followed. I would like to propose keeping the amendment as a more explicit warning.

p0mvn:
Addressing, just noticed a CI failure from a Rust change in concurrent PR

nullcopy (approved):
utACK 8f64127ac0c0e297f8d1301f28ac07d924eca33b

PR #1737, opened 2026-05-12
[MOB-1039] Retry broadcaster transactions with submit plans

czarcas7ic:
## Summary
- Prevents the sync resubmission path from needing an ever-growing list of Broadcaster transactions to skip by recording the submit plan it should use for retry (cref https://github.com/zcash/zcash-android-wallet-sdk/pull/1937#discussion_r3222459596)
- Adds pending submit plan storage for transactions created through `Broadcaster`.
- Records the endpoints used by `Broadcaster.submit` and retries pending transactions with those endpoints instead of the synchronizer default endpoint.
- Keeps legacy synchronizer submit paths on the existing SDK owned retry behavior.

## Testing
- `swift test --filter OfflineTests`
- Targeted `swiftlint lint` on touched files, with only the existing `SDKSynchronizer.swift:268` force unwrap warning reported.

### DIFF
--- a/Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
+++ b/Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
@@ -33,6 +33,8 @@ actor PendingSubmitPlanStore {
     private let logger: Logger
 
     private var plansByTransactionId: [String: [StoredEndpoint]] = [:]
+    // In-memory cache only. After restart, raw transaction submissions recover
+    // the transaction id through RawTransactionLookup before recording endpoints.
     private var transactionIdsByRawTransaction: [String: String] = [:]
     private var loadedFromPersistence = false
 
@@ -150,6 +152,9 @@ actor PendingSubmitPlanStore {
             }
 
             let storedPlans = try JSONDecoder().decode(StoredPlans.self, from: data)
+            guard storedPlans.version == StoredPlans.currentVersion else {
+                throw PendingSubmitPlanStoreError.unsupportedVersion(storedPlans.version)
+            }
             plansByTransactionId = storedPlans.plansByTransactionId
         } catch {
             logger.warn("Failed to load pending submit plans: \(error)")
@@ -171,13 +176,19 @@ private struct StoredPlans: Codable {
     let version: Int
     let plansByTransactionId: [String: [StoredEndpoint]]
 
+    static let currentVersion = 1
+
     init(
-        version: Int = 1,
+        version: Int = Self.currentVersion,
         plansByTransactionId: [String: [StoredEndpoint]]
     ) {
         self.version = version
         self.plansByTransactionId = plansByTransactionId
     }
+}
+
+private enum PendingSubmitPlanStoreError: Error {
+    case unsupportedVersion(Int)
 }
 
 private struct StoredEndpoint: Codable, Equatable {
--- a/Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
+++ b/Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
@@ -105,10 +105,11 @@ actor PendingSubmitPlanStore {
         loadFromPersistenceIfNeeded()
 
         let retainedTransactionIds = Set(transactionIds.map(\.stablePlanKey))
-        let previousCount = plansByTransactionId.count
+        let previousPlanCount = plansByTransactionId.count
         plansByTransactionId = plansByTransactionId.filter { retainedTransactionIds.contains($0.key) }
+        transactionIdsByRawTransaction = transactionIdsByRawTransaction.filter { retainedTransactionIds.contains($0.value) }
 
-        if plansByTransactionId.count != previousCount {
+        if plansByTransactionId.count != previousPlanCount {
             saveToPersistence()
         }
     }
--- a/Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
+++ b/Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
@@ -169,13 +169,13 @@ actor PendingSubmitPlanStore {
 
     private func loadFromPersistenceIfNeeded() {
         guard !loadedFromPersistence else { return }
-        defer { loadedFromPersistence = true }
 
         do {
             guard
                 let data = try persistence?.load(),
                 !data.isEmpty
             else {
+                loadedFromPersistence = true
                 return
             }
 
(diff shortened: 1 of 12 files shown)


Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
```diff
@@ -0,0 +1,367 @@
...
+        transactionIdsByRawTransaction = transactionIdsByRawTransaction.filter { retainedTransactionIds.contains($0.value) }
+
+        if plansByTransactionId.count != previousPlanCount {
+            saveToPersistence()
+        }
+    }
+
+    private func loadFromPersistenceIfNeeded() {
+        guard !loadedFromPersistence else { return }
+        defer { loadedFromPersistence = true }
```
Chlup: No matter what happens this is set to `true`. So if loading of persistence fails then `plansByTransactionId` is empty and on next `saveToPersistence()` call all the stored data are rewritten and lost.

Sources/ZcashLightClientKit/Transaction/SubmitPlanExecutor.swift
```diff
@@ -0,0 +1,114 @@
...
+        submitPlan: TransactionSubmitPlan
+    ) async throws {
+        var lastError: Error?
+
+        for endpoint in submitPlan.endpoints {
+            do {
+                try await transactionSubmitter.submit(transaction: transaction, to: endpoint)
+                return
+            } catch {
+                lastError = error
```
Chlup, line 114: It would be useful to at least lot these errors. Otherwise all the errors (except the last one) are silently lost.

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
```diff
@@ -0,0 +1,367 @@
...
+
+actor PendingSubmitPlanStore {
+    enum StoredSubmitPlan {
+        case awaitingPlan
+        case ready(TransactionSubmitPlan)
+    }
+
+    private let persistence: PendingSubmitPlanPersistence?
+    private let logger: Logger
+    private let lock = AsyncLock()
```
Chlup: Can you please explain intention behind this solution? I see that you use actor for `PendingSubmitPlanStore` and then put lock into it. Which is bit unusual.

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
```diff
@@ -0,0 +1,367 @@
...
+struct TransactionSubmitPlan {
+    let endpoints: [LightWalletEndpoint]
+
+    init(endpoints: [LightWalletEndpoint]) {
+        precondition(!endpoints.isEmpty, "Transaction submit plan must include at least one endpoint.")
+        self.endpoints = endpoints
+    }
+}
+
+actor PendingSubmitPlanStore {
```
Chlup, line 32: Can you please add some documentation to this? At least to explain what this object does in general.

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
```diff
@@ -0,0 +1,367 @@
...
+            streamingCallTimeoutInMillis: streamingCallTimeoutInMillis
+        )
+    }
+}
+
+private extension Data {
+    var stablePlanKey: String { hexEncodedString() }
+}
+
+private final class AsyncLock {
```
Chlup: This works in happy paths but it has also issues. I used similar solution in past but it is not ideal. 

For example it doesn't support cancelation right. If the task awaiting acquire() is cancelled while suspended in the continuation, the continuation is leaked forever. It never resumes, which means:

- The cancelled task hangs indefinitely on await state.acquire().
- The waiter stays in the waiters array.
- When the current holder calls release(), it resumes a continuation belonging to a cancelled task — which now proceeds as if it acquired the lock, but the original caller is gone, so release() may never be called. The lock is permanently leaked.

I think we should get rid of `AsyncLock` completely and find solution without it because it is hidden ticking bomb.

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1737/commits/bd3a313d2fdefd075de85b2043df7dd194034e19

Fixed here and added a regression test

Sources/ZcashLightClientKit/Transaction/SubmitPlanExecutor.swift
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1737/commits/e749fec0317c03b1da5bc5f98435470fc9f82dcf

Log added

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
czarcas7ic: I removed AsyncLock here: e236ddf 02ff1de

The store now just uses the actor for isolation. When candidate lookup overlaps with transaction creation or endpoint updates, it skips pruning for that pass and lets the next clean pass prune stale plans. I added tests for those cases too.

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
czarcas7ic: The lock was trying to cover the case where actor methods interleave at an `await`, especially around pruning plans from a transaction candidate lookup.

I removed it here: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1737/commits/e236ddf32c4fb02d3edf6fa7d99e091f9df86d7f

Now, if that lookup overlaps with submit plan creation or endpoint updates, we just skip pruning for that pass.

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
czarcas7ic: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1737/commits/fbd26aefb918464bf20d21b3612c9e0e51a723a1

Chlup: 👍

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
Chlup: 👍

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
Chlup: 👍

Sources/ZcashLightClientKit/Transaction/PendingSubmitPlanStore.swift
Chlup: 👍

Sources/ZcashLightClientKit/Transaction/SubmitPlanExecutor.swift
Chlup: 👍

Chlup:
This was finally implemented in #1757

PR #1738, opened 2026-05-13
Bump zcash_voting to 0.8.1

czarcas7ic:
Updates the Rust voting dependency to `zcash_voting 0.8.1`, refreshes `Cargo.lock`, and records the bump in both changelogs.

This carries the existing voting SDK updates forward and adds the 0.8.x voting DB behavior. Pre-launch voting SQLite state resets to the current schema, recovery store operations now fail from the crate when the target bundle or vote row is missing, and the SDK FFI no longer does its own delegation hash existence read before storing.

Validation:
- `cargo fmt`
- `git diff --check`
- `cargo tree -p libzcashlc -i zcash_voting --locked --edges normal -e features`
- `cargo test -p libzcashlc voting:: --lib --locked`
- `cargo test -p libzcashlc --lib --locked`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # Unreleased
 
 ## Changed
-- Bumped `zcash_voting` to `0.7.0` for required voting input validation and paginated vote commitment tree sync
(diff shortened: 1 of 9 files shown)


nuttycom (approved):
utACK. The `zcash_voting` dependency bump doesn't really need to be noted in the CHANGELOG, but that can be cleaned up later; we'll ultimately want the changelog to describe the added functionality since 2.5.0, not these interim changes.

PR #1744, opened 2026-05-19
Use released zcash_voting 0.10.1

p0mvn:
## Summary
- Use the published `zcash_voting 0.10.1` crate instead of the temporary git revision.
- Refresh `Cargo.lock` so `zcash_voting`, `vote-commitment-tree`, `vote-commitment-tree-client`, and `voting-circuits` resolve from crates.io.
- Update the Rust changelog dependency note to match the released crate version.

## Test plan
- `cargo metadata --locked --format-version 1 --no-deps`
- `cargo tree --locked -i zcash_voting`
- `cargo tree --locked -p zcash_voting`
- `cargo check --locked`
- `cargo test --locked voting:: --lib`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,7 +84,7 @@ rust_decimal = { version = "1", default-features = false, features = ["c-repr"] 
 xz2 = { version = "0.1", features = ["static"] }
 
 # Voting
-zcash_voting = { version = "0.10.0", default-features = false, features = [
+zcash_voting = { version = "=0.10.1", default-features = false, features = [
     "client-pir",
     "client-tree-sync",
 ] }
@@ -92,7 +92,7 @@ zcash_keys = { version = "0.13", features = ["orchard"] }
 incrementalmerkletree = { version = "0.8", default-features = false }
 
 [dev-dependencies]
-zcash_voting = { version = "0.10.0", default-features = false, features = ["test-fixtures"] }
+zcash_voting = { version = "0.10.1", default-features = false, features = ["test-fixtures"] }
 
 [build-dependencies]
 bindgen = "0.72"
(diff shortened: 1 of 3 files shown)


czarcas7ic (dismissed):

greg0x (dismissed):

greg0x (dismissed):

greg0x (dismissed):

Cargo.toml
```diff
@@ -84,15 +84,15 @@ rust_decimal = { version = "1", default-features = false, features = ["c-repr"]
...
+zcash_voting = { version = "=0.10.1", default-features = false, features = [
     "client-pir",
     "client-tree-sync",
 ] }
 zcash_keys = { version = "0.13", features = ["orchard"] }
 incrementalmerkletree = { version = "0.8", default-features = false }
 
 [dev-dependencies]
-zcash_voting = { version = "0.8.1", default-features = false, features = ["test-fixtures"] }
+zcash_voting = { version = "0.10.1", default-features = false, features = ["test-fixtures"] }
```
nullcopy, line 95: Dev dependency should also be pinned, right?

nullcopy (changes requested):

Cargo.toml
greg0x: When we did the pinning we have left the dev dependencies without pinning. 

Given that we have the `Cargo.lock` in the repo the dependencies won't be bumped without explicit user action.

Anyhow I'm pinning in a new commit and worst case we revert it.

greg0x: https://github.com/zcash/zcash-swift-wallet-sdk/pull/1744/changes/ff2c7180cf084289c53671c9e39b876ac6c6a33c

nullcopy (approved):
utACK ff2c7180cf084289c53671c9e39b876ac6c6a33c

nullcopy (dismissed):
NACK. @nuttycom explained that we only pin with the lock file. `Cargo.toml` should not add pinned versions. Sorry for the confusion

nullcopy (approved):
utACK 0c26adefecece5401552ccce4e963d192e15002c

PR #1746, opened 2026-05-26
[#1745] Complete the FFI release-deployment workflow

### LINKED ISSUE #1745: Complete the FFI release-deployment CI workflow
The `Build FFI XCFramework` workflow (`.github/workflows/build-ffi.yml`) is an incomplete release-deployment pipeline. Today it builds the XCFramework and uploads a draft release, then only *prints* the values a human must paste into `Package.swift`. It never verifies that the SDK compiles and passes offline tests against the newly built FFI, and it leaves `Package.swift`/`CHANGELOG.md` untouched.

## Goal

Extend the workflow (alpha/beta pre-releases) to:

- Build the full 5-architecture XCFramework.
- Verify the Swift SDK builds (`swift build`) and `OfflineTests` pass against the new FFI, using the `LocalPackages` override — **before** publishing the release artifact.
- Upload the artifact as a draft GitHub **pre-release**.
- Update `Package.swift` (binaryTarget URL + checksum) and stamp `CHANGELOG.md` (`# Unreleased` -> `# <version> - <date>`).
- Commit those changes to a `release/ffi-<version>` branch and open a PR referencing this issue.

CI stops at opening the PR: signing the release tag requires the maintainer's GPG key. A new `Scripts/finalize-release.sh` lets a maintainer, after the PR merges, create the signed tag on the merge commit and publish the draft pre-release.

Shared logic is factored into small scripts (`update-package-swift.sh`, `stamp-changelog.sh`, `upload-ffi-draft.sh`, `finalize-release.sh`) so the workflow and the existing local `prepare-release.sh`/`release.sh` reuse it without duplication.

---
*Filed with [Claude Code](https://claude.com/claude-code) assistance.*

nuttycom:
Completes the FFI release-deployment CI workflow per the design discussed under #1745.

## What changes

`.github/workflows/build-ffi.yml` now drives the full release-prep pipeline for alpha/beta pre-releases:

1. Build the 5-architecture XCFramework.
2. Verify it: `swift build` + `OfflineTests` against the new FFI via the `LocalPackages/` override (same mechanism `swift.yml` uses).
3. Upload it as a draft pre-release (only reached if verification passed).
4. Update `Package.swift` and stamp `CHANGELOG.md`.
5. Commit on `release/ffi-<version>` and open a PR.

After the PR merges, a maintainer runs `./Scripts/finalize-release.sh <remote> <version>` from the **tip of the `release/ffi-<version>` branch** — the new script signs the tag (on the commit CI actually produced and verified) and publishes the draft.

Shared logic is factored into small focused scripts so the CI workflow and the local `release.sh`/`prepare-release.sh` path reuse it instead of duplicating it:

- `Scripts/update-package-swift.sh` — rewrites the `binaryTarget` URL + checksum.
- `Scripts/stamp-changelog.sh` — converts `# Unreleased` → `# <version> - <date>`.
- `Scripts/upload-ffi-draft.sh` — archives + uploads the draft (with `--prerelease` for alpha/beta).
- `Scripts/finalize-release.sh` — maintainer post-merge signed-tag + publish.

`Scripts/prepare-release.sh` and `Scripts/release.sh` are refactored to delegate to these helpers; `release.sh` now produces the same commit shape as CI and delegates the tag/publish tail to `finalize-release.sh`.

## Key design decisions

- **CI stops at opening the PR.** Signing the tag needs the maintainer's GPG key — keeping it out of CI avoids storing a signing key as a workflow secret.
- **Tag the release branch tip, not the merge commit.** `finalize-release.sh` enforces this with a pre-flight that compares `HEAD` to `refs/heads/release/ffi-<version>`. A merge or squash commit carries base-branch state that wasn't in the verified release artifact, so the tag must point to the commit CI produced.
- **Version validation in two places.** The workflow rejects anything other than `x.y.z-alpha.n` / `x.y.z-beta.n`; `finalize-release.sh` re-enforces it before tagging. These pre-releases produce artifacts for internal test builds of the wallet, so the release PR is not required to reference a tracking issue.
- **`upload-ffi-draft.sh` writes `release.env` only after a successful upload**, so a failed upload can't leave behind metadata claiming success.
- **`finalize-release.sh` distinguishes local-only from fully-pushed tags**, so a partial-failure recovery (where `git push` failed after `git tag -s` succeeded) gets actionable guidance instead of a dead-end.

## Verification

- Each script passes `bash -n`. Argument-validation and pre-flight error paths were exercised on Linux against the actual scripts (version-regex rejection, missing-XCFramework guard, missing-`# Unreleased` guard, etc.).
- The workflow YAML parses cleanly. zizmor-relevant controls: top-level `permissions: {}`; job-scoped `contents: write` + `pull-requests: write`; `persist-credentials: false` on checkout with explicit `x-access-token` push URL; all `uses:` pinned to commit SHA; every `${{ }}` interpolation confined to `env:`.
- `CHANGELOG.md` is not stamped in this PR — it'll be stamped by the first release that uses the new workflow.

## Caveat

The release PR opened by CI is authored by `GITHUB_TOKEN`, so `swift.yml` PR checks do not run on it (GitHub blocks recursive workflow triggers). Verification was already performed in the release-workflow run itself; the caveat is stated in the PR body the workflow generates.

Closes #1745

---
*Filed with [Claude Code](https://claude.com/claude-code) assistance.*

### DIFF
--- a/.github/workflows/build-ffi.yml
+++ b/.github/workflows/build-ffi.yml
@@ -1,22 +1,77 @@
 name: Build FFI XCFramework
 
 on:
-  # Manual trigger for preparing releases.
-  # The branch selected in the GitHub UI determines which commit is checked out and built.
+  # Manual trigger for preparing alpha/beta pre-releases.
+  # The branch selected in the GitHub UI determines which commit is built and
+  # is used as the base branch for the release PR.
+  #
+  # These pre-releases produce artifacts for internal-only test builds of the
+  # wallet, so the release PR does not need to reference a tracking issue.
   workflow_dispatch:
     inputs:
       version:
-        description: 'SDK version (e.g., 2.5.0)'
+        description: 'Pre-release version (x.y.z-alpha.n or x.y.z-beta.n, e.g. 2.6.0-alpha.2)'
         required: true
 
 permissions: {}
 
+# Two concurrent dispatches for the same version would race on draft
+# creation and the release-branch push. group-by-version cancels the
+# older run if a newer dispatch comes in for the same version.
+concurrency:
+  group: build-ffi-${{ inputs.version }}
+  cancel-in-progress: true
+
 jobs:
-  build-xcframework:
+  build-test-release:
     runs-on: macos-15
     permissions:
-      contents: write # Required for creating draft releases
+      contents: write       # create the draft release and push the release branch
+      pull-requests: write  # open the release PR
+    env:
+      VERSION: ${{ github.event.inputs.version }}
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      REPO: ${{ github.repository }}
+      BASE_BRANCH: ${{ github.ref_name }}
     steps:
+      - name: Validate inputs
+        env:
+          REF_TYPE: ${{ github.ref_type }}
+          REF_NAME: ${{ github.ref_name }}
+          APP_ID: ${{ vars.RELEASE_BOT_APP_ID }}
+        run: |
+          if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+-(alpha|beta)\.[0-9]+$ ]]; then
+            echo "::error::version '$VERSION' must match x.y.z-alpha.n or x.y.z-beta.n"
+            exit 1
+          fi
+          # The release PR needs a branch as its base. Refuse to run if the
+          # workflow was dispatched against a tag (possible via the API even
+          # though the UI dropdown only shows branches) — the 60-min build
+          # would fail at gh pr create later.
+          if [[ "$REF_TYPE" != "branch" ]]; then
+            echo "::error::workflow must be dispatched from a branch (got ${REF_TYPE} '${REF_NAME}')"
+            exit 1
+          fi
+          # The release-bot GitHub App is required: app-installation tokens
+          # trigger downstream PR-check workflows (swift.yml, zizmor) on the
+          # release PR, which GITHUB_TOKEN cannot. Refuse to run without it
+          # rather than producing a PR that can't pass required checks.
+          if [[ -z "$APP_ID" ]]; then
+            echo "::error::repo variable RELEASE_BOT_APP_ID is not set; configure the release-bot GitHub App"
+            exit 1
+          fi
+
+      # Mint an installation token for the release-bot GitHub App. App
+      # tokens (unlike GITHUB_TOKEN) trigger downstream workflows, so the
+      # release PR's required status checks (swift.yml, zizmor) fire
+      # normally. Required: enforced above by the Validate inputs step.
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ vars.RELEASE_BOT_APP_ID }}
+          private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
+
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
@@ -33,21 +88,120 @@ jobs:
       - name: Cache Cargo
         uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
 
-      - name: Build and upload draft release
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          VERSION: ${{ github.event.inputs.version }}
-        run: ./Scripts/prepare-release.sh "$VERSION"
+      # Build the full 5-architecture XCFramework — the release artifact.
+      # Past runs of the old workflow took ~42-44 min for the same build;
+      # 60 gives a safety margin without being absurd.
+      - name: Build XCFramework (all architectures)
+        timeout-minutes: 60
+        working-directory: ./BuildSupport
+        run: |
+          make clean
+          make xcframework
 
-      - name: Output Package.swift Update
+      # Point Package.swift at the freshly built FFI via the LocalPackages
+      # override (same mechanism as swift.yml). LocalPackages/ is gitignored.
+      - name: Configure local FFI for verification
+        run: |
+          mkdir -p LocalPackages
+          cp -R BuildSupport/products/libzcashlc.xcframework LocalPackages/
+          cp BuildSupport/LocalPackages-Package.swift LocalPackages/Package.swift
+
+      - name: Build Swift package against new FFI
+        timeout-minutes: 15
+        run: swift build -v
+
+      - name: Run OfflineTests against new FFI
+        timeout-minutes: 10
+        run: swift test --filter OfflineTests
+
+      # Verification passed — remove the override so it is never committed.
+      - name: Remove local FFI override
+        run: rm -rf LocalPackages
+
+      # Only reached if build + tests succeeded.
+      # CI passes --force-overwrite-existing-release so a partial-failure
+      # retry can complete: upload-ffi-draft.sh's existing-draft path will
+      # update the assets in place. The script refuses to clobber a release
+      # that has been promoted out of draft state, so this stays safe.
+      - name: Upload draft pre-release
+        run: ./Scripts/upload-ffi-draft.sh --force-overwrite-existing-release "$VERSION"
+
+      - name: Update Package.swift and CHANGELOG
         run: |
           source BuildSupport/products/release.env
-          echo ""
-          echo "========================================"
-          echo "Update Package.swift with:"
-          echo "========================================"
-          echo ""
-          echo "url: \"${DOWNLOAD_URL}\","
-          echo "checksum: \"${CHECKSUM}\""
-          echo ""
-          echo "Then commit, tag ${VERSION}, and publish the draft release."
+          ./Scripts/update-package-swift.sh "$VERSION" "$CHECKSUM"
+          ./Scripts/stamp-changelog.sh "$VERSION"
+
+      - name: Open release PR
+        env:
+          APP_SLUG: ${{ steps.app-token.outputs.app-slug }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          BRANCH="release/ffi-${VERSION}"
+          # Installation tokens aren't user-bound, so /user is unreliable.
+          # Derive the bot identity from the app slug; the /users/<slug>[bot]
+          # endpoint is public and resolves the bot's numeric id.
+          BOT_USER="${APP_SLUG}[bot]"
+          BOT_ID="$(gh api "/users/${BOT_USER}" --jq '.id')"
+          git config user.name "$BOT_USER"
+          git config user.email "${BOT_ID}+${BOT_USER}@users.noreply.github.com"
+          # -B (force) so a retry of this workflow re-uses the branch
+          # instead of aborting on "branch already exists" against the
+          # runner's prior local state.
+          git checkout -B "$BRANCH"
+          git add Package.swift CHANGELOG.md
+          git commit -m "Prepare release ${VERSION}"
+          # Capture the release commit SHA so the PR body and finalize
+          # invocation pin to this exact commit, not the (mutable) branch.
+          RELEASE_SHA=$(git rev-parse HEAD)
+          # --force-with-lease lets a retry replace its own prior release
+          # branch on the remote without overwriting unrelated work that
+          # may have happened to land there in between runs.
+          git push --force-with-lease="${BRANCH}": \
+            "https://x-access-token:${GH_TOKEN}@github.com/${REPO}.git" "HEAD:${BRANCH}"
+          mkdir -p tmp
+          cat > tmp/pr-body.md <<BODY
+          Updates \`Package.swift\` to the \`${VERSION}\` FFI draft pre-release and stamps the CHANGELOG.
+
+          Built and verified by the \`Build FFI XCFramework\` workflow run: \`swift build\` and the \`OfflineTests\` suite pass against the \`${VERSION}\` XCFramework.
+
+          After this PR is approved, a maintainer signs the tag and publishes the draft pre-release **before merging** the PR (the repo auto-deletes branches on merge, so finalize must run while the branch still exists on the remote). The tag must point to the commit CI produced — \`${RELEASE_SHA}\` — passed explicitly to finalize so the script can verify it.
+
+          \`\`\`
+          git fetch <remote> ${RELEASE_SHA}
+          git checkout ${RELEASE_SHA}
+          ./Scripts/finalize-release.sh <remote> ${VERSION} ${RELEASE_SHA}
+          \`\`\`
+
+          Then merge this PR.
+          BODY
+          gh pr create \
+            --repo "$REPO" \
+            --base "$BASE_BRANCH" \
+            --head "$BRANCH" \
+            --title "Prepare release ${VERSION}" \
+            --body-file tmp/pr-body.md
+          # Surface the release SHA to later steps (Summary).
+          echo "RELEASE_SHA=${RELEASE_SHA}" >> "$GITHUB_ENV"
+
+      - name: Summary
+        if: always()
+        run: |
+          {
+            echo "## Release ${VERSION}"
+            echo ""
+            echo "- Draft pre-release: https://github.com/${REPO}/releases/tag/${VERSION}"
+            if [[ -n "${RELEASE_SHA:-}" ]]; then
+              echo "- Release commit: \`${RELEASE_SHA}\` on branch \`release/ffi-${VERSION}\`"
+            fi
+            echo ""
+            echo "### Maintainer next steps"
+            echo "1. Review and approve the release PR (but **do not merge yet** — the release branch must still exist on the remote for step 2)."
+            echo "2. Finalize the release **before merging**:"
+            echo "   \`\`\`"
+            echo "   git fetch <remote> ${RELEASE_SHA:-<release-sha-see-PR-body>}"
+            echo "   git checkout ${RELEASE_SHA:-<release-sha-see-PR-body>}"
+            echo "   ./Scripts/finalize-release.sh <remote> ${VERSION} ${RELEASE_SHA:-<release-sha-see-PR-body>}"
+            echo "   \`\`\`"
+            echo "3. Merge the release PR. The branch auto-deletes; the signed tag preserves the release commit."
+          } >> "$GITHUB_STEP_SUMMARY"
--- a/Scripts/finalize-release.sh
+++ b/Scripts/finalize-release.sh
@@ -1,0 +1,199 @@
+#!/bin/bash
+# Tag and publish a pre-release whose FFI artifacts and Package.swift /
+# CHANGELOG update were already produced by the "Build FFI XCFramework"
+# CI workflow.
+#
+# Run this AFTER approving the CI-opened `release/ffi-<version>` PR but
+# BEFORE merging it: the release branch must still exist on the remote so
+# its tip can be fetched and tagged. The tag must point to the commit CI
+# produced — passed explicitly as <release-sha> so the script can verify
+# HEAD matches and refuse to tag a merge or post-merge commit.
+#
+# It:
+#   1. Verifies HEAD equals <release-sha> and that SHA is the tip of
+#      <remote>/release/ffi-<version>, and that the draft release exists.
+#   2. Creates a GPG-signed tag <version> on HEAD.
+#   3. Pushes the tag to <remote>.
+#   4. Publishes the draft pre-release.
+#
+# Usage: ./Scripts/finalize-release.sh <remote> <version> <release-sha>
+#
+# The release commit SHA is printed in the body of the CI-opened release
+# PR (and in the workflow run's Summary).
+#
+# Prerequisites:
+#   - gh CLI installed and authenticated
+#   - GPG signing key configured for the repo
+#   - The release/ffi-<version> PR approved (but not yet merged) and the
+#     release commit checked out, e.g.:
+#       git fetch <remote> <release-sha>
+#       git checkout <release-sha>
+
+set -e
+cd "$(dirname "$0")/.."
+
+REMOTE="$1"
+VERSION="$2"
+RELEASE_SHA="$3"
+REPO="zcash/zcash-swift-wallet-sdk"
+
+if [[ -z "$REMOTE" || -z "$VERSION" || -z "$RELEASE_SHA" ]]; then
+    echo "Usage: $0 <remote> <version> <release-sha>" >&2
+    echo "Example: $0 upstream 2.6.0-alpha.2 4838fa913fea1234..." >&2
+    exit 1
+fi
+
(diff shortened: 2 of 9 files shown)


.github/workflows/build-ffi.yml
```diff
@@ -33,21 +50,93 @@ jobs:
...
+        run: |
+          {
+            echo "## Release ${VERSION}"
+            echo ""
+            echo "- Draft pre-release: https://github.com/${REPO}/releases/tag/${VERSION}"
+            echo "- Release PR opened from branch \`release/ffi-${VERSION}\`"
+            echo ""
+            echo "### Maintainer next steps"
+            echo "1. Review and merge the release PR."
+            echo "2. Check out the **release branch tip** (not the merge commit) and run:"
```
nuttycom: Can the specific commit hash be referenced here? That would be simpler and more precise than depending on the branch name.

Chlup:
## Code review — local review of branch `kris/1745-finish-release-workflow`

**Edited:** removed one finding that doesn't hold up on re-examination, combined two overlapping items, and tightened severity claims that were overstated.

---

### Blockers (genuinely break the release flow)

**1. CI-opened PR can't pass required status checks** — [`.github/workflows/build-ffi.yml:91`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/.github/workflows/build-ffi.yml#L91)

`build` (swift.yml) and `Run zizmor 🌈` are required on `main`; both trigger only on `pull_request`. Workflows triggered by `GITHUB_TOKEN` don't fire downstream workflows (GitHub's documented anti-recursion behavior). The PR body itself acknowledges this at line 113 but offers no path forward.

Compounding: branch protection has `dismiss_stale_reviews_on_push: true`. So the workaround (push a commit from a user account to trigger the checks) also dismisses any approval already collected. The dance is: push commit → wait for checks → re-approve → merge.

_Verified via `gh api repos/.../rules/branches/main` — ruleset 15170778._

**2. `deleteBranchOnMerge: true` breaks the documented finalize flow** — [`Scripts/finalize-release.sh:84`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/finalize-release.sh#L84)

After merge, `release/ffi-<version>` is gone from the remote. The script requires `refs/heads/release/ffi-<version>` to exist locally. The PR body, step summary, and `CLAUDE.md` all tell the maintainer to fetch the branch AFTER merging — which fails with `couldn't find remote ref`.

Workaround exists: fetch the branch locally BEFORE clicking merge (the local copy survives the remote deletion). But the documented flow needs updating, or `finalize-release.sh` should accept a SHA fallback.

_Verified via `gh repo view --json deleteBranchOnMerge` → `true`._

---

### High severity (real retry/recovery gaps)

**3. No retry path on the CI workflow after a partial failure** — [`.github/workflows/build-ffi.yml:83`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/.github/workflows/build-ffi.yml#L83), [`:99`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/.github/workflows/build-ffi.yml#L99)

Two combined issues:

a) `./Scripts/upload-ffi-draft.sh "$VERSION"` is invoked without `--force-overwrite-existing-release`. Any failure between upload (line 82-83) and PR creation (line 91-123) leaves the draft release dangling; re-running aborts at [`upload-ffi-draft.sh:60`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/upload-ffi-draft.sh#L60).

b) The release branch push (line 99) is non-force. If a previous run got past the upload step but failed on PR creation, the remote branch exists and the second run's `git push HEAD:$BRANCH` is rejected non-fast-forward.

Either issue alone blocks retries; together there's no clean recovery path. Maintainer has to manually `gh release delete` AND delete the remote branch before retrying.

**4. `prepare-release.sh` does `git checkout -b` unconditionally; local `release.sh` retry fails** — [`Scripts/prepare-release.sh:64`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/prepare-release.sh#L64)

`release.sh:84` only checks for a pre-existing tag, not branch. Re-running after any failure aborts at line 64 with `branch already exists`. No documented recovery — maintainer must manually `git branch -D`.

**5. `finalize-release.sh` has no recovery for "tag pushed, release-edit failed"** — [`Scripts/finalize-release.sh:75`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/finalize-release.sh#L75)

Lines 67-78 handle the local-only-tag case but only that. If `gh release edit --draft=false` fails after `git push` succeeded (rare but possible), re-running lands at the `else` branch (line 75) and exits with no actionable hint. Maintainer ends up with a published tag pointing at a still-draft release.

---

### Medium severity (real but narrow triggers)

**6. `release.sh` cleanup on failure is incomplete** — [`Scripts/release.sh:107`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/release.sh#L107)

Error path is just `git checkout Package.swift`. The freshly-created release branch, the draft release on GitHub, and any CHANGELOG.md modification are left dangling. Trigger is narrow — most failures hit `stamp-changelog.sh:17-19` before any file mutation — but the cleanup is incomplete when it matters.

**7. `--force-overwrite-existing-release` doesn't verify the existing release is still a draft** — [`Scripts/upload-ffi-draft.sh:65`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/upload-ffi-draft.sh#L65)

The clobber path runs `gh release upload --clobber` without checking `.isDraft`. A user passing this flag has opted into destructive behavior, but adding an `isDraft` guard would prevent accidentally replacing published binaries.

**8. Belt-and-braces grep doesn't pin the commit SHA** — [`Scripts/finalize-release.sh:100`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/finalize-release.sh#L100)

`Package.swift` / `CHANGELOG.md` text checks pass on any commit with the right strings. The realistic risk is a maintainer accidentally amending the local branch — text content stays the same, the check still passes, the tag ends up on the wrong commit. Ergonomic gap, not a security gap (anyone with write to the release branch could do worse), but worth tightening.

**9. `BASE_BRANCH: github.ref_name` is unsafe when dispatched from a tag** — [`.github/workflows/build-ffi.yml:28`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/.github/workflows/build-ffi.yml#L28)

`workflow_dispatch` from the GitHub UI dropdown only shows branches, but API dispatch can target a tag. In that case the 45-min build runs, draft is created, then `gh pr create --base <tag>` fails. Narrow trigger.

---

### Lower severity (nice-to-haves, not bugs)

**10. `stamp-changelog.sh` doesn't reinsert `# Unreleased`** — [`Scripts/stamp-changelog.sh:23`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/stamp-changelog.sh#L23)

The `sed` consumes the `# Unreleased` header without reinserting it. This matches the existing convention in this repo (current CHANGELOG.md shows `# Unreleased` was manually re-added after 2.5.1), and the old release scripts didn't reinsert it either — so not a regression. Nice-to-have automation.

**11. No `concurrency:` block on the workflow** — [`.github/workflows/build-ffi.yml`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/.github/workflows/build-ffi.yml)

Two simultaneous dispatches with the same `VERSION` race on draft creation and branch push. Unlikely trigger; worth adding `concurrency: group: build-ffi-${{ inputs.version }}` for hygiene.

**12. `release.sh` tag check is local-only** — [`Scripts/release.sh:84`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/release.sh#L84)

If the tag exists on the remote but not locally, 45 min of build runs before downstream catches it. `git ls-remote --tags` pre-check would fail fast.

**13. `release.sh` removed the manual abort prompt before the remote push** — [`Scripts/release.sh:124`](https://github.com/zcash/zcash-swift-wallet-sdk/blob/kris/1745-finish-release-workflow/Scripts/release.sh#L124)

The previous script paused before any remote write; the new script pushes the release branch first, then prompts in `finalize-release.sh`. Branch pushes are non-destructive (delete-able) so this is a design change with minor ergonomic cost rather than a regression.

---

### Recommended priority

- **Block merge on**: #1, #2 — these break the very first real CI-driven release.
- **Fix before relying on the CI path**: #3, #4, #5 — retry/recovery will bite on the first transient failure.
- **Follow-up**: #6–#13.

---

### What changed from the original review

- **Removed** an asset-verification finding (`finalize-release.sh` doesn't verify the draft has its asset attached) — doesn't hold up: `gh release create FILE` is atomic; if asset upload fails the script exits non-zero before finalize ever runs.
- **Combined** two retry-path findings (originally #4 and #5) into one (now #3) — they're effectively the same issue.
- **Re-scoped** several findings that were overstated as "high severity" when they're actually narrow, ergonomic, or design choices.

---

_Reviewed locally with [Claude Code](https://claude.com/claude-code)._

nuttycom:
Closes #1745.

nuttycom:
[force-pushed](https://github.com/zcash/zcash-swift-wallet-sdk/compare/1b52c8f8e6d4e252edc09c3f7884d04d6349d939..21c180fc9624c8c9bee7b511727be1a5b2521118) to address comments from review.

.github/workflows/build-ffi.yml
```diff
@@ -1,28 +1,78 @@
...
+      # If a release-bot GitHub App is configured (repo variable
+      # RELEASE_BOT_APP_ID set and secret RELEASE_BOT_PRIVATE_KEY available),
+      # mint an installation token. App-installation tokens do trigger
+      # downstream workflows, so the release PR's required status checks
+      # (swift.yml, zizmor) will fire normally. Without the app, the push
+      # and PR creation fall back to GITHUB_TOKEN, which by design does
+      # not trigger those checks.
+      - name: Generate app token (optional)
+        id: app-token
+        if: ${{ vars.RELEASE_BOT_APP_ID != '' }}
```
nuttycom: This should probably fail if the variable is not set, instead of skipping token generation?

.github/workflows/build-ffi.yml
```diff
@@ -1,28 +1,78 @@
...
+          app-id: ${{ vars.RELEASE_BOT_APP_ID }}
+          private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
+
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
 
       - name: Select Xcode
-        run: sudo xcode-select -s /Applications/Xcode_16.0.app
+        run: sudo xcode-select -s '/Applications/Xcode_16.0.app/Contents/Developer'
```
nuttycom: Why was this change necessary?

.github/workflows/build-ffi.yml
```diff
@@ -33,21 +83,124 @@ jobs:
...
         uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
 
-      - name: Build and upload draft release
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          VERSION: ${{ github.event.inputs.version }}
-        run: ./Scripts/prepare-release.sh "$VERSION"
+      # Build the full 5-architecture XCFramework — the release artifact.
+      - name: Build XCFramework (all architectures)
+        timeout-minutes: 45
```
nuttycom: Check previous workflow runs to determine whether this timeout is long enough.

.github/workflows/build-ffi.yml
```diff
@@ -33,21 +83,124 @@ jobs:
...
+            # Derive the bot identity from the app slug; the /users/<slug>[bot]
+            # endpoint is public and resolves the bot's numeric id.
+            BOT_USER="${APP_SLUG}[bot]"
+            BOT_ID="$(gh api "/users/${BOT_USER}" --jq '.id')"
+            git config user.name "$BOT_USER"
+            git config user.email "${BOT_ID}+${BOT_USER}@users.noreply.github.com"
+            CHECKS_CAVEAT=""
+          else
+            git config user.name "github-actions[bot]"
+            git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
nuttycom: Whaere does the magic number 41898282 come from?

.github/workflows/build-ffi.yml
```diff
@@ -33,21 +83,124 @@ jobs:
...
+            CHECKS_CAVEAT=$'\n'"Note: this PR was opened by \`GITHUB_TOKEN\`, so the \`swift.yml\` and \`zizmor\` PR checks do not fire on it (GitHub's anti-recursion behavior). Configure the \`RELEASE_BOT_APP_ID\` repo variable and \`RELEASE_BOT_PRIVATE_KEY\` secret to have those checks run automatically. Verification was already performed in the release workflow run linked above."$'\n'
```
nuttycom: This caveat should no longer be correct, right? Let's force the RELEASE_BOT_APP_ID installation.

.github/workflows/build-ffi.yml
```diff
@@ -33,21 +83,124 @@ jobs:
...
+          \`\`\`
+          git fetch <remote> release/ffi-${VERSION}:release/ffi-${VERSION}
+          git checkout release/ffi-${VERSION}
+          ./Scripts/finalize-release.sh <remote> ${VERSION}
+          \`\`\`
+
+          Then merge this PR.
+          ${CHECKS_CAVEAT}
+          ---
+          *Filed with [Claude Code](https://claude.com/claude-code) assistance.*
```
nuttycom: The release PR is filed by the script, not by Claude. This line can be removed.

nuttycom:
Force-pushed `kris/1745-finish-release-workflow` (`2c490b6b` → `107987ef`) addressing each review comment:

**[`build-ffi.yml`, file-level — "reference the commit hash, not the branch name"]**

The workflow now captures `RELEASE_SHA=$(git rev-parse HEAD)` after committing Package.swift+CHANGELOG and embeds that SHA into the PR body, the workflow Summary, and the finalize invocation. The maintainer instructions are now:

```
git fetch <remote> <release-sha>
git checkout <release-sha>
./Scripts/finalize-release.sh <remote> <version> <release-sha>
```

`finalize-release.sh` gained a third required positional `<release-sha>` and validates that:
- the arg is a full 40-char SHA,
- `HEAD` equals the supplied SHA,
- `<remote>/release/ffi-<version>` tip equals the supplied SHA.

This drops the previous dependence on a local branch ref entirely — the SHA is the authoritative identifier of the release. `release.sh` was updated to pass `RELEASE_SHA=$(git rev-parse HEAD)` to its `finalize-release.sh` invocation.

**[L64 — "fail if variable not set, instead of skipping"]**

Done. The `if:` is gone from the app-token step, and `Validate inputs` now fails the workflow if `vars.RELEASE_BOT_APP_ID` is empty:

```yaml
if [[ -z "$APP_ID" ]]; then
    echo "::error::repo variable RELEASE_BOT_APP_ID is not set; configure the release-bot GitHub App"
    exit 1
fi
```

**[L75 — "why was this change necessary?" (xcode-select path)]**

It wasn't strictly necessary. Both `/Applications/Xcode_16.0.app` and `/Applications/Xcode_16.0.app/Contents/Developer` work; `xcode-select` resolves the short form. I'd switched to the longer one for cosmetic consistency with `swift.yml`. Reverted to the short form.

**[L88 — "check if 45-min timeout is long enough"]**

Past runs of the old workflow's `Build and upload draft release` step took 44m6s and 42m — too close to 45. Bumped to 60.

**[L149 — "where does 41898282 come from"]** and **[L150 — "force RELEASE_BOT_APP_ID"]**

Together: removed the entire GITHUB_TOKEN fallback path, which is where the hardcoded `41898282+github-actions[bot]@users.noreply.github.com` lived. The `Open release PR` step is now unconditional — always uses the app-installation token, always derives the bot identity from `steps.app-token.outputs.app-slug` + a public `/users/<slug>[bot]` lookup. `CHECKS_CAVEAT` and its accompanying conditional are gone.

**[L179 — "remove the Claude attribution"]**

Removed. The workflow-generated PR body no longer includes the `*Filed with Claude Code assistance.*` line — the release PR is filed by the workflow itself, not by an interactive Claude session.

**Other small things in the same fold:**

- Scratch file in the workflow moved from `/tmp/pr-body.md` to `tmp/pr-body.md` (the repo's gitignored scratch dir), consistent with the project tmp/ convention.

The branch still needs a rebase onto `main` to pick up `2.6.0-alpha.3`; I'll do that once the workflow is signed off on.

Closes #1745.

Scripts/finalize-release.sh
```diff
@@ -0,0 +1,196 @@
...
+
+# RELEASE_SHA must be a full 40-character SHA (not a prefix). Abbreviated
+# SHAs would still work for git commands but fail the equality checks below.
+if [[ ! "$RELEASE_SHA" =~ ^[0-9a-f]{40}$ ]]; then
+    echo "Error: <release-sha> must be a full 40-character commit SHA." >&2
+    exit 1
+fi
+
+# 1. Version must be an alpha/beta pre-release.
+if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+-(alpha|beta)\.[0-9]+$ ]]; then
```
nuttycom: This constraint should only apply when the run is triggered by the workflow; I believe that that's handled elsewhere, so when running the full release script by hand it shouldn't be checked here?

nuttycom: Good catch — fixed in `ea6460b3`. Relaxed `finalize-release.sh`'s check #1 to a generic semver shape check (`^[0-9]+\.[0-9]+\.[0-9]+(-[a-zA-Z0-9.]+)?$`) so stable releases (`2.6.0`) and other pre-release identifiers (`-rc.1`) flow through. The alpha/beta-only constraint stays in the workflow's `Validate inputs` step where it belongs.

nuttycom (commented):
self-review utACK

Chlup:
## Code review — round 2

Another pass over the release workflow. The earlier round clearly landed (token/zizmor hardening, the `isDraft` clobber guard, SHA-pinned finalize, `release.env`-after-success ordering all look good). Three issues remain that I'd want fixed before this merges — each verified against the actual code/git behavior, most severe first.

---

### 🔴 1. CI release-branch push does the opposite of its comment — every re-dispatch is rejected

**Location:** `.github/workflows/build-ffi.yml:160` (comment at 157–159)

```bash
git push --force-with-lease="${BRANCH}": \
```

The trailing colon with nothing after it sets an **empty** expected value. Git defines `--force-with-lease=<ref>:` with an empty `<expect>` as *"the named ref must not already exist"* — the exact opposite of the comment's claim that it "lets a retry replace its own prior release branch."

Verified with the exact form used here (explicit URL + `HEAD:branch` refspec):

```
push #1  (branch absent on remote)   →  * [new branch]            ✅ exit 0
push #2  (branch already exists)      →  ! [rejected] (stale info) ❌ exit 1
control: plain --force on push #2     →  forced update            ✅ exit 0
```

**Trigger:** The first dispatch of a version pushes `release/ffi-<version>`. Any re-dispatch of the same version — a documented recovery action (finalize's own error says *"Re-run the release workflow to recreate it"*) — fails at this push **after the full ~45-min build**, because the branch already exists and nothing deletes it between runs. It compounds with `cancel-in-progress: true`: a cancelled-then-retried run re-uploads the draft fine but can never re-push its branch, leaving an orphaned draft + stale branch.

**Suggested fix:** this is a workflow-owned, throwaway branch namespace, so a plain `--force` is correct and simplest. (If you want to keep a lease, drop the trailing colon and fetch the ref first, or pin an explicit expected OID — but `--force` is fine here.)

---

### 🟠 2. `finalize-release.sh` can advise deleting a correct, published, signed tag

**Location:** `Scripts/finalize-release.sh` — tag-state check (lines 88–114, message at 94–102) runs **before** the `HEAD == RELEASE_SHA` check (lines 116–127)

At line 94 the script compares the existing tag to `HEAD`, but it has not yet compared it to the **authoritative `RELEASE_SHA` it already holds as `$3`**. So when the tag is on the real release commit but `HEAD` is elsewhere, it concludes the *tag* is misplaced and prints:

```
Error: existing tag <v> points to <release>, but HEAD is <main>.
Delete the misplaced tag before re-running:
  git tag -d <v>
  git push <remote> :refs/tags/<v>
```

Following that advice **destroys a correct, GPG-signed release tag** (and the second line deletes it from the remote — the thing clients depend on).

**Trigger:** A prior finalize run created the tag at the release commit (even a fully successful run), the maintainer returns to a branch (`git checkout main`), then re-runs finalize (re-running is explicitly supported). Steps 1–4 all pass (a clean checkout leaves a clean tree), execution reaches the tag-state block, `TAG_LOCAL_SHA (release) != HEAD_SHA (main)`, and the destructive message fires. Step 6's accurate message ("check out the release commit") never runs.

**Suggested fix:** decide misplacement against `RELEASE_SHA`, not `HEAD`. Either move the `HEAD == RELEASE_SHA` check (step 6) ahead of the tag-state check, or branch the message on `TAG_LOCAL_SHA` vs `RELEASE_SHA` — only `TAG_LOCAL_SHA != RELEASE_SHA` is a genuinely misplaced tag; otherwise tell the user to check out the release commit.

---

### 🟠 3. `stamp-changelog.sh` is not idempotent — a `release.sh` retry ships a duplicate CHANGELOG header

**Location:** `Scripts/stamp-changelog.sh:27–37` (the awk), reached twice via the `release.sh` retry path

The awk inserts a new `# <version> - <date>` below the **preserved** `# Unreleased` header, with no guard against an existing version header. Running it twice for the same version produces two identical headers (reproduced):

```
# Unreleased

# 2.6.0-alpha.2 - <date>
# 2.6.0-alpha.2 - <date>      ← duplicate
```

**Trigger (local `./Scripts/release.sh` path only — CI is immune):** Step 2 stamps + Step 3 commits on `release/ffi-<version>`, then Step 4 `git push` fails (network/auth, or the lease bug in #1). Re-running `release.sh` is the engineered recovery (`prepare-release.sh:68` uses `git checkout -B` specifically "so re-running after a partial failure re-uses the branch"). But `checkout -B` onto the current commit is a content no-op, so the already-stamped CHANGELOG survives, `stamp-changelog.sh` runs again and double-stamps, and finalize's presence-check `grep -qE "^# ${VERSION} "` happily waves the malformed file through to tag + publish. CI is unaffected because each run starts from a fresh checkout of the base branch (only `# Unreleased`).

**Suggested fix:** make the stamp idempotent — no-op (or error) if `grep -qE "^# ${VERSION} "` already matches, mirroring finalize's own header check. Alternatively, reset `CHANGELOG.md` to the base-branch version before stamping on a re-run.

---

#1 is the one that will bite in normal operation (it silently breaks the documented re-dispatch recovery); #2 and #3 are correctness footguns on the partial-failure retry paths. Happy to push fixes for all three if useful.

<sub>🤖 Round-2 review performed with [Claude Code](https://claude.com/claude-code); findings adversarially re-verified before posting.</sub>

PR #1754, opened 2026-06-09
Small updates when creating new wallet on chain tip

Chlup:
Just very small update for code that creates new wallet on a chain tip.


LukasKorba (approved):
utACK

PR #1757, opened 2026-06-12
[MOB-1039] Multi-server transaction submission with submit-plan retry

Chlup:
Supersedes #1737. Linear: MOB-1039.

## Summary

- Reimplements multi-server transaction submission in the SDK, absorbing the fan-out logic that previously lived app-side in zodl-inc/zodl-ios#1751 (`SDKSynchronizerLive.swift`) so it is implemented and tested once.
- **`Broadcaster` API redesigned** (breaking for the 2.6.0-alpha line, see MIGRATING.md):
  - `createProposedTransactions` / `createTransactionFromPCZT` return `[CreatedTransaction]` (non-optional raw bytes) instead of `[ZcashTransaction.Overview]`.
  - The raw-bytes `submit(_:to:)` is replaced by `submit(transaction:to:timing:)`, which submits to all given endpoints in parallel — first acceptance resumes the caller immediately, in-flight submissions continue through a grace window, and the result is a structured `TransactionSubmissionOutcome` (`accepted(by:)` / `rejected` / `unreachable` / `timedOut` / `notAttempted` / `cancelled`) instead of a thrown error. A batch variant submits sequentially and stops at the first transaction that isn't accepted.
  - Carrying transaction identity through `submit` removes #1737's raw-bytes→txid correlation layer (`RawTransactionLookup`, the unindexed blob-scan query, the in-memory hex maps).
- **Submit plans persist in SQLite** under the SDK's `generalStorageURL` (per network + alias) instead of the Keychain. A row with no endpoints marks a transaction as *awaiting* (created but never submitted by the app); recorded endpoints make it *ready*. All store operations are warn-and-continue so persistence problems can never fail a send.
- **`TxResubmissionAction` is plan-aware**: it skips awaiting transactions (never auto-broadcasts something the app didn't release), retries ready ones sequentially through their recorded endpoints, and keeps byte-identical legacy behavior for transactions without a plan. Stale plans are pruned per-row from current repository state (mined / expired / missing), which replaces #1737's revision/generation token machinery and is race-free by construction.
- `wipe()` clears the plan store; legacy `Synchronizer` stream APIs are behavior-identical to `main`.

## Testing

- `swift test --filter OfflineTests` — **464 tests, 0 failures** (59 new: plan store incl. corruption/unopenable-DB degradation, race engine incl. grace-window/timeout/cancellation semantics and single-endpoint cases, executor, Broadcaster API over real localhost gRPC servers, resubmission action incl. all four prune predicates).
- SwiftLint clean on all touched files (only pre-existing `SDKSynchronizer` warnings remain).

## Notes for reviewers

- The design spec and implementation plan are included under `docs/superpowers/` for review context; happy to drop those commits before marking ready for review if they shouldn't live upstream.
- ZODL app migration to the new API is a separate follow-up; the spec's §9 contains the outcome→UI mapping table for it.


LukasKorba (approved):
utACK

PR #1759, opened 2026-06-13
Treat "already in mempool" submit rejections as success

zcash-harry:
## Summary
lightwalletd / zcashd returns `transaction already exists in mempool` or `transaction dropped because it is already queued for download` as a non-zero submit error when an identical tx was previously broadcast. The submit was actually a success — the tx is on the network — but `Synchronizer.submitTransactions` currently classifies these descriptions as `.submitFailure`, which makes downstream UI render a misleading failure screen for a tx that landed.

This PR adds a static helper `SDKSynchronizer.isAlreadyKnownToNetwork(_:)` that matches the two descriptions case-insensitively on substring (so wrapped / prefixed variants from different node versions all hit), and uses it in the `TransactionEncoderError.submitError` catch block to return `TransactionSubmitResult.success` instead of `.submitFailure`.

## Changes
- `Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift` — classify on submit catch.
- `Tests/OfflineTests/SubmitResultClassificationTests.swift` — covers exact phrase, wrapped/prefixed variants, case-insensitivity, and unrelated failures.
- `CHANGELOG.md` — Unreleased / Fixed entry.

## Test plan
- [x] `swift test --filter OfflineTests` — passes locally including the new `SubmitResultClassificationTests`.


Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
```diff
@@ -487,13 +487,22 @@ public class SDKSynchronizer: Synchronizer {
...
                     return TransactionSubmitResult.submitFailure(txId: transaction.rawID, code: code, description: message)
                 }
             }
         }
     }
-    
+
+    static func isAlreadyKnownToNetwork(_ message: String) -> Bool {
+        let lower = message.lowercased()
+        return lower.contains("already exists in mempool")
```
Chlup: I understand what you are trying to do. I just don't think that detection of this state based on error message is not a good idea. Errors messages can easily change. We need to find at least specific error codes for these states. Or some other method. In long term this can bite us.

Chlup (changes requested):

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
pacu: the string "already exists in mempool" does not exist in Zcashd's and Lightwalletd's codebase. It's a Zebrad text format to `MempoolError::InMempool` error. 

So I can confirm that this approach won't work for Zcashd backed lightwalletd's.

pacu (changes requested):

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
Cosmos-Harry: Pushed a follow-up that:

- Drops the message match entirely: when the submit catch block fires, the SDK now calls fetchTransaction on the same lightwalletd, and if the server confirms it has the tx (anything other than txidNotRecognized), we reclassify as success. 
- Backend-agnostic by construction, works for Zebra's MempoolError variants, zcashd's RPC_VERIFY_ALREADY_IN_CHAIN, and any future "already known" phrasing without depending on codes or text.
- zcashd's sendrawtransaction actually returns success on a duplicate mempool submit, so there's no error to match in the first place

Cosmos-Harry: Pushed a follow-up that:

Drops the message match entirely: when the submit catch block fires, the SDK now calls fetchTransaction on the same lightwalletd, and if the server confirms it has the tx (anything other than txidNotRecognized), we reclassify as success.
Backend-agnostic by construction, works for Zebra's MempoolError variants, zcashd's RPC_VERIFY_ALREADY_IN_CHAIN, and any future "already known" phrasing without depending on codes or text.
zcashd's sendrawtransaction actually returns success on a duplicate mempool submit, so there's no error to match in the first place

Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift
pacu: please remove force unwrapping. use `guard let` instead 🙏

pacu (changes requested):

pacu (approved):
utACK.

Ball is on @Chlup's court ☺️

Chlup (approved):

PR #1760, opened 2026-06-14
[MOB-1339] Fix TxResubmissionAction first-cycle throttle + isolate per-candidate errors

zcash-harry:
Linear: [MOB-1339](https://linear.app/zodl/issue/MOB-1339)

Two scoped fixes to `TxResubmissionAction` that close an SDK-internal race producing redundant re-broadcasts of just-submitted user transactions. Companion hotfix PR #1759 (merged) closes the user-visible `.submitFailure` surface via server-verification.

## Bug 1 — Throttle is a no-op on first invocation

`TxResubmissionAction.latestResolvedTime` is meant to throttle the action so it doesn't re-broadcast unmined txs more often than every 5 minutes. The field was initialised to `0`, so on the action's first run `diff = now - 0 ≈ 1.78 billion seconds` — overwhelmingly above the 300s threshold. The throttle was a no-op on the very first invocation, which is exactly when a freshly-submitted user transaction is most likely to appear in `findForResubmission`'s result set. The action then re-broadcasts the just-submitted tx; the resulting server error gets swallowed by the action's `catch`, but lightwalletd has still seen the redundant submit.

### Fix
```diff
-    var latestResolvedTime: TimeInterval = 0
+    var latestResolvedTime: TimeInterval = Date().timeIntervalSince1970
```
The throttle gate now engages on first invocation. The action skips the first sync cycle's resubmit, sees the user's submit complete normally, and proceeds with the next 5-minute window as designed.

**Trade-off:** a wallet that starts up with a genuinely-stuck pending tx from a previous session waits ~5 minutes for the first re-broadcast attempt. Acceptable — the manual "Refresh Transaction Data" UI path (`Synchronizer.enhanceTransactionBy`) remains available, and the alternative (the documented race) is worse.

## Bug 2 — One candidate's failure abort the whole loop

The resubmission `for`-loop was wrapped in a single `do/catch` around the whole loop, not per-iteration. A throw on one candidate's `encodedTransaction()` or `submit()` aborted the remaining candidates while still advancing `latestResolvedTime`, so any pending re-broadcasts got hidden until the next 5-minute window.

### Fix
Per-candidate `do/catch` inside the loop (in this PR's final shape, this dovetails with main's `resubmit(transaction:)` helper introduced in #1757 — the per-tx error handling lives at the same boundary, just one level of indirection cleaner). A throw on one candidate now logs (with the offending txid and underlying error) and continues to the next candidate.

## Files
- **Modified**: `Sources/ZcashLightClientKit/Block/Actions/TxResubmissionAction.swift`
- **Modified**: `CHANGELOG.md` — Unreleased / Fixed entries.

## What this PR does *not* do
- Does **not** address the user-visible `.submitFailure` surface from the race. That originates in `SDKSynchronizer.submitTransactions`'s catch block, not in the resubmit-action, and is handled in #1759 via server-verification when a submit returns a non-zero error code.
- Does **not** persist any state across SDK restarts. The throttle is per-process; on a cold start the throttle is freshly engaged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+## Fixed
+- `TxResubmissionAction.latestResolvedTime` now seeds to the current wall-clock time at construction instead of `0`. The previous zero-init made the 5-minute throttle a no-op on the action's first invocation (`diff = now - 0` is ~56 years, well over the 300s threshold), so the action could re-broadcast a freshly-submitted transaction during the very first sync cycle of the session. The throttle now engages on first invocation as intended.
+- `TxResubmissionAction` now catches per-transaction submit errors inside the for-loop instead of around it. Previously, a throw on one candidate's encode or submit aborted the remaining candidates for the cycle while still advancing the throttle, hiding pending re-broadcasts until the next 5-minute window. Errors are also logged with the offending txid and the underlying cause.
+
 ## Changed
 - New wallets now use a recent tree state from the lightwalletd server as the wallet birthday, reducing unnecessary block scanning on first launch while retaining reorg safety. Falls back to the bundled checkpoint if the server is unreachable.
 - `ZcashTransaction.Overview.State.init` now accepts an optional `expiryHeight:` argument and treats an unmined transaction whose `expiryHeight` is at or below the supplied `currentHeight` as `.expired` even when the `expiredUnmined` column hasn't been flipped to `true`. This makes the Swift-side state-machine resilient to lagging or missed updates of that column (in particular: sent transactions that were unmined when the wallet migrated across a consensus-rule change, which previously stayed reported as `.pending` indefinitely). Existing call sites that don't pass `expiryHeight` keep their prior behaviour.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 - `TxResubmissionAction.latestResolvedTime` now seeds to the current wall-clock time at construction instead of `0`. The previous zero-init made the 5-minute throttle a no-op on the action's first invocation (`diff = now - 0` is ~56 years, well over the 300s threshold), so the action could re-broadcast a freshly-submitted transaction during the very first sync cycle of the session. The throttle now engages on first invocation as intended.
-- `TxResubmissionAction` now catches per-transaction submit errors inside the for-loop instead of around it. Previously, a throw on one candidate's encode or submit aborted the remaining candidates for the cycle while still advancing the throttle, hiding pending re-broadcasts until the next 5-minute window. Errors are also logged with the offending txid and the underlying cause.
 
 ## Changed
 - New wallets now use a recent tree state from the lightwalletd server as the wallet birthday, reducing unnecessary block scanning on first launch while retaining reorg safety. Falls back to the bundled checkpoint if the server is unreachable.
--- a/Sources/ZcashLightClientKit/Block/Actions/TxResubmissionAction.swift
+++ b/Sources/ZcashLightClientKit/Block/Actions/TxResubmissionAction.swift
@@ -12,7 +12,7 @@ final class TxResubmissionAction {
         static let thresholdToTrigger = TimeInterval(300.0)
     }
     
-    var latestResolvedTime: TimeInterval = 0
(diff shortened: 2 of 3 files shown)


Chlup:
## Review notes

Mechanically the change is clean: the dedup key is consistent (the `record` call and the skip-check both resolve to `rawID`), the tracker is a correctly-shared DI singleton that survives `switchTo`, and the init ordering is crash-safe. My concern is that the core approach doesn't actually close the failure it targets, plus a few coverage/robustness gaps.

### 1. The fix doesn't prevent the user-visible failure it targets

The tracker records a txid only *after* the gRPC submit succeeds (`WalletTransactionEncoder.submit`), but the user-visible `.submitFailure` comes from the user's own `submitTransactions` call (`SDKSynchronizer.swift:491`). The resubmission action's own "already exists" error is swallowed in its `catch` (`TxResubmissionAction.swift:65`) and reaches the UI through no event, no `TransactionSubmitResult`, and no tx-state change — `run`'s `didUpdate` callback is never invoked, and `.submitFailure` is constructed only in `submitTransactions`.

So in the only ordering that actually produces a user failure — a resubmission's submit landing before the user's lazily-iterated `submitTransactions` submit — the tracker is still empty for that txid, so the new skip never fires. In the other ordering (user submit records first), there was never a user-visible failure on `main` either, because the redundant resubmission's error was already swallowed. Note this holds even on the ordering described in the PR body ("re-submits the freshly-submitted tx … which `submitTransactions` surfaces as `.submitFailure`") — the resubmission doesn't go through `submitTransactions`, so that path is swallowed.

Net effect: the change avoids a redundant (already-silent) re-broadcast, but I don't think it changes the user-visible outcome. The cleanest fix is #4 below; #5 is a simpler alternative that does close the race.

### 2. `Broadcaster.submit(_:to:)` bypasses the tracker

`SDKBroadcaster.submit(_:to:)` (`SDKBroadcaster.swift:98`) builds an ephemeral service and submits directly, never calling `submitTracker.record`. A transaction sent via this public multi-server path is persisted (so it's a `findForResubmission` candidate) but never recorded, leaving it eligible for resubmission for its entire unmined life. Only the standard `submitTransactions` path records.

### 3. In-memory state is lost on restart, and the CHANGELOG overclaims

The tracker is purely in-memory. Submit → force-close → relaunch leaves the first sync with an empty tracker, so it re-broadcasts the still-unmined tx. The PR body acknowledges this trade-off, but `CHANGELOG.md` still states unconditionally that it "Closes the session-start race." Combined with #1, that claim is optimistic.

### 4. "transaction already exists in mempool" is treated as a hard failure

Any `errorCode < 0` becomes `TransactionEncoderError.submitError` → `.submitFailure` (`WalletTransactionEncoder.swift:150`, `SDKSynchronizer.swift:489`), so a duplicate rejection — which means the tx *is* on the network — is indistinguishable from a genuine failure. Treating "already exists" as success on the submit path would close the failure screen regardless of race ordering, which the tracker can't do. This looks like the cleanest actual fix.

### 5. The documented root cause (`latestResolvedTime = 0`) is left in place

The PR identifies the global throttle's first-tick no-op (`diff = now - 0` is enormous) as the cause, then layers a new subsystem on top rather than seeding `latestResolvedTime = Date().timeIntervalSince1970` at construction (`TxResubmissionAction.swift:16`). Seeding would throttle the resubmission for the first 300s of the session, so it can't race the user's first-sync submit at all, and it also covers prior-session txs the in-memory tracker can't. Trade-off: a genuinely-stuck prior-session tx waits ~5 min for its first re-broadcast. This is a smaller change that addresses the reported race more directly.

### 6. One failed candidate aborts the whole resubmission batch

The `do/catch` wraps the entire `for` loop (`TxResubmissionAction.swift:51-67`), so a throw on one candidate's submit (a transient error, or `encodedTransaction()` throwing on a nil `raw`) skips all remaining candidates for the cycle, and `latestResolvedTime` is still advanced (~300s gate). The error log omits the txid and the underlying error. Pre-existing and unchanged here, but more reachable now that the skip path can leave a real submit for a later candidate. A per-iteration `do/catch` (continue on error) would be more robust.

### 7. Wall-clock used for interval measurement (minor)

`wasSubmittedRecently` computes `clock() - last` with a non-monotonic wall clock (`TransactionSubmitTracker.swift:30`); a backward clock step makes it negative (`< window`), suppressing that txid's re-broadcast until the clock catches up (the retention filter misbehaves the same way). Low severity — the existing throttle already uses wall-clock, the trigger is narrow (manual time change / large NTP step), and it only delays a safety-net broadcast — but a monotonic source would be correct.

---

Chlup (changes requested):
Issues added in comment above.

Chlup (dismissed):
utACK

Chlup (approved):
utACK

PR #1761, opened 2026-06-14
[PRO-336] Diagnostic logging for the enhance pipeline

zcash-harry:
Linear: [PRO-336](https://linear.app/zodl/issue/PRO-336)

> Substantially revised after [@Chlup](https://github.com/Chlup)'s feedback ("intra-cycle Task.sleep can slow the sync process by a lot — we can't afford this") plus a deeper read of what actually drives the user-visible "Receiving forever" bug. The original tracker subsystem is removed in the second commit on this branch; see commit history for the journey. A subsequent code review then surfaced that the diagnostic this PR adds had a gap on its primary failure mode — see "What changed during review" below.

## What this PR does
Adds structured, metadata-only diagnostic logging to `BlockEnhancer.enhance(at:)` so engineers can determine what's actually happening when a user reports a stuck "Receiving" transaction.

Log shape (one cycle):

```
BlockEnhancer cycle started [a7f3c1] requests=3
BlockEnhancer [a7f3c1.0] type=enhancement attempt=1
BlockEnhancer [a7f3c1.0] fetch returned status=mined(2845123) has_tx=true has_mined_height=true
BlockEnhancer [a7f3c1.0] decryptAndStoreTransaction called has_mined_height=true
BlockEnhancer [a7f3c1.1] type=enhancement attempt=1
BlockEnhancer [a7f3c1.1] type=enhancement attempt=1 error_type=ZcashError
BlockEnhancer [a7f3c1.1] type=enhancement attempt=2
...
BlockEnhancer [a7f3c1.1] type=enhancement retry exhausted after 5 attempts; will retry on next sync cycle
BlockEnhancer cycle complete [a7f3c1]
```

Opaque correlation IDs (`<6-hex>.<index>`) let engineers trace one request's journey without ever recording the txid. The previous pre-PR error log `"could not enhance transactionDataRequest \(transactionDataRequest) ..."` (which printed the txid via the enum's description) is replaced with the metadata-only variant.

## What changed during review
The original commits left `retry = false` in place immediately after the fetch returned in the `.getStatus` and `.enhancement` branches (pre-existing behavior — main has the same ordering). On a write-step failure (`setTransactionStatus` / `decryptAndStoreTransaction`), the loop exited after one attempt and the new "retry exhausted" diagnostic never fired — meaning the diagnostic was blind to exactly the stuck-transaction signature this PR is meant to make debuggable.

The third commit moves `retry = false` past the writes in both branches, so a write failure is retried up to `maxRetries` and the exhaustion log fires on final failure. `.transactionsInvolvingAddress` already had the correct ordering; the three cases are now consistent. This is a small intentional behavior change relative to main — see the CHANGELOG `## Fixed` entry.

## Why this and not the original tracker
Reading the code more carefully (and combined with a recent user-support thread): the user-visible `.submitFailure` surface produced by this race lives in `SDKSynchronizer.submitTransactions`' catch block, not in the resubmit/enhance path. The original tracker, even if conceptually correct, doesn't actually change the user-visible outcome — it just makes the SDK politer about a failure mode whose root cause we can't yet pinpoint without log evidence.

The three plausible hypotheses for PRO-336 — (1) auto-loop's `fetchTransaction` failing repeatedly, (2) server returning `Success(tx, mined_height: nil)` for a mined tx, (3) rust's `transactionDataRequests()` no longer returning the request after one mempool-only call — point to fundamentally different fixes. Logging is the prerequisite to picking the right one.

## Files
- **Modified**: `Sources/ZcashLightClientKit/Block/Enhance/BlockEnhancer.swift` — adds per-cycle and per-request structured logs, opaque correlation IDs, replaces the pre-existing txid-leaking error log with a metadata-only variant, and (review-driven) moves `retry = false` past the writes so the retry loop covers the write step on `.getStatus` and `.enhancement`.
- **Modified**: `CHANGELOG.md` — `Unreleased` / `Added` entry for the logging plus a `Fixed` entry for the retry-coverage change.

## Privacy
No transaction ids, addresses, or other PII appear in any new log line. This aligns with prior zodl-side security-review guidance that sensitive identifiers should not appear in release-build logs.

## Test plan
- [x] `swift build` passes.
- [ ] Verify in a real wallet that cycle/request logs appear with reasonable cadence and no PII.


Sources/ZcashLightClientKit/Block/Enhance/BlockEnhancer.swift
```diff
@@ -84,12 +92,27 @@ extension BlockEnhancerImpl: BlockEnhancer {
...
                 var retries = 0
-                let maxRetries = 5
+                let maxRetries = Constants.maxRetries
 
                 while retry && retries < maxRetries {
                     try Task.checkCancellation()
+
+                    if retries > 0 {
+                        let delay = retryDelay(forAttempt: retries)
+                        try? await Task.sleep(nanoseconds: UInt64(delay * 1_000_000_000))
```
Chlup: This code is called in main sync loop from `CompactBlockProcessor`. So this can potentially make sync process slower by a lot. I don't think we can afford this.

Chlup (changes requested):
I think that this PR should have [PRO-336] prefix in title.

zcash-harry:
Just settled for some debugging logs for now to troubleshot the root cause more. Otherwise, it's a fix based on a hypothesis.

Chlup (changes requested):
### Code review

Requesting changes on one issue.

**The new "retry exhausted" diagnostic never fires for a post-fetch write failure, and that failure is not retried.**

In the `.getStatus` and `.enhancement` branches, `retry = false` is set immediately after the fetch succeeds, before the `rustBackend.setTransactionStatus` / `rustBackend.decryptAndStoreTransaction` write calls:

https://github.com/zcash/zcash-swift-wallet-sdk/blob/036943486eda79f169717a16362cd0d702363cd6/Sources/ZcashLightClientKit/Block/Enhance/BlockEnhancer.swift#L105-L115

https://github.com/zcash/zcash-swift-wallet-sdk/blob/036943486eda79f169717a16362cd0d702363cd6/Sources/ZcashLightClientKit/Block/Enhance/BlockEnhancer.swift#L121-L135

If one of those writes throws, the `catch` increments `retries` but `retry` is already `false`, so `while retry && retries < maxRetries` exits without retrying the write, and the new `if retry { ... "retry exhausted ..." }` summary log is skipped:

https://github.com/zcash/zcash-swift-wallet-sdk/blob/036943486eda79f169717a16362cd0d702363cd6/Sources/ZcashLightClientKit/Block/Enhance/BlockEnhancer.swift#L185-L193

Net effect: a write failure after a successful fetch is neither retried nor surfaced by the new "retry exhausted" log. It shows up only as a single `attempt=1 error_type=...` line followed by a normal `cycle complete [<cycleID>]`, which is exactly the stuck "Receiving" transaction signature this PR is meant to make diagnosable.

Suggestion: move `retry = false` to after the write calls succeed (so a write failure is retried and the exhaustion log fires on the final attempt), or emit the exhaustion/failure summary on the write-failure path as well.

Generated with Claude Code.

Chlup (approved):
utACK

PR #1767, opened 2026-06-19
Rust: remove useless dead code annotation

### LINKED ISSUE #1768: Rust: remove useless dead code annotation

dannywillems:
The code is actually used

### DIFF
--- a/rust/src/voting/helpers.rs
+++ b/rust/src/voting/helpers.rs
@@ -70,7 +70,6 @@ pub(super) fn open_wallet_db(
         .map_err(|e| anyhow!("failed to open wallet DB: {}", e))
 }
 
-#[allow(dead_code)]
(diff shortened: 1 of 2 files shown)


rust/src/voting/helpers.rs
```diff
@@ -70,7 +70,6 @@ pub(super) fn open_wallet_db(
         .map_err(|e| anyhow!("failed to open wallet DB: {}", e))
 }
 
-#[allow(dead_code)]
 pub(super) fn round_phase_to_u32(phase: voting::storage::RoundPhase) -> u32 {
```
pacu, line 73: Could it be the case that there are versions of the SDK that actually don't use this. @LukasKorba I'm a bit lost in how our release process is with this code being on main already and our apps not surfacing any of this at the moment

Issue #353, opened 2022-03-24
Add a (swift) script to update checkpoint files


dh-ecc:
## Is your feature request related to a problem? Please describe.
Currently the json checkpoint files we use (or atleast _will be_ using after #326 is closed) are updated manually. This creates a possibility for human error, and adds work needed to keep them up to date.

## Describe the solution you'd like
Create a swift script that can be run as part of a build process or a CI step or similar, that pulls down the most recent chcekpoints we like and adds them to the bundled resources in the SDK. This should be using GRPC directly to a `lightwalletd` instance we have access to.

This:
- reduces the possibility for error
- reduces the manual work needed to keep these updated
- is testable
- Removes an issue we encountered with json responses from `lightwalletd` that we use to create the checkpoint files having varying serialisation semantics for one of it's properties.

## Alternatives you've considered
- Bash script: decided against since it's not as easily testable and could be susceptable to differences between machines (more so that a swift script)
- Status Quo: i.e continue to do this manually

## Additional context
This came out of an issue discovered as part of #326 (please have a quick read of the description for that). The idea of this came out of the realisation that we could solve two issues at once with this approach, specifically, the ambiguity of the json produced by calling the json rpc of `lightwalletd` and the fact that the process is still manual.

PR #1789, opened 2026-06-26
Add update-checkpoints skill and refresh checkpoints

Chlup:
## Summary

This branch contains two related changes:

### 1. New `update-checkpoints` skill (`93675ce3`)

A Claude Code skill under `.claude/skills/update-checkpoints/` that automates the recurring "Checkpoints updated" workflow:

- Resolves the last stored checkpoint per network and downloads new `GetTreeState` checkpoints up to the chain tip via a bundled copy of `checkmate.py` (mainnet `zec.rocks:443` / interval 2500, testnet `testnet.zec.rocks:443` / interval 10000).
- Keeps only interval-aligned files, drops the non-aligned chain-tip file, and restores the anchor afterwards, so a run is purely additive and an already-committed checkpoint can never be corrupted by a transient network error.
- Ships a verifier that validates exactly the files about to be committed — JSON shape, `network` value, height↔filename match, interval alignment, `hash`/`saplingTree`/`orchardTree` validity, and gap-free contiguity from the previous committed max.

`checkmate.py` is vendored verbatim (MIT, Electric Coin Company Prototypes) so the skill is self-contained and does not depend on a copy in a developer's `~/Downloads`.

Relates to #353 (automating checkpoint-file updates).

### 2. Checkpoint refresh (`8816e136`)

First real run of the skill:

| Network | Added | Range |
|---|---|---|
| mainnet | 14 files | `3357500 → 3390000` (interval 2500) |
| testnet | 6 files | `4040000 → 4090000` (interval 10000) |

`CHANGELOG.md` gains a `## Checkpoints` entry under `# Unreleased` listing the new files.

## Verification

- The bundled verifier passed for all 20 new files: valid JSON, correct `network`, height matches filename, interval-aligned, valid hash/trees, and contiguous from the previous committed max.
- The change is additive only — no previously committed checkpoint was modified, and the non-aligned chain tips were dropped.


LukasKorba (approved):
utACK

PR #1795, opened 2026-07-01
Update flags to init-local-ffi.sh to optimize build times 

Chlup:
## Summary

Adds three arm64-only build modes to `Scripts/init-local-ffi.sh` so local FFI iteration on Apple Silicon can skip the two slow x86_64 slices (which can't run there anyway).

| Flag | Builds | Slices |
|---|---|---|
| `--arm-macos` | `aarch64-apple-darwin` | 1 (`macos-arm64_x86_64`) |
| `--arm-ios` | `aarch64-apple-ios-sim`, `aarch64-apple-ios` | 2 |
| `--arm-all` | all three above | 3 |
| *(none)* | all 5 (unchanged, via `make xcframework`) | 3 universal |
| `--cached` | download release (unchanged) | — |

Multi-slice arm frameworks are assembled in-script and swapped in atomically. The `Info.plist` is generated via `plutil` using the same slice identifiers as the full build but arm64-only `SupportedArchitectures`, so output stays interchangeable with `rebuild-local-ffi.sh`.

## Also in this PR

- **Rename `--macos-only` → `--arm-macos`** (hard rename). Nothing external depends on the old flag — CI builds via `make macos` directly — so this is a clean rename. On Intel hosts `--arm-macos` now builds `aarch64-apple-darwin` where `--macos-only` built `x86_64-apple-darwin`, consistent with the "arm" naming.
- **Reject unknown/extra arguments** with a usage message and non-zero exit, instead of silently falling through to the full 5-architecture build.
- Update `CLAUDE.md` and `docs/LOCAL_DEVELOPMENT.md`; add a design spec under `docs/superpowers/specs/`.

## Verification

- `bash -n` clean; unknown flag and too-many-args → usage + exit 1 (no `LocalPackages/` created).
- `--arm-all` → valid 3-slice XCFramework; each slice's static lib confirmed **arm64**; every `SupportedArchitectures = [arm64]`; plist lint-clean and structurally identical to `BuildSupport/Info.plist`.
- `--arm-ios` → exactly 2 slices; `--arm-macos` → exactly 1 slice; both near-instant on a warm Cargo cache (incremental).
- `swift build` against the `--arm-macos` slice → **Build complete!**, exit 0 (only pre-existing SwiftProtobuf deprecation warnings).

## Note

No linked issue — this is a developer-tooling change to the local FFI build script. Flagging it since `CONTRIBUTING.md` normally asks that PRs reference an issue.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -32,7 +32,7 @@ The Rust code in `rust/` is compiled into the `libzcashlc` XCFramework. Two mode
 
 Scripts:
 
-- `./Scripts/init-local-ffi.sh` — one-time setup; default builds all 5 architectures and creates `LocalPackages/`. **`--macos-only`** builds only the macOS slice from your `rust/` (good for `swift build` / `swift test` on the Mac). Use `--cached` only when your branch has no FFI changes relative to the release. Use --macos-only to rebuild for fast local development.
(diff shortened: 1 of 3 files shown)


LukasKorba (approved):
utACK

PR #1799, opened 2026-07-03
[MOB-1096] Retry Tor rust-layer errors instead of failing sync fatally

zcash-harry:
## Summary
- Tor rust-layer errors (`rustTorConnectToLightwalletd`, `rustTorLwdGetInfo`, `rustTorLwdSubmit`, `rustTorLwdFetchTransaction`, `rustTorLwdLatestBlockHeight`, `rustTorLwdGetTreeState`) are `ZcashError` cases, so they fell through `CompactBlockProcessor`'s retry-classification switch to `serviceError = false` and went straight to a fatal `handleSyncFailure` — no retry, no reconnect.
- This is the most frequently reported issue in support (MOB-1096): errors like "remote hostname lookup failure", "Failed to obtain exit circuit for ports", and "Tor network protocol violation" require a full force-quit + relaunch to recover, sometimes losing in-progress send state.
- Fix: classify these Tor errors alongside the existing `service*Failed` cases, so they go through the same reset-and-retry path (which also tears down cached Tor connections via `service.closeConnections()`) up to `ZcashSDK.serviceFailureRetries` (3) times, instead of failing fatally on the first transient Tor hiccup.

## Test plan
- [x] `swift build` succeeds
- [x] `swift test --filter OfflineTests` — 473 tests pass
- [ ] Manual verification with a flaky/dropped Tor circuit (hard to simulate offline; relying on code review + existing retry-path coverage for other transport errors)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+## Fixed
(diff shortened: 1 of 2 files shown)


Chlup (approved):
utACK

PR #1802, opened 2026-07-07
Mark GitHub releases as pre-releases for SemVer pre-release versions

Chlup:
Part of #1745.

## Summary

Releasing a version like `2.6.0-alpha.6` currently creates a normal GitHub release, and someone has to tick the "pre-release" checkbox in the UI by hand (the existing `2.6.0-alpha.*` releases were marked that way manually). This PR makes the release scripts do it automatically.

- `Scripts/prepare-release.sh` detects a SemVer pre-release suffix (any hyphen in the version, e.g. `2.6.0-alpha.1`, `2.7.0-rc.2`) and passes `--prerelease` to `gh release create`. Since this script is the single place releases are created, the fix covers both the local flow and the `Build FFI XCFramework` workflow. When `--force-overwrite-existing-release` updates an already-existing release, the pre-release bit is applied with a follow-up `gh release edit`, since `gh release upload` can only replace assets.
- `Scripts/release.sh` passes `--prerelease` when publishing the draft (`gh release edit`), includes the flag in the printed manual-resume command, and announces the pre-release status in its output.
- `docs/ci.md` documents the auto-detection.

Auto-detection was chosen over an explicit flag so it can't be forgotten: per SemVer, a hyphenated suffix *is* what defines a pre-release, and it matches this repo's tag history exactly (all `-alpha.*` tags are pre-releases on GitHub, all stable tags are not).

## Testing

- `bash -n` passes on both scripts.
- Detection and flag expansion exercised under macOS `/bin/bash` 3.2.57 (the interpreter these scripts run with) for `2.5.2`, `2.6.0-alpha.7`, and `2.7.0-rc.1`: stable versions produce exactly the previous `gh` invocations; pre-release versions append `--prerelease` to both `gh release create` and `gh release edit`.
- The `--force-overwrite-existing-release` path exercised the same way with a stubbed `gh`: pre-release versions follow the asset upload with `gh release edit --prerelease`; stable versions skip the edit call.
- Verified the installed `gh` CLI supports `--prerelease` on both `release create` and `release edit`.

### DIFF
--- a/Scripts/prepare-release.sh
+++ b/Scripts/prepare-release.sh
@@ -9,6 +9,9 @@
 #   4. Writes release info to BuildSupport/products/release.env
 #   5. Outputs the values needed for Package.swift
 #
+# Versions with a SemVer pre-release suffix (e.g. 2.6.0-alpha.1, 2.7.0-rc.2)
+# are detected automatically and the GitHub release is marked as a pre-release.
+#
 # After running this script:
 #   1. Update Package.swift with the URL and checksum
 #   2. Commit the Package.swift change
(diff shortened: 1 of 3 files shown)


LukasKorba (dismissed):
utACK

LukasKorba (commented):
utACK

LukasKorba (approved):
utACK

PR #1811, opened 2026-07-17
Update CLAUDE.md: run cargo fmt before committing Rust changes

Chlup:
## Summary

Adds a convention to `CLAUDE.md` instructing that `cargo fmt` be run in `rust/` before committing Rust changes, to keep diffs minimal and avoid spurious conflicts when rebasing.

## Changes

- Added a **Rust formatting** bullet under "Conventions and gotchas" in `CLAUDE.md`.

## Test plan

- Documentation-only change (agent guidance); no code affected.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -101,6 +101,7 @@ Generated files and `Tests/` are excluded from the main `.swiftlint.yml` (tests 
 - **TODOs**: format as `TODO: [#<issue_number>] ...` — bare `TODO:`/`FIXME:` warn.
 - **SwiftLint disables**: only the exceptions listed in `SWIFTLINT.md` are permitted, always scoped with `// swiftlint:disable:next` / `disable:previous` / region blocks.
 - **Commits and PRs**: every PR must reference an issue. Commit title format is `[#<issue_number>] <self-descriptive title>` (see `CONTRIBUTING.md`). PRs are typically squash-merged.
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -101,6 +101,7 @@ Generated files and `Tests/` are excluded from the main `.swiftlint.yml` (tests
 - **TODOs**: format as `TODO: [#<issue_number>] ...` — bare `TODO:`/`FIXME:` warn.
 - **SwiftLint disables**: only the exceptions listed in `SWIFTLINT.md` are permitted, always scoped with `// swiftlint:disable:next` / `disable:previous` / region blocks.
 - **Commits and PRs**: every PR must reference an issue. Commit title format is `[#<issue_number>] <self-descriptive title>` (see `CONTRIBUTING.md`). PRs are typically squash-merged.
+- **Rust formatting**: always run `cargo fmt` in `rust/` before committing changes to the Rust code. Consistent formatting keeps diffs minimal and avoids spurious conflicts when rebasing.
```
pacu, line 104: I personally prefer that formatting is done on its own commit. That way allows the commit to be less noisy. It's a subtle difference but yet substantial when reviewing code

nuttycom: @pacu the point here is that *every* commit should be formatted. There should not be any standalone formatting commits required. If current `main` is not correctly formatted, then yes we need a formatting commit, but that should be the last standalone formatting commit ever.

nuttycom: Current `main` formats clean, so this instruction is correct.

nuttycom (approved):

PR #1807, opened 2026-07-17
[#1806] Ironwood (NU6.3) receive/sync readiness

Chlup:
Part of #1806.

This PR makes the SDK Ironwood-ready: NU6.3 third-pool receive/sync support end to end (protocol fields → Rust core → welding → Swift balance surface), riding the Ironwood-era dependency graph, plus the custom-network activation-heights support needed to exercise the path before mainnet/testnet activation.

## What's included

- **Ironwood (NU6.3) receive/sync readiness** — lightwalletd proto fields (`CompactTx.ironwoodActions`, `ChainMetadata.ironwoodCommitmentTreeSize`, `TreeState.ironwoodTree`, `ShieldedProtocol.ironwood`), `putIronwoodSubtreeRoots` (FFI + welding + `UpdateSubtreeRootsAction`, skipping gracefully on servers that don't serve the fields yet), `Checkpoint.ironwoodTree`, `AccountBalance.ironwoodBalance` (repr(C) + Swift), `WalletSummary.nextIronwoodSubtreeIndex`, and shielded-pool convenience accessors so hosts never hand-sum pools. Dormant until NU6.3 activates and a lightwalletd serves the fields.
- **Dependency graph** — the Rust core rides `zcash/librustzcash` main (Ironwood merged + ungated there), rev-pinned via `[patch.crates-io]` until the 0.24-era crates.io release; `orchard` 0.15.0 and `shardtree` 0.7.0 from crates.io. Protobuf Swift stubs regenerated with pinned `swift-protobuf` 1.35.1 (isolated in its own commit so the Ironwood field diff stays readable).
[...]

Issue #1806, opened 2026-07-17
Ironwood (NU6.3) support


Chlup:
Tracking issue for making the SDK Ironwood-ready on the `release/2.6.0` line:

- lightwalletd protocol fields for the third shielded pool (`CompactTx.ironwoodActions`, `ChainMetadata.ironwoodCommitmentTreeSize`, `TreeState.ironwoodTree`, `ShieldedProtocol.ironwood`)
- Ironwood subtree-root storage and the balance surface (`AccountBalance.ironwoodBalance`, wallet-summary support)
- the Ironwood-era dependency graph (`zcash/librustzcash` main via `[patch.crates-io]`, orchard 0.15)
- custom-network (regtest-style) activation heights so the path can be exercised before NU6.3 activates

A PR implementing the receive/sync readiness follows. Ironwood send/spend beyond the balance surface is follow-up work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1812, opened 2026-07-18
[#1806] Orchard→Ironwood pool-migration Synchronizer surface (engine, host, privacy gate)

Chlup:
Exposes the complete Orchard→Ironwood pool-migration flow to hosts **through the `Synchronizer` protocol** — the app-facing surface is the protocol plus the public data types; everything behind it is internal. Builds on the pool-migration FFI/welding layer of the stacked base PR #1813 (Cargo pin advance, `zcashlc_migration_*` FFI, `@DBActor` welding, public models, error codes); the slipstream layer rebases on top of this branch.

Stack: `release/2.6.0` ← #1807 ← #1813 ← **this PR** ← slipstream.

## What's here

1. **The internal migration machinery** — the per-account `OrchardMigration` engine actor, the dedicated-Tor `MigrationBroadcaster`, the persisted 10-minute `MigrationSyncGate`, backup-excluded storage provisioning, and the per-synchronizer `OrchardMigrationHost` that lazily builds per-account engines over **one shared broadcaster** (fixing the latent cross-account double Tor bootstrap of per-actor construction).
2. **Public API — the `Synchronizer` migration group.** 23 new protocol requirements: state machine/progress, note split, gradual + immediate schedule proposal, one-confirmation sign-and-store, height-gated `executeNextPendingMigrationTransfer`, engine-backed reschedule readback, overdue/invalid detection + restart/refresh recovery, sub-0.01-ZEC residual opt-in, the full external-signer (PCZT) path for hardware wallets, and the wallet-scope privacy gate (`isMigrationSyncBlocked()` / `migrationSyncBlockedStream` / `migrationPrivacySyncBufferDuration`). Account-scoped members take an `AccountUUID`; **two migrations can run concurrently** (e.g. a software account and a Keystone account) — all engine state is account-keyed, per-account broadcasts stay single-flight without cross-account serialization, and every account shares one migration Tor runtime. The previously-public `OrchardMigration` actor is internal. New requirements ship with throwing protocol-extension defaults so stacked conformers keep compiling; `SDKSynchronizer` (this branch) and `SlipstreamSynchronizer` (stacked branch) implement every member — pinned by tests. `ZcashNetwork.ironwoodActivationHeight` replaces hardcoded NU6.3 heights. Migration members don't require `prepare()`, so background sessions can broadcast without starting sync. The Closure/Combine wrappers deliberately do not mirror the group (matches the repo's recent convention).
3. **Privacy invariants (enforced by construction, reviewed adversarially):**
   - Migration broadcasts run over a **dedicated Tor runtime** (`<torDir>/migration_tor`, isolated circuit per submission, one bootstrap shared across accounts), independent of the global `tor(enabled:)` flag.
   - **Fail closed**: Tor requested but unavailable ⇒ `ZcashError.migrationTorUnavailable`; never a silent clearnet fallback.
   - **Exactly one submission endpoint per attempt, chosen explicitly by the host**: `MigrationNetworkPrivacyOptions.submissionEndpoint` is **required** — the SDK never defaults to (or falls back on) the sync server; no multi-endpoint fan-out; the regular-send `Broadcaster` is untouched.
   - **No txid polling**: the migration files contain no transaction-fetch path; confirmation comes from block scanning (engine-enforced).
   - **Session separation is SDK-enforced in both directions**: `start()` throws `migrationSyncBlocked` (ZRUST0125) while the wallet-scope 10-minute broadcast→sync gate is active (aggregated across every account, including dormant ones with only a persisted gate file), and the broadcast-performing members throw `migrationBroadcastDuringSync` (ZRUST0126) while a sync runs — an advisory point-in-time check, not a hard lock, so hosts should still sequence sessions.
4. **Tests** — the migration suites inside OfflineTests: pure engine/gate/broadcaster logic over mocked welding, actor composition over injected seams, keystone prep-first flow, gate math, host composition (per-account caching, shared-bootstrap canary, dormant-account gate aggregation, degrade-open), protocol-defaults pinning, SDKSynchronizer forwarding + both enforcement errors (proven red-first), a deallocation-leak regression for the host registration, and the actor-integration-over-real-FFI additions to `MigrationFFITests` (the welding-only parts of that file ride #1813). Full suite at this tip: **616 tests, 0 failures** locally.


rust/src/migration.rs
```diff
@@ -262,6 +291,106 @@ fn clear_invalid_marks(wallet: &mut MigrationWallet, account: &[u8; 16]) -> anyh
...
+/// Persists the account's immediate-run record, replacing any previous one: only the most
+/// recently broadcast immediate sweep is ever tracked (one row per account).
+fn record_immediate_run(
+    conn: &Connection,
+    account: &[u8; 16],
+    txid: [u8; 32],
+    recorded_at_height: BlockHeight,
+) -> rusqlite::Result<()> {
+    conn.execute(
+        "INSERT OR REPLACE INTO sdk_immediate_runs (account_uuid, txid, recorded_at_height)
```
nuttycom, line 334: FIXME: this needs to use `ext_` prefixing.

rust/src/migration.rs
```diff
@@ -262,6 +291,106 @@ fn clear_invalid_marks(wallet: &mut MigrationWallet, account: &[u8; 16]) -> anyh
...
+/// The account's raw immediate-run row, if any. Cheap (touches only this SDK-owned table), so
+/// callers can check for a row's existence before paying for a wallet-database chain-tip lookup
+/// (which errors on a not-yet-synced wallet — see the callers in `zcashlc_migration_state` /
+/// `zcashlc_migration_progress`).
+fn immediate_run_row(
+    conn: &Connection,
+    account: &[u8; 16],
+) -> rusqlite::Result<Option<ImmediateRunRow>> {
+    conn.query_row(
+        "SELECT txid, recorded_at_height FROM sdk_immediate_runs WHERE account_uuid = ?1",
```
nuttycom, line 350: FIXME: `ext_`

nuttycom (approved):
lightly-reviewed utACK with a couple of required follow-up comments. In order to avoid facing conflicts multiple times, I plan to merge the full stack to feature/ironwood-slipstream and then apply required changes as follow-ups

PR #1813, opened 2026-07-20
[#1806] Orchard→Ironwood pool-migration FFI, welding, models, and error codes

Chlup:
> [!IMPORTANT]
> **Round 5 (2026-07-21): rewired onto the FINAL migration engine.** The `zcash_pool_migration` v1 crate is gone. The librustzcash family now pins the tip of `michal/MOB-1455/MOB-1495-pool-migration-account-key` @ `1974ce536c` — `feat/pool-migration-sqlite` (the final `zcash_pool_migration_backend` + `zcash_pool_migration_sqlite` crates) extended by the **account-keyed migration store** (draft [librustzcash PR #2712](https://github.com/zcash/librustzcash/pull/2712)), which the multi-account wallet (software + imported Keystone accounts in one wallet DB) requires. `orchard` pins the merged #535 deferred-anchor commit (`fa6e5fee`) the family itself patches; libzcashlc is versioned `2.6.0`.
>
> The C ABI is preserved except the external-signer note-split pair, which went **plural** (`zcashlc_migration_create_unsigned_note_split_pczts` / `zcashlc_migration_store_signed_note_split_pczts`): the final engine builds N preparation transactions, not one split transaction. What v1 rust did internally now lives in this layer: the propose→commit plan cache (`MIGRATION_PLAN_STALE` → new `migrationPlanStale` ZRUST0128), prove-at-broadcast (ZIP 374; **transfers currently prove against the natural anchor — an approved, loudly-documented ZIP 318 stopgap pending upstream anchor-checkpoint retention, librustzcash #2700**; hard proving failures → new `migrationProvingUnavailable` ZRUST0127), mined-reconciliation, rejection marks (`sdk_invalid_marks`), the 6-state derivation (`Complete` is **per-run**; `ReadyToPropose`/`SyncRequiredBeforeNext` never emitted), and sequential runs (commit over a terminal run starts the next). `includeResidual` is documented-inert. Branch gates: `cargo test` 40/0, OfflineTests **511/0**.

Binds the `zcash_pool_migration` crate (Orchard→Ironwood pool-migration engine, librustzcash `michal/ironwood-migration` @ `5115cf26`) into the SDK: the Cargo pin advance, the complete `zcashlc_migration_*` FFI surface, the `@DBActor` welding, the public migration model types, and the full migration error vocabulary. **No `Synchronizer` API changes here** — the app-facing migration surface (the `Synchronizer` migration group, engine actor, host, broadcaster, and privacy gate) lands in the stacked follow-up #1812, which is based on this branch.

Stacked on the Ironwood receive/sync-readiness line (#1807). Split out of #1812 so the FFI/welding layer can review and land independently of the Synchronizer surface.

## What's here

1. **Cargo** — the `[patch.crates-io]` librustzcash family advances from `4faf8711` (main) to `5115cf26`, the `michal/ironwood-migration` tip: its main base, plus a cherry-pick of main's `e9581a78` "Add Ironwood historical note selection" (required by `zcash_voting@464f9748`), plus the merged `pr-2572` line — **{1, 2, 5}×10ⁿ migration denominations with a 0.01 ZEC threshold** (ZIP-318-conformant) and the `pending_transfer_proposal()` accessor. `zcash_pool_migration` rides a direct git dep at the same rev — one family instance for the whole graph. Returns to a main rev in one move once the crate lands upstream.
2. **FFI** — 22 `zcashlc_migration_*` functions + `zcashlc_ironwood_activation_height`, house conventions throughout (catch_panic, thread-local last-error, paired free functions, `Ffi*` header names).
3. **Welding + models + errors** — the `@DBActor` welding group on `ZcashRustBackendWelding`, the public model types (`MigrationState`, `MigrationProgress`, `NoteSplitProposal`, `MigrationTransferProposal`, `MigrationSchedule`, `PreparedMigrationTransfer`, `MigrationTransferResult`, `MigrationAttentionReason`, `MigrationUnsignedTransferPczt`, `MigrationSignedTransferPczt`), and error codes `ZRUST0098–0108`, `0111–0126` (0093–0097 stay slipstream's; 0109–0110 are the base's Ironwood subtree-roots codes). The error vocabulary ships complete here, **including the codes whose throwing call sites arrive with #1812** (`migrationTorUnavailable`, `migrationRecordFailedAfterBroadcast` ZRUST0124, `migrationSyncBlocked` ZRUST0125, `migrationBroadcastDuringSync` ZRUST0126) — kept together so the generated error trio is touched exactly once in the stack.
4. **Welding-layer hardening from the adversarial review rounds** — the FFI last-error channel is cleared before every sentinel read (a sentinel producer that sets-but-doesn't-throw can no longer pollute a later legitimate `false`/`nil` answer on the same thread), one-time rust initialization is race-free, `withScheduleFFIArgs` deduplicates the schedule → parallel-arrays marshaling, and PCZT marshaling is zero-copy where the FFI's ownership contract allows.
5. **Tests** — the welding-scoped suites inside OfflineTests: the real-FFI empty-db state machine, marshaling determinism, and last-error-channel hygiene (`MigrationFFITests`), the txid byte-order round-trip pin (`TxIdTests`), and the error-localization additions. The actor/composition/host/Synchronizer suites ride #1812 together with the code they exercise.

## Stack

`release/2.6.0` ← #1807 ← **this PR** ← #1812 ← slipstream

---

## Round 6 (2026-07-22) — real-main pin, locking, rounds estimate

- **librustzcash family pin → `e51d502c7d`**, the recomposed standing aggregation branch: librustzcash **main** (now carrying the pool-migration engine, the merged account-keyed store inside `zcash_client_sqlite::pool_migration` — #2712 — and the run-count estimate with signing sessions — #2714) plus the still-unmerged **note-locking (#2716)** and **boundary-anchor proving (#2710)** cherry-picks; orchard → `3bd2b736` (the rev the pinned librustzcash patches to).
- **Store port**: the separate `zcash_pool_migration_sqlite` crate is gone; the store now lives in `zcash_client_sqlite::pool_migration` (same `PoolMigrations::for_account` shape, `AccountUuid`-typed) and its tables ride the standard `init_wallet_db` schema migrations — `open()` no longer creates them.
- **New surface**: `zcashlc_migration_lock_residual` / `zcashlc_migration_unlock_residual` (residual "Lock balance" primitive, ZRUST0132/0133), `locked_value` on the FFI `Balance` + `PoolBalance.lockedValue` (locked funds leave spendable but stay in the total), and `zcashlc_migration_estimate_runs` + the `MigrationRunEstimate` model (multi-run "rounds" preview incl. per-run signing-session math for capacity-limited signers, ZRUST0134).
- **Proving flip attempted, blocked upstream**: retention (#2700/#2710) landed but retains checkpoints every **288** blocks while boundaries draw on the **144**-block grid — flipping would strand the odd-multiple half of draws. The natural-anchor stopgap stays, now documented against the concrete blocker (see `HANDOFF-MOB-1495-R5-proving-flip.md` in the workspace for the fix paths).
- Gates: cargo 43/43 · OfflineTests 526/0 at this branch's tip (`794c38b4`).


rust/src/migration_engine.rs
```diff
@@ -0,0 +1,199 @@
...
+
+/// The migration backend for one account of this SDK's wallet database.
+pub(crate) struct Backend<'a> {
+    wallet: &'a MigrationWallet,
+    account: AccountUuid,
+    usk: Option<UnifiedSpendingKey>,
+    store: PoolMigrations<&'a mut rusqlite::Connection>,
+}
+
+impl<'a> Backend<'a> {
```
dannywillems, line 64: There will be a new trait to implement here, introduced in https://github.com/zcash/librustzcash/pull/2710

rust/src/migration_finalize.rs
```diff
@@ -0,0 +1,201 @@
...
+/// unused while the upstream retention grid, 288 blocks, misses half the 144-block boundary draw
+/// grid; flipping now would stall those transfers forever).
+pub(crate) fn resolve_proving_anchor(
+    wallet: &MigrationWallet,
+    _tx: &MigrationTransaction,
+) -> anyhow::Result<BlockHeight> {
+    natural_anchor_height(wallet)
+}
+
+/// Attempts to complete one migration transaction's PCZT: matches every redacted (witness-less)
```
dannywillems: This would need to be changed when https://github.com/zcash/librustzcash/pull/2710 is merged.

rust/src/migration.rs
```diff
@@ -0,0 +1,2575 @@
...
+
+/// An optional block height as an `i64`, with `-1` standing for "none".
+fn height_opt_to_i64(h: Option<BlockHeight>) -> i64 {
+    h.map_or(-1, |h| i64::from(u32::from(h)))
+}
+
+/// A count as a `u32`, erroring (rather than truncating) on overflow. The engine's per-run counts
+/// (crossings, layers, transactions) are bounded by the note cap, so overflow never happens in
+/// practice; this keeps the marshaling honest anyway.
+fn count_to_u32(v: usize, what: &str) -> anyhow::Result<u32> {
```
dannywillems, line 123: I guess this could go in a utils or upstream or reuse existing methods.

rust/src/migration.rs
```diff
@@ -0,0 +1,2575 @@
...
+}
+
+/// A spendable-value amount as a signed 64-bit integer (zatoshi). Every migration amount is a
+/// valid [`Zatoshis`] (`<= MAX_MONEY`, ~2.1e15), well within `i64`.
+fn zat_to_i64(z: Zatoshis) -> i64 {
+    u64::from(z) as i64
+}
+
+/// An optional block height as an `i64`, with `-1` standing for "none".
+fn height_opt_to_i64(h: Option<BlockHeight>) -> i64 {
```
dannywillems, line 116: I guess this could go in a utils or upstream or reuse existing methods.

rust/src/migration.rs
```diff
@@ -0,0 +1,2575 @@
...
+
+/// Proving a migration transaction failed hard (as opposed to the transient "not witnessable yet"
+/// state, which is reported as "nothing due").
+fn proving_unavailable(detail: impl std::fmt::Display) -> anyhow::Error {
+    anyhow!("{PROVING_UNAVAILABLE_PREFIX}: {detail}")
+}
+
+/// A spendable-value amount as a signed 64-bit integer (zatoshi). Every migration amount is a
+/// valid [`Zatoshis`] (`<= MAX_MONEY`, ~2.1e15), well within `i64`.
+fn zat_to_i64(z: Zatoshis) -> i64 {
```
dannywillems, line 111: I guess this could go in a utils or upstream or reuse existing methods.

rust/src/migration.rs
```diff
@@ -0,0 +1,2575 @@
...
+
+/// A commit was requested without a matching previewed plan (process restart between propose and
+/// confirm, or the wallet changed underneath the preview). The platform re-proposes.
+fn plan_stale(detail: &str) -> anyhow::Error {
+    anyhow!("{PLAN_STALE_PREFIX}: {detail}")
+}
+
+/// Proving a migration transaction failed hard (as opposed to the transient "not witnessable yet"
+/// state, which is reported as "nothing due").
+fn proving_unavailable(detail: impl std::fmt::Display) -> anyhow::Error {
```
dannywillems: I guess this could go in a utils or upstream or reuse existing methods.

rust/src/migration.rs
```diff
@@ -0,0 +1,2575 @@
...
+// ----- error / value marshaling -----
+
+/// The stable prefix the Swift layer maps to `ZcashError.migrationPlanStale` (ZRUST0128).
+const PLAN_STALE_PREFIX: &str = "MIGRATION_PLAN_STALE";
+/// The stable prefix the Swift layer maps to `ZcashError.migrationProvingUnavailable` (ZRUST0127).
+const PROVING_UNAVAILABLE_PREFIX: &str = "MIGRATION_PROVING_UNAVAILABLE";
+
+/// A commit was requested without a matching previewed plan (process restart between propose and
+/// confirm, or the wallet changed underneath the preview). The platform re-proposes.
+fn plan_stale(detail: &str) -> anyhow::Error {
```
dannywillems, line 98: I guess this could go in a utils or upstream or reuse existing methods.

rust/src/migration.rs
```diff
@@ -0,0 +1,2575 @@
...
+    DerivedState::InProgress {
+        completed_transfers: completed,
+        total_transfers: transfers.len() as u32,
+        next_transfer_ready_at_height: next_ready,
+    }
+}
+
+/// The amount a stored transfer crosses, from the run's funding notes (`Transfer { crossing }`
+/// indexes into them).
+fn transfer_amount(state: &MigrationState, tx: &MigrationTransaction) -> Option<Zatoshis> {
```
dannywillems, line 923: This should be moved upstream in librustzcash.

rust/src/migration.rs
```diff
@@ -0,0 +1,2575 @@
...
+        Ok(Box::into_raw(Box::new(FfiUnsignedTransferPczts {
+            ptr,
+            len,
+        })))
+    }
+}
+
+/// Build an owned C string from `s`, erroring (rather than panicking across the FFI) if it
+/// contains an interior NUL byte.
+fn cstring_raw(s: &str, what: &str) -> anyhow::Result<*mut c_char> {
```
dannywillems, line 1256: Could be moved in a `utils.rs` or something independent of `migration.rs`.

rust/src/migration.rs
```diff
@@ -0,0 +1,2575 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rand::SeedableRng;
+    use rand::rngs::StdRng;
+    use zcash_pool_migration_backend::note_splitting::NoteSplitPlan;
+    use zcash_pool_migration_backend::preparation::PreparationPlan;
+    use zcash_pool_migration_backend::scheduling;
+
+    fn zat(v: u64) -> Zatoshis {
```
dannywillems, line 2755: This exists/should exist upstream/can be inlined.

rust/src/migration.rs
```diff
@@ -0,0 +1,2575 @@
...
+        Zatoshis::from_u64(v).unwrap()
+    }
+
+    fn h(v: u32) -> BlockHeight {
```
dannywillems, line 2759: This exists/should exist upstream/can be inlined.

rust/src/lib.rs
```diff
@@ -1034,7 +1038,7 @@ pub unsafe extern "C" fn zcashlc_get_verified_transparent_balance(
                 target,
                 confirmations_policy,
                 CoinbaseFilter::AllTransparentOutputs,
-                LockFilter::Unfiltered,
+                LockFilter::Policy(&LockedInputPolicy::Exclude),
```
nuttycom: I think that for balance, we should not be excluding locked notes? We should always report the full balance of the wallet.

rust/src/lib.rs
```diff
@@ -1099,7 +1103,7 @@ pub unsafe extern "C" fn zcashlc_get_verified_transparent_balance_for_account(
                         target,
                         confirmations_policy,
                         CoinbaseFilter::AllTransparentOutputs,
-                        LockFilter::Unfiltered,
+                        LockFilter::Policy(&LockedInputPolicy::Exclude),
```
nuttycom: Same

rust/src/lib.rs
```diff
@@ -1151,7 +1155,7 @@ pub unsafe extern "C" fn zcashlc_get_total_transparent_balance(
                 target,
                 wallet::ConfirmationsPolicy::new_symmetrical(NonZeroU32::MIN, true),
                 CoinbaseFilter::AllTransparentOutputs,
-                LockFilter::Unfiltered,
+                LockFilter::Policy(&LockedInputPolicy::Exclude),
```
nuttycom: Same

rust/src/lib.rs
Chlup: Fixed in 90306346 — reverted to `LockFilter::Unfiltered`. Balance reporting is lock-inclusive again; only the selection paths (`propose_send_max_transfer`, residual locking, the migration engine) keep excluding locked notes.

rust/src/lib.rs
Chlup: Fixed in 90306346 — `LockFilter::Unfiltered`.

rust/src/lib.rs
Chlup: Fixed in 90306346 — `LockFilter::Unfiltered`.

rust/src/migration.rs
```diff
@@ -0,0 +1,4420 @@
...
+fn init_invalid_marks(conn: &Connection) -> rusqlite::Result<()> {
+    conn.execute_batch(
+        "CREATE TABLE IF NOT EXISTS sdk_invalid_marks (
+            account_uuid BLOB NOT NULL,
+            tx_id INTEGER NOT NULL,
+            reason TEXT NOT NULL,
+            PRIMARY KEY (account_uuid, tx_id)
+        )",
+    )
+}
```
nuttycom: If we need additional API methods in the engine we should add them there, not work around their absence here. Also, this table does not use the `ext_` prefix required for tables that are stored in the wallet database - we *must* be able to handle all database updates via sql migrations. Blocking.

nuttycom (dismissed):

Sources/ZcashLightClientKit/Model/MigrationModels.swift
```diff
@@ -0,0 +1,339 @@
...
+    /// nothing-to-migrate proposal).
+    public let proposalHandle: UInt64
+
+    /// Creates a `NoteSplitProposal`.
+    public init(outputNotes: [Zatoshi], fee: Zatoshi, proposalHandle: UInt64) {
+        self.outputNotes = outputNotes
+        self.fee = fee
+        self.proposalHandle = proposalHandle
+    }
+}
```
nuttycom, line 76: Note to self: look for the construction of this, because "Split" mischaracterizes the operation - this could either be splitting or aggregation, depending on the wallet.

Sources/ZcashLightClientKit/Model/MigrationModels.swift
```diff
@@ -0,0 +1,339 @@
...
+
+/// A single scheduled Orchard -> Ironwood transfer, as one element of a `MigrationSchedule`.
+public struct MigrationTransferProposal: Identifiable, Equatable, Sendable, Codable {
+    /// The transfer's opaque, engine-issued id.
+    public let id: String
```
nuttycom: I'm surprised this is a string.

rust/src/migration.rs
nuttycom: This is kind of a Fairbairn threshold violation: https://wiki.haskell.org/Fairbairn_threshold

rust/src/migration_finalize.rs
```diff
@@ -0,0 +1,164 @@
...
+/// persisted `anchor_boundary()`; a missing boundary is a corrupt store and a hard error), a
+/// PREPARATION via [`engine::prove_preparation`] with the caller-resolved `natural_anchor`. On
+/// success the engine has stored the proven bytes into `state` (`Signed -> Proved` via
+/// `set_transaction_proved`); the caller persists `state` and re-reads the proven PCZT from it.
+///
+/// `natural_anchor` is `Option`al because only a PREPARATION uses it — the caller resolves it
+/// lazily, per kind, so proving a transfer never depends on the natural anchor being resolvable
+/// (a wallet with a chain tip but no scanned blocks yet has none, and must still prove transfers
+/// against their persisted boundaries). A preparation reaching this function without one is a
+/// caller bug, surfaced as a hard error rather than a silent wrong anchor.
```
nuttycom, line 103: I don't understand this "natural anchor" concept. Each transaction is assigned an anchor; it cannot have proofs generated before that anchor height, but the proving can be done at any time once the wallet has fully scanned past that anchor height. Proving should be done opportunistically, not at broadcast time. At the time of broadcast, the wallet should not do any syncing or proving - it should *exclusively* broadcast the transaction.

nuttycom: Agreed, and fixed in #1853 (based on this branch).

The "natural anchor" was just this layer's unexplained name for the anchor upstream requires the caller to supply for a preparation (`engine::prove_preparation(prover, state, id, anchor)`; preparations carry `anchor_boundary: None` by upstream design). It is now the *preparation anchor*, and the module doc records why the scanned tip is the right anchor for one: a preparation is a shielded self-send within the source pool that crosses no pool and reveals no balance, so it needs no bucket — bucketing is what pool-crossing transfers need, and only they carry a drawn boundary.

The substantive half is the proving/broadcast split. Proving was only ever initiated from the delivery lane, so the transaction that first came due paid full proving latency on the broadcast path, and a transiently unprovable row aborted the drive and reported "nothing due" — indistinguishable from having no work. Now:

- `zcashlc_migration_prove_pending` proves everything currently provable, persisting each proof, and returns the count. A row whose anchor is not scanned/retained yet is skipped rather than fatal, so the rows behind it still prove and a later sweep retries it.
- `zcashlc_migration_next_due_transfer` never proves. It reports `MigrationAwaitingProof` (naming the waiting row) when something is due without a proof, so that stays distinct from `MigrationNothingDue`.
- The note-split submission keeps its inline first-preparation prove — that lane exists to hand back something to broadcast — but routes through the same `prove_one` helper.

Calling the sweep from the sync path is left to the Synchronizer-level migration API, since `migrationNextDueTransfer` has no production caller yet.

One thing worth noting for the record: anchor availability was never the risk here. `AnchorRetention::retains` keeps every boundary at or above `from_height` with no upper bound, so proving late never loses a transfer's anchor — this was latency and failure-reporting, not correctness.

---

This comment was written with Claude Code assistance.

rust/src/migration.rs
dannywillems: 😄 - I don't know why I commented this...

PR #1818, opened 2026-07-22
Add support for slipstream

Chlup:


docs/SYNC_PERFORMANCE_PROPOSAL.md
nuttycom, line 1: nit, this shouldn't be committed.

rust/src/migration.rs
```diff
@@ -3178,6 +3191,133 @@ pub extern "C" fn zcashlc_ironwood_activation_height(network_id: u32) -> i64 {
...
+            .map_err(|e| anyhow!("Error reading migration row: {e}"))?;
+        let Some(migration_id) = migration_id else {
+            return Ok(0i64);
+        };
+
+        let tx_ids: Vec<i64> = {
+            let mut stmt = ctx
+                .store_conn
+                .prepare(
+                    "SELECT tx_id FROM orchard_ironwood_migration_transactions \
```
nuttycom, line 3279: FIXME: we should not be doing direct database access here, this is fragile.

nuttycom:
Merging. Eight items are known-open against this work and each needs follow-up on `feature/ironwood-slipstream` — none of them blocks landing the branch, but two are memory-safety and one means a shipped feature is currently inert.

**Memory safety**
- #1886 — `start`/`stop` call the FFI off-actor with a captured handle, so `close`/`reopen` (via `wipe()` or `switchTo`) can free it mid-call. Commented with two scope additions: the serialization must cover the `&`-taking read entry points too (`zcashlc_slipstream_stop` derives `&mut` and can alias them), and `deinit` frees on the same race.

**Correctness**
- #1890 — the reconciliation gate queries `slipstream_v_tx_reconciled`, but the engine creates `ext_slipstream_v_tx_reconciled` and drops the unprefixed legacy name. The error is swallowed, so the phantom-receive suppression never engages; the unit test stubs the wrong name, so the suite stays green.
- #1887 — `rescanFrom(height:)` truncates without the stop → truncate → restart sequence `rewindImpl` documents as necessary.
- #1888 — a failed `switchTo` leaves the handle closed, the engine's endpoint updated and `currentEndpoint` stale, and the same-endpoint fast path then no-ops the recovery attempt.
- #1889 — `wipeImpl` never clears `isRunning`, so post-wipe lifecycle calls can spuriously restart sync.

**Ironwood coverage**
- #1891 — the collapsed recovery balance is reported as Orchard value (false "Migration Required" during a post-NU6.3 restore) and `Output.Pool` lacks `.ironwood`. Both are fixed on `kris/merge-maint-v2.8.x` (#1884); the issue carries the open question about which pool the collapse should name.
- zodl-inc/slipstream#9 — the engine's reconciliation view has no `spend_pool = 4` arm, so after activation it degrades toward "always reconciled". This caps what #1890's fix can deliver.

**Hygiene**
- #1892 — the alternate-server rejection check is an `assert`, so it crashes in debug and silently discards the update in release.

Not deleting `michal/slipstream-support` on merge: it is the base of #1884.

---
Posted with Claude Code assistance.

PR #1836, opened 2026-07-24
Select the ZIP 318 anchor bucket interval per network

### LINKED ISSUE #1835: Use a 12-block anchor bucket interval on test networks
The ZIP 318 pool-migration anchor bucketing interval is 144 blocks, roughly three hours at the target block spacing. Exercising a migration end to end against a test network therefore means waiting out multiple buckets per transfer, which makes testnet testing impractical.

librustzcash is making the interval configurable (zcash/librustzcash#2759). This SDK should select it per network: the ZIP 318 default on mainnet, and a 12-block interval with delay distributions derived from it on test networks.

The interval must be applied wherever the SDK opens a `WalletDb`, not just on the scanning path: it governs both which note commitment tree checkpoints are retained and which boundaries a migration may anchor to, and those must agree.

---

nuttycom:
Closes #1835. Resolves MOB-1523

Depends on zcash/librustzcash#2759, which makes the ZIP 318 anchor bucketing interval configurable. The `[patch.crates-io]` block is pinned to that PR's tip, `cd2b9a6a27dd784df3d245cd047bfeade21d9430`.

## What this does

`anchor_retention_interval(NetworkParams)` selects the grid: `AnchorRetentionInterval::ZIP_318` on mainnet, a 12-block interval on testnet. `wallet_db()` applies it, and since that helper builds every wallet handle the FFI hands out, the retention grid and the migration grid cannot be configured apart. `voting/helpers.rs`'s `open_wallet_db` applies the same interval so the voting path retains the same boundaries.

No Swift changes. `network_id` is already plumbed across the FFI for consensus parameters, so the interval is derived on the Rust side; plumbing a second value from Swift would be a source of truth that could drift from the grid the wallet actually retains.

**Worth a reviewer's opinion:** custom-parameter networks (regtest, and the modified-mainnet backends used for Ironwood testing) get the 12-block grid, even though `Custom { .. }` can carry `base: NetworkType::Main`. The reasoning is that a test deployment is not the production anonymity set. Say if you would rather custom networks inherit mainnet's grid.

## Why `scheduling_params` had to be implemented here

`MigrationBackend::scheduling_params` is a new required trait method upstream. This SDK does not use `zcash_pool_migration::wallet::WalletMigration` — `migration_engine.rs` has its own adapter, because `WalletMigration` requires a `UnifiedSpendingKey` that a UFVK-imported Keystone account does not have. The implementation mirrors `WalletMigration`'s: it derives the delay distributions from the wallet's own retention interval, so the grid remains the single configured value.

## Dependency changes

Bumping the pin pulls in a sizeable delta, so this also carries:

- `zcash_pool_migration_backend` renamed to `zcash_pool_migration` (now published as `0.1.0-alpha.1`). It was this repo's one remaining direct git dependency; it is now a version requirement redirected by the patch block, so every librustzcash crate goes through `[patch.crates-io]`.
- `zcash_proofs` added to the patch block; requirements raised to admit the patched versions — `zcash_primitives`/`zcash_proofs` 0.29→0.30, `zcash_transparent` 0.9→0.10, `pczt` 0.8.0-rc.1→0.8
- `MigrationState::from_parts` gained an `AnchorBucketInterval` argument; production sites pass `state.anchor_bucket_interval()`, so a rebuild or cancel keeps the grid its run was committed under
- `scheduling::schedule` takes `&SchedulingParams`; `BundleType` became `BundlePadding` in `BuildConfig::Standard` and `create_pczt_from_proposal`

## The `zcash_voting` pin moved too

The previous fork rev required `zcash_primitives = "0.29.0"`, which the patched librustzcash rev supplies as 0.30.0, so cargo resolved two `zcash_primitives` copies and `zcash_voting` failed to compile — a `PcztParts<Network>` from 0.29 handed to a 0.30-based `pczt`. The pin now points at `zodl-inc/zcash_voting` `e5d885c2`, which is on the 0.30 generation, so the patch reaches it and the graph stays single-copy.

Keep that pin in step with the patch block: a rev on the 0.29 generation cannot accept the patched librustzcash and splits the graph again.

## Verification

On the tree as committed (no local modifications):

- `cargo fmt --check`: clean
- `cargo check --all-targets`: no errors (4 pre-existing "never used" warnings in `voting/`, present at baseline)
- `cargo test`: `72 passed; 0 failed; 0 ignored` (baseline 70; the 2 new ones cover the network-to-interval mapping)
- `cargo tree -d`: no duplicated zcash crates

**Not verified: anything Swift.** There is no Swift/Xcode toolchain on the machine this was prepared on, so `swift test --filter OfflineTests` and the XCFramework build were not run. No `extern "C"` signatures changed, so the cbindgen-generated header should be unchanged, but that was not confirmed by running the generator.

## Note for reviewers

An in-flight **testnet** migration committed under the 144-block grid will now fail proving with `ProveError::AnchorIntervalMismatch` and must be restarted. The upstream change records the grid a migration was committed under and refuses to prove against a different one, rather than letting it fail later as an unexplained missing checkpoint. Mainnet is unaffected.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2774,11 +2774,11 @@ dependencies = [
  "zcash_keys",
  "zcash_note_encryption",
  "zcash_pool_migration",
- "zcash_primitives 0.30.0",
+ "zcash_primitives",
  "zcash_proofs",
  "zcash_protocol",
  "zcash_script",
- "zcash_transparent 0.10.0",
+ "zcash_transparent",
  "zcash_voting",
  "zeroize",
  "zip32",
@@ -3369,10 +3369,10 @@ dependencies = [
  "serde",
  "serde_with",
  "zcash_note_encryption",
- "zcash_primitives 0.30.0",
+ "zcash_primitives",
  "zcash_protocol",
  "zcash_script",
- "zcash_transparent 0.10.0",
+ "zcash_transparent",
 ]
 
 [[package]]
@@ -6656,7 +6656,7 @@ checksum = "6a02e4885ed3bc0f2de90ea6dd45ebcbb66dacffe03547fadbb0eeae2770887d"
 [[package]]
 name = "vote-commitment-tree"
 version = "0.3.2"
-source = "git+https://github.com/zodl-inc/zcash_voting.git?rev=a4daaf77f793b35a98a3d811b920a01b95fbfa7a#a4daaf77f793b35a98a3d811b920a01b95fbfa7a"
+source = "git+https://github.com/zodl-inc/zcash_voting.git?rev=e5d885c2128f59ba39f44631dee537a0ade2f3fd#e5d885c2128f59ba39f44631dee537a0ade2f3fd"
 dependencies = [
  "anyhow",
  "ff",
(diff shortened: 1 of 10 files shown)


nuttycom:
Draft until https://github.com/zcash/librustzcash/pull/2759 merges

nuttycom (commented):
self-utACK

PR #1837, opened 2026-07-25
Retain the wallet's configured anchor grid on the Slipstream sync path

### LINKED ISSUE #1835: Use a 12-block anchor bucket interval on test networks

nuttycom:
Closes #1835. Depends on zcash/librustzcash#2759 and zodl-inc/slipstream#7. Resolves MOB-1523

Based on #1840 (the CI credential fix), so that the Rust-compiling jobs can resolve the private `slipstream-core` dependency and actually exercise this change. Retarget to `michal/slipstream-support` once that merges.

## What this does

The Slipstream engine implements its own tree-update path rather than calling `ll::wallet::put_blocks`, and until recently retained checkpoints on a grid hard-coded in its `persist.rs`. So the interval this crate configures on the `WalletDb` was **silently ignored whenever sync ran through Slipstream**: on a test network the wallet would retain a 12-block grid while the engine retained a 144-block one, and a transfer anchored to a boundary the engine dropped would have become unprovable.

This bumps the `slipstream-core` pin to a rev that takes its grid from `zcash_client_backend` instead of mirroring it (zodl-inc/slipstream#7), and passes the policy through `EngineConfig::anchor_retention` — floor and grid together, the grid being the one `anchor_retention_interval` already selects for the wallet. The two now agree by construction rather than by coincidence.

The pin rides `[patch.crates-io]` like every other override in this file, pinned to a **commit hash rather than a branch**: a branch pin silently follows a force-push, so a build can pick up code nobody reviewed, whereas a stale hash fails loudly. The previous `michal/anchor-retention-interval-144` branch is superseded — it corrected the mirrored constant rather than removing it.

Also drops the `zcash_encoding` and `orchard` patch entries, both dead at this rev and warning on every build.

## Verification

On the branch tip, as pushed:

- `cargo check --all-targets`: no errors
- `cargo fmt --check`: clean
- `cargo test`: `140 passed; 0 failed; 0 ignored`
- `cargo tree -d`: no duplicated zcash crates

**Not verified: anything Swift.** No Swift/Xcode toolchain on the machine this was prepared on, so `swift test` and the XCFramework build were not run. No `extern "C"` signatures changed, so the cbindgen-generated header should be unchanged, but that was not confirmed by running the generator. With #1840 in place, CI is now the first real check of both.

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,12 +113,13 @@ incrementalmerkletree = { version = "0.8", default-features = false }
 # Slipstream engine — the alternative sync engine the SlipstreamSynchronizer
 # drives. Consumed as a remote crate from zodl-inc/slipstream, the engine's
 # home since the personal-fork era (the old PR #6 adaptations are merged
-# there). Tracks the anchor-retention branch (upstream-aligned 144-block
-# anchor retention, typed multi-account sync_once, EngineConfig-owned
-# endpoint probe-then-commit, report.rs presentation split); Cargo.lock pins
-# the exact commit. Return to a release tag once one is cut with these
-# aboard. The engine rides the same librustzcash rev as this package, so the
-# graph stays a single family instance.
+# there). Return to a release tag once one is cut with these aboard.
+#
+# The rev rides [patch.crates-io] below, pinned to a COMMIT HASH rather than a
+# branch: a branch pin silently follows a force-push, so a build can pick up
+# code nobody reviewed, whereas a stale hash fails loudly. The engine rides the
+# same librustzcash rev as this package, so the graph stays a single family
+# instance.
 slipstream-core = "0.6"
 tokio = { version = "1", features = ["rt-multi-thread", "macros", "sync", "time"] }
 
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -113,12 +113,13 @@ incrementalmerkletree = { version = "0.8", default-features = false }
...
 # home since the personal-fork era (the old PR #6 adaptations are merged
-# there). Tracks the anchor-retention branch (upstream-aligned 144-block
-# anchor retention, typed multi-account sync_once, EngineConfig-owned
-# endpoint probe-then-commit, report.rs presentation split); Cargo.lock pins
-# the exact commit. Return to a release tag once one is cut with these
-# aboard. The engine rides the same librustzcash rev as this package, so the
-# graph stays a single family instance.
+# there). Return to a release tag once one is cut with these aboard.
+#
+# The rev rides [patch.crates-io] below, pinned to a COMMIT HASH rather than a
```
dannywillems, line 118: Useless comments I would say.

rust/src/ffi.rs
```diff
@@ -550,7 +549,9 @@ impl WalletSummary {
         if self.account_balances.is_null() {
             &mut []
         } else {
-            unsafe { std::slice::from_raw_parts_mut(self.account_balances, self.account_balances_len) }
```
dannywillems, line 553: Weird to have these git diff from rustfmt.

rust/src/lib.rs
```diff
@@ -194,10 +194,8 @@ unsafe fn wallet_db(
         .map_err(|e| anyhow!("Error setting wallet database busy_timeout: {}", e))?;
     rusqlite::vtab::array::load_module(&conn)
         .map_err(|e| anyhow!("Error loading wallet database array module: {}", e))?;
-    Ok(
```
dannywillems, line 197: No change, simply rustfmt.

dannywillems (approved):
utACK: only adding a new field in cfg, unused for now IIUC. I'll read about slipstream later.

PR #1843, opened 2026-07-25
[#1806] Orchard-to-Ironwood pool migration proposal (send max)

dannywillems:
Adds `Synchronizer.proposeOrchardToIronwoodMigration(accountUUID:)` (plus the
closure and Combine facades): a proposal that migrates the account's entire
Orchard balance into the Ironwood pool before the Orchard turnstile closes at
NU6.3. Mirrors the Android SDK's feature/send-max-2.5.x (PR #2030).

Stacked on the Ironwood support branch (`feature/ironwood-support-2.5.3`).

## What it does

Spends every Orchard note and sends the maximum to the account's own internal
Orchard receiver, with the fee computed so no change returns to Orchard.
Sapling and transparent funds are untouched. Fails unless NU6.3 is active at
the chain tip (before activation the same call would build an Orchard output, a
fee-costing self-send that migrates nothing).

## Reuse

The generic capability is upstream and is not reimplemented:
`propose_send_max_transfer` already spends the whole balance of a chosen set of
pools with the fee computed to leave nothing over. `rust/src/migration.rs` only
pins the parameters (Orchard-only, `MaxSpendMode::Everything`, the internal
receiver) that make that call the migration, exposed through the new
`zcashlc_propose_orchard_to_ironwood_migration` FFI. There is no general
send-max in the SDK's API.

Gates: `cargo build` green, `swift build` green, OfflineTests 325/0.

### DIFF
--- a/rust/src/migration.rs
+++ b/rust/src/migration.rs
@@ -1,0 +1,95 @@
+//! Orchard to Ironwood pool migration.
+//!
+//! Builds on the upstream `propose_send_max_transfer`, which spends the entire
+//! spendable balance of a chosen set of shielded pools to one recipient with
+//! the fee computed so nothing is left over. This module only pins the
+//! parameters that make that call the pool migration; it is not a general
+//! send-max and none is exposed.
+
+use anyhow::anyhow;
+use rand::rngs::OsRng;
+use zcash_address::{ToAddress, ZcashAddress, unified, unified::Encoding as _};
+use zcash_client_backend::{
+    data_api::{
+        Account as _, InputSource, MaxSpendMode, WalletRead,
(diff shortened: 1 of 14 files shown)


rust/src/migration.rs
```diff
@@ -0,0 +1,95 @@
+//! Orchard to Ironwood pool migration.
```
dannywillems, line 1: This follows the code from Android. As a reviewer, you should compare to https://github.com/zcash/zcash-android-wallet-sdk/pull/2030

Chlup (dismissed):

Chlup (approved):

PR #1853, opened 2026-07-26
[#MOB-1455] Prove migration transactions opportunistically, not at broadcast time

nuttycom:
Follow-up to the review discussion on #1813 (https://github.com/zcash/zcash-swift-wallet-sdk/pull/1813#discussion_r3651582760). Targets that branch so it lands with the FFI it corrects. Part of #1806.

## The problem

A migration transaction's anchor becomes witnessable long before its broadcast schedule arrives, and upstream separates `next_provable` from `next_broadcastable` precisely so proving can happen in that window ("Proving is decoupled from broadcasting so a transfer is proved while its anchor boundary checkpoint is still within the wallet's pruning window, then broadcast later at its scheduled height").

This layer collapsed the two. Proving was only ever initiated from the delivery lane, so:

- the transaction that first came due paid full Orchard+Ironwood proving latency **on the broadcast path**, on mobile, against a transaction with a ZIP 203 expiry;
- a transiently unprovable row aborted the whole drive and reported "nothing due" — indistinguishable from having no work at all, and blocking the rows behind it.

Anchor availability was never at risk (`AnchorRetention::retains` keeps every boundary at or above `from_height` with no upper bound), so this is a latency/robustness problem, not a correctness one.

## The change

- **`zcashlc_migration_prove_pending`** (new) proves every currently-provable transaction, persisting each, and returns the count proved (`-1` = error). A row whose anchor the wallet has not scanned or retained yet is **skipped, not fatal** — the rows behind it still prove, and a later sweep retries it. The skip asks a scratch clone for the next provable id rather than re-deriving upstream's prove-readiness rules here.
- **`zcashlc_migration_next_due_transfer` never proves.** `FfiPreparedTransfer` gains a `status` (`MigrationNothingDue` / `MigrationReady` / `MigrationAwaitingProof`), so "nothing is due" stays distinct from "due, but its proof has not been produced yet" — the latter names the waiting row, which a sweep clears. The variants carry their own `Migration` prefix because cbindgen emits C enum variants unqualified unless two enums collide, and a bare `Ready` would land in a header that ships to consuming apps.
- `prove_if_needed` splits into `prove_one` (`Signed -> Proved`, persisted) and `serve_proved` (a pure read of the stored artifact). `zcashlc_migration_sign_note_split` keeps its inline first-preparation prove — that lane exists to hand back something to broadcast — but routes through `prove_one`, so there is one proving code path.
- The "natural anchor" is renamed to the preparation anchor, with the rationale documented: a preparation is a shielded self-send within the source pool that crosses no pool and reveals no balance, so anchoring it at the scanned tip leaks nothing. Bucketing is what pool-crossing transfers need, and only they carry a drawn boundary.

Swift gains `migrationProvePending(for:)` and a `DueMigrationTransfer` return for the delivery lane (plus `ZRUST0135`). **No sync-pipeline wiring here** — `migrationNextDueTransfer` has no production caller yet, so the "call the sweep as blocks are scanned" step belongs with the Synchronizer-level migration API.

## Testing

- `cargo test --lib`: 74/74. Five new/rewritten migration tests — sweep-then-serve-without-consulting-a-prover, `AwaitingProof` reporting for a due unproven row, skip-transient-and-keep-sweeping, prove-ahead-of-schedule, hard-failure propagation.
- `swift test --filter OfflineTests`: 531/531, against a locally built FFI (the released XCFramework has no `zcashlc_migration_prove_pending`).
- `cargo fmt` clean; no new clippy warnings. SwiftLint was not available in this environment.

## Notes for #1812's rebase

#1812 rebases onto `michal/MOB-1455/MOB-1495-sdk-pool-migration-ffi` after this lands. Two call-site changes it needs to absorb:

1. **`migrationNextDueTransfer` no longer returns an Optional.** It returns `DueMigrationTransfer` -- `.nothingDue`, `.ready(PreparedMigrationTransfer)`, or `.awaitingProof(id:)`. #1812's `guard let prepared = try await welding.migrationNextDueTransfer(for: accountUUID) else { ... }` becomes a match on `.ready`, with `.nothingDue` taking the old `else` path and `.awaitingProof` meaning "a transfer is due but unproven" -- run the proving sweep, then retry, rather than treating it as nothing to do.

2. **Proving belongs in the sync path.** `migrationProvePending(for:)` proves everything currently provable and returns the count; it is meant to be called as blocks are scanned, not before broadcasting. #1812's Synchronizer surface is the natural caller -- without it, the delivery lane will report `.awaitingProof` indefinitely.

---

This PR was prepared with Claude Code assistance.


rust/src/migration.rs
```diff
@@ -680,132 +681,206 @@ fn map_rebuild_err(e: engine::RebuildError<anyhow::Error>) -> anyhow::Error {
...
-/// transfer against the boundary anchor persisted on its row, a preparation against the wallet's
-/// natural anchor. The engine persists the proven bytes through its own `Proved` state. Returns
-/// the broadcastable `(proven pczt bytes, txid)` — or `None` when the wallet has not
-/// scanned/retained the needed anchor yet (a restored wallet mid-sync, a boundary not yet
-/// checkpointed), the ordinary transient state the caller maps to "nothing due".
-fn prove_if_needed(
+/// Serves an already-`Proved` row's stored artifact as `(proven pczt bytes, txid)`. A pure read:
+/// the bytes were persisted when the row was proved, so a retry after a failed broadcast attempt
+/// re-serves exactly the same transaction (same anchor, same txid) rather than re-proving.
+fn serve_proved(state: &MigrationState, id: MigrationTxId) -> anyhow::Result<(Vec<u8>, [u8; 32])> {
```
dannywillems: LGTM. Nit: the return types could be pczt and txid types instead. We should later add in AGENTS the same as we have in https://github.com/zcash/librustzcash/blob/main/AGENTS.md

rust/src/migration.rs
```diff
@@ -680,132 +681,206 @@ fn map_rebuild_err(e: engine::RebuildError<anyhow::Error>) -> anyhow::Error {
...
             {
-                // Not scanned/retained yet — transient, retry on a later call.
-                return Ok(None);
+                // Not scanned/retained yet — transient, retry on a later sweep.
+                return Ok(false);
             }
             let mut backend = Backend::new(&ctx.wallet, ctx.account, None, &mut ctx.store_conn)?;
             backend.replace_migration(state)?;
-            // Re-read the engine-stored proven bytes and extract the txid to serve alongside.
-            let tx = state
```
dannywillems, line 837: LGTM. This is the code extracted above.

rust/src/migration.rs
```diff
@@ -680,132 +681,206 @@ fn map_rebuild_err(e: engine::RebuildError<anyhow::Error>) -> anyhow::Error {
...
-/// `next_provable`'s contract), so the drive may prove a not-yet-due transfer on the way to
-/// serving a due one behind it. Aside from `zcashlc_migration_sign_note_split`'s explicit
-/// first-preparation prove, this drive is the only place proving is initiated.
+/// Prove-readiness is upstream's rule, not this crate's, so the skip is expressed by asking a
+/// SCRATCH clone: the skipped rows are flipped to `Proved` in the clone (with their own bytes —
+/// `set_transaction_proved` then only advances the lifecycle state), which is exactly what makes
+/// `next_provable` step past them. Nothing persists, and a skipped row cannot unblock a dependent
+/// one, because dependency readiness is keyed on `Mined`, not `Proved`. The same trick backs
+/// [`due_assuming_proving`].
+fn next_provable_excluding(
```
dannywillems, line 882: Nit: the doc should explain that it is the next provable tx (using the txid) that is returned. It eases the review for the context.

rust/src/migration.rs
```diff
@@ -680,132 +681,206 @@ fn map_rebuild_err(e: engine::RebuildError<anyhow::Error>) -> anyhow::Error {
...
+            .iter()
+            .find(|t| t.id() == *id)
+            .map(|t| t.pczt().clone())
+            .unwrap_or_default();
+        scratch.set_transaction_proved(*id, bytes);
+    }
+    scratch.next_provable(tip)
+}
+
+/// Proves every row that can be proved right now, persisting each, and returns how many were
```
dannywillems, line 903: Follow-up: `row` is used instead of `transaction`. This should be rephrased.

rust/src/migration.rs
```diff
@@ -680,132 +681,206 @@ fn map_rebuild_err(e: engine::RebuildError<anyhow::Error>) -> anyhow::Error {
...
-        MigrationTxId,
-    ) -> anyhow::Result<Option<(Vec<u8>, [u8; 32])>>,
-) -> anyhow::Result<Option<(MigrationTxId, [u8; 32], Vec<u8>)>> {
-    while state.next_broadcastable(tip).is_none() {
-        let Some(provable) = state.next_provable(tip) else {
-            return Ok(None);
-        };
-        if prove(state, provable)?.is_none() {
-            return Ok(None);
+    mut prove: impl FnMut(&mut MigrationState, MigrationTxId) -> anyhow::Result<bool>,
```
dannywillems: Nit: there is a trait for the prover in the engine. Maybe we can mix with it. For a follow-up it is fine, to improve the codebase for maintenance by humans.

rust/src/migration.rs
```diff
@@ -999,9 +1074,30 @@ pub struct FfiNoteSplitProposal {
     pub proposal_handle: u64,
 }
 
+/// What a [`FfiPreparedTransfer`] carries, so the delivery lane's three outcomes stay distinct
```
dannywillems, line 1294: >  so the delivery lane's three outcomes stay distinct instead of collapsing into one all-null sentinel.

🤔 - interesting AI prose.

rust/src/migration.rs
```diff
@@ -680,132 +681,206 @@ fn map_rebuild_err(e: engine::RebuildError<anyhow::Error>) -> anyhow::Error {
...
-    ) -> anyhow::Result<Option<(Vec<u8>, [u8; 32])>>,
-) -> anyhow::Result<Option<(MigrationTxId, [u8; 32], Vec<u8>)>> {
-    while state.next_broadcastable(tip).is_none() {
-        let Some(provable) = state.next_provable(tip) else {
-            return Ok(None);
-        };
-        if prove(state, provable)?.is_none() {
-            return Ok(None);
+    mut prove: impl FnMut(&mut MigrationState, MigrationTxId) -> anyhow::Result<bool>,
+) -> anyhow::Result<u32> {
```
dannywillems, line 919: The return type is u32 here, but a i64 is used by a call in the FFI.

rust/src/migration.rs
```diff
@@ -2021,14 +2138,49 @@ pub unsafe extern "C" fn zcashlc_migration_sign_and_store_schedule(
...
+        let Some(mut state) = reconcile_mined(&mut ctx)? else {
+            return Ok(0);
+        };
+        if state.is_terminal() {
+            return Ok(0);
+        }
+        let tip = ctx.tip()?;
+        let proved =
+            prove_pending_rows(&mut state, tip, |state, id| prove_one(&mut ctx, state, id))?;
+        Ok(i64::from(proved))
```
dannywillems, line 2577: `proved` is a u32 in the callee. The FFI uses a i64 conversion. This is sound as the call is properly passed with `from`, but u32 might be used directly.

rust/src/migration.rs
```diff
@@ -4183,19 +4380,18 @@ mod tests {
...
+    /// account-keyed store. The preparation anchor is never resolved (these fixtures sweep
+    /// transfers only, which prove against their persisted boundary).
     fn prove_with_test_prover<P>(
         path: &std::path::Path,
         account: &[u8; 16],
         prover: &mut P,
         state: &mut MigrationState,
         id: MigrationTxId,
-    ) -> anyhow::Result<Option<(Vec<u8>, [u8; 32])>>
+    ) -> anyhow::Result<bool>
```
dannywillems, line 6297: Type changed mostly because of the refacto.

rust/src/migration.rs
```diff
@@ -680,132 +681,206 @@ fn map_rebuild_err(e: engine::RebuildError<anyhow::Error>) -> anyhow::Error {
...
-    let id = state
-        .next_broadcastable(tip)
-        .expect("the drive loop exits with a broadcastable row");
-    Ok(prove(state, id)?.map(|(proven, txid)| (id, txid, proven)))
+    Ok(proved)
+}
+
+/// What the delivery lane has for the platform at `tip`.
+#[derive(Debug, PartialEq, Eq)]
+enum DueOutcome {
```
dannywillems, line 948: Nit: there are already states for this in the engine in librustzcash. Methods already exist for these.

rust/src/migration.rs
```diff
@@ -680,132 +681,206 @@ fn map_rebuild_err(e: engine::RebuildError<anyhow::Error>) -> anyhow::Error {
...
+    /// Nothing is due.
+    Nothing,
+    /// This row is proved and due: its stored artifact is what gets broadcast.
```
dannywillems, line 951: I think redirecting the AI by rephrasing `row` by `transaction` or `migration transfer` or `migration state` would help reaching the related context of this review note https://github.com/zcash/zcash-swift-wallet-sdk/pull/1853/changes#r3652285066.

rust/src/migration_finalize.rs
```diff
@@ -92,12 +102,12 @@ impl<TE, NE, RE> ProveErrorClass for WalletProveError<TE, NE, RE> {
...
 /// kind: a TRANSFER via [`engine::prove_transfer`] (the anchor comes from the transfer's
 /// persisted `anchor_boundary()`; a missing boundary is a corrupt store and a hard error), a
-/// PREPARATION via [`engine::prove_preparation`] with the caller-resolved `natural_anchor`. On
+/// PREPARATION via [`engine::prove_preparation`] with the caller-resolved `preparation_anchor`. On
 /// success the engine has stored the proven bytes into `state` (`Signed -> Proved` via
 /// `set_transaction_proved`); the caller persists `state` and re-reads the proven PCZT from it.
 ///
-/// `natural_anchor` is `Option`al because only a PREPARATION uses it — the caller resolves it
-/// lazily, per kind, so proving a transfer never depends on the natural anchor being resolvable
+/// `preparation_anchor` is `Option`al because only a PREPARATION uses it — the caller resolves it
```
dannywillems, line 109: This is only a renaming.

rust/src/migration_finalize.rs
```diff
@@ -17,15 +21,21 @@
...
 //!   retention (active from NU6.3 activation) keeps every boundary of the wallet's anchor
 //!   retention interval — necessarily the same grid the engine draws boundaries on, since the
 //!   engine reads that interval back off the wallet. [`crate::anchor_retention_interval`] is where
 //!   the SDK selects it per network.
-//! - PREPARATIONS carry no drawn boundary (they anchor to their already-mined dependencies, not
-//!   to a bucketed boundary) and prove against the wallet's current natural anchor
-//!   ([`natural_anchor_height`]) via `engine::prove_preparation`.
+//! - PREPARATIONS carry no drawn boundary and prove against the wallet's scanned tip
+//!   ([`preparation_anchor_height`]) via `engine::prove_preparation`, which is why that anchor is
+//!   supplied by this caller rather than read off the row. Anchoring at the tip is correct for
```
dannywillems, line 32: > supplied by this caller rather than read off the row.

Rephrasing as mentioned before would be nice for the context.

dannywillems (approved):
utACK. Left a few nitpick comments. I would like to see how it is actually plugged in the app, and how the UI handles this; this is the missing context.
The code could be simplified by using the existing logics in the migration code in librustzcash.

PR #1854, opened 2026-07-26
Merge the 2.7.x and 2.8.0-rc.1 release lines back to `main`, updating shielded voting for Ironwood

### LINKED ISSUE #1855: Reintroduce the shielded voting module on the Ironwood (NU6.3) stack
The `maint/v2.7.x` → `main` merge removed the Swift shielded-voting surface
(`VotingRustBackend`, `VotingTypes`, `VotingConstants`, `PirSnapshotResolver`
and their tests) and gated the Rust `rust/src/voting/` module behind
`#[cfg(zcash_voting)]`, a cfg that is never set. This issue tracks bringing
voting back on the Ironwood dependency stack.

## Background

The gating comment in `Cargo.toml` and `rust/src/lib.rs` states that
`zcash_voting` cannot resolve against `orchard 0.15`. That is accurate only for
the crates.io release: `zcash_voting 1.0.0` pins `orchard ^0.14`,
`zcash_client_backend ^0.23`, `zcash_keys ^0.14` and `zcash_protocol ^0.9` —
the pre-Ironwood set. The `[patch.crates-io]` git revision already present
(commented) in `Cargo.toml` resolves cleanly, adding 205 lines to `Cargo.lock`
and changing no existing crate version.

Compilation is a separate matter. With the module ungated, `cargo check` fails
with 9 errors across 7 files. These are not incidental drift: upstream's
`README.md` §"Migrating 0.11 to 0.12" documents a restructuring in which the
crate absorbed orchestration that `rust/src/voting/` currently hand-rolls.
There is no revision offering both Ironwood support and the older API shape —
Ironwood support exists only above the `v1.0.0` changelog heading.

## Scope

- Uncomment the four voting dependencies and the mandatory `[patch.crates-io]`
  pin; ungate `mod voting;`. No Cargo feature gate is introduced: the Rust
  library is an implementation detail of the Swift package, and `Package.swift`
  (tools-version 5.6) cannot gate voting for consumers, so a Rust-only flag
  would gate nothing reachable while leaving a footgun where a bare
  `cargo build` yields a silently non-linkable library.
- Port `rust/src/voting/` per the upstream migration guide, adopting
  `vote::commit`, `share::*`, `session::resume_plan`,
  `delegate::PreparedDelegationBundle`, `confirmation::*` and the `VotingDb`
  phase APIs, deleting the local orchestration they supersede.
- Restore the five deleted Swift files and repair them against the resulting
  FFI, adopting the `Undescribable` marker on secret-carrying types.
- Document the breaking changes in `MIGRATING.md` and `CHANGELOG.md`.

## Behaviour changes for consumers

- **Hotkeys are no longer derived from the wallet seed.** Upstream replaced
  `hotkey::generate_hotkey(seed)` with `generate_random_voting_hotkey(network)`;
  wallets generate a random hotkey once and persist
  `VotingHotkey::stored_secret()`. Existing seed-derived hotkeys do not carry
  over, and restore-from-seed cannot recover voting ability.
- `VotingRustBackend.decomposeWeight` is removed — no upstream equivalent
  exists, and share construction is now fully delegated to the crate.
- The proof-progress callback signature changes, as
  `DelegationProgressReporter::on_progress` takes a structured
  `DelegationProgress` rather than an `f64`.

## Follow-ups tracked separately

- Swift API-rules remediation: several public voting APIs take bare `[UInt8]`
  seeds and a bare `String` UFVK, and `VotingRustBackendError` carries raw FFI
  error strings rather than using generated `ZcashError` cases.
- PIR snapshot probing uses raw `URLSession` on clearnet rather than the SDK's
  Tor client.
- Hotkey backup/restore semantics.

---

nuttycom:
Merges the 2.7.x maintenance line and the 2.8.0-rc.1 release back into `main`, and brings the shielded voting surface onto the Ironwood (NU6.3) dependency stack in the process.

Closes #1855.
Resolves MOB-1545
Resolves MOB-1558
Resolves MOB-1557
Resolves MOB-1556
Resolves MOB-1555

## Why voting had to change

The two framings are both accurate, depending on where you look from:

- Relative to **`main`**, this is an **update**. `main` builds voting against `zcash_voting 0.11` from crates.io, on the pre-Ironwood librustzcash family.
- Relative to the **2.7.x line**, this is a **reinstatement**. Voting was removed on 2.7.0-rc.1 and gated behind `#[cfg(zcash_voting)]`, on the recorded grounds that `zcash_voting` could not resolve against the Ironwood `orchard` release.

That recorded reason turned out to be true only of the crates.io release, which pins the pre-Ironwood family. It does not hold for the git line, so voting is compiled again here rather than withheld.

Voting is compiled **unconditionally**, not behind a Cargo feature. `Package.swift` cannot gate voting for SDK consumers, so a Rust-only feature would gate nothing reachable while leaving a bare `cargo build` producing a library the Swift layer cannot link against.

## Dependency change

`zcash_voting` is taken from git at [`zodl-inc/zcash_voting@bde9ee02`](https://github.com/zodl-inc/zcash_voting/commit/bde9ee0283ff3e23995bd6996dff3ba1f8969076) (version `2.0.0-rc.1`), whose librustzcash pins match this release exactly, including the `zcash_client_sqlite 0.22.0-rc.4` update 2.8.0-rc.1 adopts.

**Reviewers: check the version requirement and the patched revision together.** A `[patch.crates-io]` entry only applies when it satisfies the dependency requirement. Leaving the requirement at `"1.0"` while patching in a `2.0.0-rc.1` revision makes cargo ignore the patch and silently resolve the crates.io `1.0.0` release instead, pulling `orchard 0.14` and `pczt 0.7` back into the graph alongside the 0.15 stack. Cargo reports this only as a warning, and the result still builds. The manifest carries a comment to this effect.

## API changes

The voting crate absorbed orchestration this SDK used to drive step by step, and made the intermediate steps private, so the FFI surface moves from 59 exported functions to 54.

**Removed (8):** `build_vote_commitment`, `sign_cast_vote`, `build_share_payloads` and `encrypt_shares`, all four superseded by the new `commit_vote`; `decompose_weight`, which has no upstream equivalent; `get_delegation_submission` and `get_delegation_submission_with_keystone_sig`, superseded by `get_delegation_submission_with_signature`; and `store_commitment_bundle`, superseded by `record_vc_position`.

**Added (3):** `commit_vote`, `get_delegation_submission_with_signature`, `record_vc_position`.

Six more changed signature, and `FfiVotingHotkey` and `FfiBundleSetupResult` changed shape. Full details are in `MIGRATING.md`.

### The change most likely to affect a shipped wallet

Voting hotkeys are no longer derived from the wallet seed. They are app-owned random values, so **the application must persist the hotkey's stored secret**. It cannot be re-derived: restoring a wallet from its seed phrase does not restore the ability to vote with a hotkey whose secret was lost, and voting power already delegated to that hotkey becomes unusable. Hotkeys created by an earlier SDK version do not carry over.

Types carrying key material or note secrets (`VotingHotkey`, `VotingNoteInfo`, `VotingPczt`, `VotingDelegationKeyInputs`) now conform to `Undescribable`, which the previous surface never did despite `VotingHotkey` existing to hold a secret.

## Blocker before this can ship

`Package.swift` points at the **2.7.0-rc.2 XCFramework, which exports no `zcashlc_voting_*` symbols** — it was built from a tree where voting was still gated off. Consumers using the binary target will fail to link the voting Swift surface. Local FFI builds are unaffected, and that is how everything below was verified.

**A new FFI release must be cut before this merges.**

## Test plan

All verification below was run against a locally built FFI (`./Scripts/rebuild-local-ffi.sh macos`), since the released XCFramework predates the voting reinstatement.

- `cargo check --all-targets`, `cargo clippy --all-targets`, `cargo fmt --check` — clean, no warnings.
- `cargo test --lib` — 73 passed, 0 failed.
- `swift build` — 0 errors.
- `swift test --filter OfflineTests` — 457 passed, 0 failed.
- `./Tests/TestUtils/Sourcery/generateMocks.sh` and `./Sources/ZcashLightClientKit/Error/Sourcery/generateErrorCode.sh` — both reproduce their committed output byte-identically.
- SwiftLint under both `.swiftlint.yml` and `.swiftlint_tests.yml` — 0 errors.

Not covered by automated tests, and worth manual verification on testnet before release:

- A full delegation and cast-vote round against a live PIR endpoint and lightwalletd. The unit tests cover the FFI boundary and rejection paths, but no test builds a real vote proof.
- Hotkey generation, persistence, and reconstruction across an app restart, which is the new storage obligation.
- Recovery after an interrupted `commitVote`, which is idempotent and should return the persisted result rather than rebuilding the proof.

## Known coverage gap

`insert_vote_fixture` is `#[cfg(test)]` upstream and so unreachable from here, which means a vote row can now only be produced by a real `vote::commit` — too expensive for a unit test. One test was removed for this reason with no cheap replacement. Reported upstream as [valargroup/zcash_voting#149](https://github.com/valargroup/zcash_voting/issues/149), which proposes a `test-dependencies` feature. Note that issue is filed against `valargroup`, where the fixture originates; the dependency itself now comes from `zodl-inc`.

## Follow-ups

Deliberately out of scope to keep this reviewable:

- #1876 — voting Swift API does not use semantic wrapper types, and `VotingRustBackendError` carries raw FFI strings rather than using generated `ZcashError` cases.
- #1877 — PIR snapshot probing bypasses Tor, revealing voting participation to a network observer.
- #1878 — vote records expose a derived `submitted` boolean rather than the four-state lifecycle phase.
- #1879 — hotkey backup and `wipe()` semantics are undefined.

## One question for reviewers

The release line added `AGENTS.md` as a pointer to `CLAUDE.md`; this branch had already made `AGENTS.md` the single source with `CLAUDE.md` symlinked to it. The merge keeps the symlink, since one file cannot drift from itself and both names still resolve, and folds in the release's guidance about reporting a plan's full path. The two conventions will collide again on every future merge until one is adopted repo-wide — worth settling here.

---
Prepared with Claude Code assistance.


nuttycom:
Draft until the rust voting API is reenabled.

zcash-harry (approved):
utACK [fb21631](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1854/commits/fb21631ec3a3d292e400b2d46211a7d0250409da)

PR #1860, opened 2026-07-26
[#1858] Redact Ironwood PCZT bundle for external signers and cache the Orchard proving key

dannywillems:
Closes #1858. Resolves MOB-1547

Ports the two items from the Android SDK's PCZT Ironwood work (zcash/zcash-android-wallet-sdk#2066) that were still missing in the Swift Rust core. The Ironwood proof-creation item from that PR was already present here.

## Changes (`rust/src/lib.rs`)

- **Redact the Ironwood bundle** in `zcashlc_redact_pczt_for_signer`. It previously cleared spend witnesses and `output_info` metadata for the Orchard, Sapling, and transparent bundles but left the Ironwood bundle untouched, exposing its Merkle witnesses and wallet output metadata to an external/hardware signer. Now mirrored via `redact_ironwood_with`.
- **Cache the Orchard-family proving key** in `zcashlc_add_proofs_to_pczt`. The Orchard and Ironwood proofs each built a fresh `orchard::circuit::ProvingKey`; both collapse onto the `PostNu6_3` circuit after NU6.3, so they now use `cached_orchard_proving_key` (built once per circuit version), and the Ironwood circuit version is derived from the PCZT's consensus branch id via `bundle_version_for_branch` instead of being hardcoded.

## Notes

- Available on this branch's deps: `pczt 0.8.0` (`redact_ironwood_with`), `zcash_primitives 0.30.0` (`cached_orchard_proving_key`).
- `cargo check` passes; additions are rustfmt-clean.
- CHANGELOG entry is in its own commit.

PR #1861, opened 2026-07-26
[#1859] Include the Ironwood pool in the general send-max spend set

### LINKED ISSUE #1859: Include the Ironwood pool in the general send-max spend-pool set
## Summary

`zcashlc_propose_send_max_transfer` (`rust/src/lib.rs`) passes `&[ShieldedPool::Sapling, ShieldedPool::Orchard]` as the `spend_pools` for `propose_send_max_transfer`, omitting `ShieldedPool::Ironwood`.

Since `propose_send_max_transfer` spends the entire spendable balance across the listed pools, a post-NU6.3 wallet holding Ironwood notes would have those funds silently excluded from a 'send max' / 'send everything' operation, so it would not actually send the maximum.

Including `ShieldedPool::Ironwood` is a no-op when the wallet holds no Ironwood notes (safe pre-activation) and correct after activation.

## Scope note

This is the **general** send-max FFI, not the migration path. The Orchard->Ironwood migration (`migration.rs` `propose_orchard_to_ironwood`) intentionally uses `[ShieldedPool::Orchard]` only and is unaffected.

Tracked separately from zcash/zcash-swift-wallet-sdk#1858 per review.  <br>Resolves zcash/zcash-swift-wallet-sdk#1910.

dannywillems:
Closes #1859. Resolves MOB-1548

`zcashlc_propose_send_max_transfer` passed only `ShieldedPool::Sapling` and `ShieldedPool::Orchard` as the spend pools for `propose_send_max_transfer`, omitting `ShieldedPool::Ironwood`.

Since send-max draws the entire spendable balance across the listed pools, a post-NU6.3 wallet holding Ironwood notes would have those funds silently excluded, so a 'send max' / 'send everything' would not actually send the maximum. This PR adds `ShieldedPool::Ironwood` to the spend set; it is a no-op when the account holds no Ironwood notes (safe pre-activation, correct after).

## Scope

This is the **general** send-max FFI, not the migration path. The Orchard-to-Ironwood migration (`migration.rs` `propose_orchard_to_ironwood`) intentionally spends from the Orchard pool only and is unaffected.

## Notes

- `cargo check` passes; the addition uses a named `spend_pools` binding and is rustfmt-clean.
- CHANGELOG entry is in its own commit.
- Kept separate from #1860 per review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The public `ZcashError.serviceGetTaddressTxidsFailed` case is unchanged
   aside from its message text.
 
+## Fixed
+- Send-max proposals now spend from the Ironwood pool in addition to
+  Sapling and Orchard, so a post-NU6.3 wallet's Ironwood funds are no
+  longer silently excluded from a send-max. This affects only the general
+  send-max proposal; the Orchard-to-Ironwood migration is unchanged.
+
 # 2.7.0-rc.1 - 2026-07-25
 
 ## Added
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -2360,11 +2360,21 @@ pub unsafe extern "C" fn zcashlc_propose_send_max_transfer(
         let locked_input_policy = LockedInputPolicy::Exclude;
         let lock_inputs = None;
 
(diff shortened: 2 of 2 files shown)


dannywillems:
This is the general send max, not the migration related one.

nuttycom (approved):
utACK

PR #1864, opened 2026-07-26
[#1863] Reuse zcash_client_backend redact_pczt_for_signer

### LINKED ISSUE #1863: Reuse zcash_client_backend redact_pczt_for_signer instead of reimplementing PCZT signer redaction
## Summary

`zcashlc_redact_pczt_for_signer` (`rust/src/lib.rs`) hand-rolls a `Redactor` chain that duplicates `zcash_client_backend::data_api::wallet::redact_pczt_for_signer`. The reimplementation has drifted from the upstream canonical redaction: it clears only spend witnesses and `output_info` metadata, and does **not** clear zk-proofs, binding signature keys (`bsk`), dummy spend keys (`spend_dummy_sk` / `dummy_ask`), or v6 anchors, nor does it call `compact_resolvable_fields`.

Follow-up to zcash/zcash-swift-wallet-sdk#1858 (which added Ironwood coverage to the hand-rolled version). The correct fix is to delete the reimplementation and call the upstream function, which already covers every bundle (Orchard, Ironwood, Sapling, transparent) and performs the full redaction.

## Change

* Replace the hand-rolled `Redactor::new(pczt)...finish()` body with `redact_pczt_for_signer(&pczt)`.
* Drop the now-unused `pczt::roles::redactor::Redactor` import.

Keeps the SDK aligned with librustzcash and avoids maintaining a redaction policy in two places.

dannywillems:
Closes #1863. Follow-up to #1860. Resolves MOB-1549

`zcashlc_redact_pczt_for_signer` hand-rolled a `Redactor` chain that duplicated `zcash_client_backend::data_api::wallet::redact_pczt_for_signer`. The reimplementation had drifted from the upstream canonical redaction: it cleared only spend witnesses and `output_info` metadata and, unlike upstream, never cleared zk-proofs, binding signature keys (`bsk`), dummy spend keys (`spend_dummy_sk`/`dummy_ask`), or v6 anchors, nor called `compact_resolvable_fields`.

## Change

- Replace the hand-rolled `Redactor::new(pczt)...finish()` body with `redact_pczt_for_signer(&pczt)`.
- Drop the now-unused `pczt::roles::redactor::Redactor` import.

Keeps the SDK aligned with librustzcash and removes a second copy of the redaction policy. `cargo check` passes; changes are rustfmt-clean; CHANGELOG entry is in its own commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,11 +32,13 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   than hardcoding it. The resulting proofs are unchanged.
 
 ## Fixed
-- Redacting a PCZT for an external signer now also redacts the Ironwood
-  bundle (clearing spend witnesses and `output_info` output metadata),
-  matching the Orchard, Sapling, and transparent bundles; previously the
-  Ironwood bundle was left untouched, exposing its Merkle witnesses and
-  wallet output metadata to the signer.
+- Redacting a PCZT for an external signer now delegates to
+  `zcash_client_backend`'s `redact_pczt_for_signer` instead of a
+  hand-rolled reimplementation. This redacts the Ironwood bundle, which the
+  previous implementation left untouched (exposing its Merkle witnesses and
+  output metadata to the signer), and also clears zk-proofs, binding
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK

PR #1866, opened 2026-07-26
Validate the seed against existing accounts in prepare

Chlup:
Ports the prepare-time seed/account integrity guard and `ZcashError.initializerSeedMismatch` (`ZINIT0006`) from `release/2.6.0` onto the 2.8 line. The 2.6.0 side shipped it inside the Ironwood work; this carries over **only** the seed-validation change.

## Why

`Initializer.initialize` is idempotent for an existing wallet, so restoring a *different* seed over existing accounts previously no-op'd silently. The keychain held seed B while `data.db` kept seed A's account — the app showed A's balance and receive address (funds receivable but unspendable) and sends failed `ZRUST0002`. The wallet stayed in that state until `wipe()`.

`initialize` now validates the caller's seed against the stored seed-derived account(s) before opening, and throws `ZcashError.initializerSeedMismatch` on a genuine mismatch.

## What's in it

- `Error/ZcashErrorCodeDefinition.swift` — new `initializerSeedMismatch` case (`ZINIT0006`). `ZcashError.swift` / `ZcashErrorCode.swift` regenerated with `Error/Sourcery/generateErrorCode.sh` (Sourcery 2.3.0), not hand-edited.
- `Initializer.swift` — `validateSeedAgainstExistingAccounts(_:existingAccounts:)`, called before the account-creation branch; `listAccounts()` is hoisted into a local so it isn't fetched twice.
- `CHANGELOG.md` / `MIGRATING.md` — user-visible behavior change documented.
- `Tests/OfflineTests/WalletTests.swift` — four new tests (see below).

The relevance check is delegated to the Rust core, which reports the seed relevant when it derives an existing account, when there are no accounts, or when there is **no seed-derived account to validate against** — so imported-only (hardware-wallet UFVK) wallets are exempt. This is deliberately *not* a Swift-side heuristic on `seedFingerprint`/`hdAccountIndex`: those are populated for imported-spending accounts too, and reading them was what bricked hardware-wallet-only wallets in the first version on 2.6.0. That fix is folded in here, so this line never carries the intermediate broken version.

No Rust or FFI work: `isSeedRelevantToAnyDerivedAccount` and its whole Swift surface already exist on this branch.

## Deliberately not ported

The rest of the `release/2.6.0` delta is unrelated and stays out: custom-network / `NetworkType.regtest` registration, `network.saplingActivationHeight`, the Ironwood balance and subtree-root surface, and the voting changes. 2.8's existing `ZcashError+LocalizedError.swift` and its `getTaddressTransactions` doc text are left as they are.

## Tests

New in `WalletTests`, following the existing MOB-1512 pattern (`ZcashRustBackendWeldingMock` + `mockContainer`):

| Test | Pins |
|---|---|
| `testInitializeThrowsSeedMismatchWhenSeedIsNotRelevantToDerivedAccounts` | accounts exist + seed not relevant ⇒ throws, and `createAccount` is never called |
| `testInitializeSucceedsWhenSeedIsRelevantToExistingAccounts` | the guard doesn't false-positive on an ordinary relaunch |
| `testInitializeSkipsSeedRelevanceCheckForEmptyWallet` | fresh-wallet creation/restore is never gated by a check with nothing to compare |
| `testInitializeSkipsSeedRelevanceCheckWithoutSeed` | view-only / no-seed callers are unaffected |

The imported-only exemption lives in the Rust core, so it isn't unit-testable from Swift — the guard only honors what Rust reports.

Verified locally against the branch's pinned `2.7.0-rc.1` XCFramework (no local FFI):

- `swift build` — clean
- `swift test --filter OfflineTests` — **396 tests, 0 failures**
- `swiftlint` on the changed sources — no new violations
- Red-green checked: commenting out the `validateSeedAgainstExistingAccounts` call fails the mismatch test and only that test

## Consumer

`zodl-ios` already consumes `ZcashError.initializerSeedMismatch` in its `RootInitialization` heal path; the symbol exists only on `release/2.6.0` today, which is what motivated the port.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+## Added
+- `ZcashError.initializerSeedMismatch` (`ZINIT0006`): `prepare(with:walletBirthday:for:name:keySource:)`
+  now validates the provided seed against the wallet's existing seed-derived accounts and throws
+  instead of silently opening a wallet the seed cannot spend from. Restoring a different wallet
+  requires `wipe()` first. Wallets whose only accounts are imported (hardware-wallet UFVKs) are
+  exempt — the relevance check is delegated to the Rust core, which treats "there is no seed-derived
+  account to validate against" as relevant, so a hardware-wallet-only wallet is never bricked. See
+  `MIGRATING.md`.
+
 ## Changed
 - `Initializer.initialize` / `Synchronizer.prepare` now return `InitializationResult.seedNotRelevant` instead of silently proceeding when the rust layer reports the provided seed is not relevant to the wallet database (breaking change: `InitializationResult` gained a new case, so exhaustive switches over it must add a case; see MIGRATING.md). Previously this case was indistinguishable from `.success`: account creation was skipped (accounts already existed) and callers proceeded as if they had prepared the wallet they expected, even when the database on disk belonged to a different wallet than the provided seed (e.g. a device-backup restore that brings back `data.db` without the matching keychain seed). Callers must now handle `.seedNotRelevant` the same way they already handle `.seedRequired`. (MOB-1512)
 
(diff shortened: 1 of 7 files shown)


nuttycom (approved):
utACK

PR #1868, opened 2026-07-26
[#MOB-1455] Report crossing values, and carry migration ids as u32

nuttycom:
Two defects in the migration FFI, from the #1813 review. Targets that branch. Part of #1806.

## 1. Transfer amounts were the note spent, not the value crossed

The engine mints each self-funding note as `crossing + fee buffer`; the transfer spends the whole note, the buffer pays its own fee, and only the remainder crosses (no change output, so the buffer is consumed exactly). Every amount the FFI served came from `funding_notes()`, so schedule rows and the pending-transfer proposal each **overstated the transfer by one transfer fee** — and showed a value that is not the round `{1,2,5}×10ⁿ` denomination the user is asked to approve, nor the amount their Ironwood balance will grow by.

Both paths now subtract the note fee buffer. Rows still pair against `funding_notes()` (so the post-reconciliation pairing with schedule entries is untouched — see the Android SDK's comment about what mispairing did live); the buffer comes off each paired row. The consent echo compares the same values on both sides, so platform and native still agree.

The Android SDK already reports crossing values; this brings Swift into line. Once the pin moves to the next engine rc, both can drop the arithmetic for the engine's own accessors — added in zcash/librustzcash#2767, with a `TODO` here pointing at it.

## 2. Transaction ids were decimal strings

`MigrationTxId` is a `u32`. The FFI rendered it as an owned C string outbound and parsed it back inbound, buying nothing — every id is engine-issued, round-tripped verbatim, never displayed — and costing an allocation and free per schedule row, a parse-failure lane on every inbound call, and a Swift API where an opaque-looking `String` invites platform code to construct one.

Now `u32` on both directions (`FfiTransferProposal`, `FfiPreparedTransfer`, `FfiUnsignedTransferPczt`, `FfiAttentionReason::InvalidTransfer`, `zcashlc_migration_record_transfer_result`, the signed-PCZT store calls' `ids` arrays), with the per-id string frees gone from the destructors. `FfiPreparedTransfer`'s "nothing due" sentinel is a null `pczt` alone, which it always also was. Swift models and `migrationRecordTransferResult` take `UInt32`.

### A test that was passing for the wrong reason

Two `MigrationFFITests` cases asserted "recording a result with no active run throws" while passing the id `"does-not-exist"` — so what threw was the id decode, and the assertion never reached the contract it named. With real ids they fail, because `.networkError` deliberately records nothing (`1 => Ok(true)`). Both now use `.success`, which does have to load the run, and a new case documents the `.networkError` no-op.

## Testing

`cargo test --lib` 73/73 (new coverage that a schedule row carries the crossing value, not the funding note); `swift test --filter OfflineTests` 530/530 against a locally built FFI; `cargo fmt` clean, no new clippy warnings.

## Note for merge order

This touches `FfiPreparedTransfer` alongside #1853 (opportunistic proving), which adds a `status` field to the same struct. Whichever lands second needs a trivial conflict resolution — the two changes are independent, and the `status` field makes the "nothing due" sentinel explicit rather than implicit in the null `pczt`.

### Which identifier this is

The engine id names the **transfer**, not one broadcast attempt, and that is why it stays the key here rather than being replaced by the real `txid`: `rebuild_expired_transfer` keeps the transfer id while producing a new transaction with a new `txid`, so an app that keyed its records on the `txid` loses the thread exactly when a transfer expires and is rebuilt. `PreparedMigrationTransfer` carries both — use `txid` to talk to the network about a transaction, `id` to talk about the transfer. zcash/librustzcash#2767 renames the engine type `MigrationTxId` → `MigrationTransferId` to make that explicit; when the pin moves to that rc, the references here rename with it.

## Notes for #1812's rebase

#1812 rebases onto `michal/MOB-1455/MOB-1495-sdk-pool-migration-ffi` after this lands. Textual conflicts are expected in `MigrationModels.swift`, `ZcashRustBackend.swift`, `ZcashRustBackendWelding.swift` and `rust/src/migration.rs`. Three semantic points to respect:

1. **Ids are `UInt32`.** `MigrationTransferProposal`, `PreparedMigrationTransfer`, `MigrationUnsignedTransferPczt`, `MigrationSignedTransferPczt`, `MigrationAttentionReason.invalidTransfer`, and `migrationRecordTransferResult(transferId:)` all move off `String`. #1812's `welding.migrationRecordTransferResult(transferId: prepared.id, ...)` keeps compiling unchanged, since both sides move together.

2. **The id join gets simpler.** #1812's transaction-status model already declares `public let id: UInt32`, and its doc says status rows join to schedule rows via `String(status.id) == MigrationTransferProposal.id`. With ids now `UInt32` on both sides that becomes plain `status.id == proposal.id`; the `String(...)` conversion and the "opaque string id" wording should go.

3. **`MigrationTransferProposal.amount` is now net.** It is the value that crosses into Ironwood, not the funding note spent. #1812's `ImmediateMigrationProposal.amount` is already documented as "the net swept amount -- what arrives in the Ironwood pool once mined" and computed as `sweptPaymentValue(of:) - fee`. Before this PR those two surfaces disagreed (schedule gross, immediate net); after it they agree, so any comment or logic compensating for the difference can be dropped.

---

This PR was prepared with Claude Code assistance.

### DIFF
--- a/Sources/ZcashLightClientKit/Model/MigrationModels.swift
+++ b/Sources/ZcashLightClientKit/Model/MigrationModels.swift
@@ -77,8 +77,8 @@ public struct NoteSplitProposal: Equatable, Sendable {
 
 /// A single scheduled Orchard -> Ironwood transfer, as one element of a `MigrationSchedule`.
 public struct MigrationTransferProposal: Identifiable, Equatable, Sendable, Codable {
-    /// The transfer's opaque, engine-issued id.
-    public let id: String
+    /// The transfer's engine-issued id.
+    public let id: UInt32
     /// The value that crosses the turnstile: what this transfer adds to the destination pool, and
     /// one of the round `{1,2,5}×10ⁿ` denominations the run was planned in. The note the transfer
     /// spends is larger — it also carries the buffer that pays the transfer's own fee — but that
@@ -96,7 +96,7 @@ public struct MigrationTransferProposal: Identifiable, Equatable, Sendable, Coda
 
     /// Creates a `MigrationTransferProposal`.
     public init(
-        id: String,
+        id: UInt32,
         amount: Zatoshi,
         anchorHeight: BlockHeight,
         nextExecutableAfterHeight: BlockHeight,
@@ -264,8 +264,8 @@ public struct MigrationRunEstimate: Equatable, Sendable {
 /// A fully proven, signed migration transaction persisted by the engine, ready for the platform
 /// to broadcast (see `ZcashRustBackendWelding.migrationExtractBroadcastTx(pczt:for:)`).
 public struct PreparedMigrationTransfer: Equatable, Sendable {
-    /// The transfer's opaque, engine-issued id.
-    public let id: String
+    /// The transfer's engine-issued id.
+    public let id: UInt32
     /// The finalized transaction's id, in the SDK's raw/internal byte order (matching `TxId.id`,
     /// not the reversed display-hex order produced by `Data.toHexStringTxId()`). Zeroed when the
     /// value is a STORAGE RECEIPT (`migrationStoreSignedNoteSplitPczts`) whose transaction has not
@@ -275,7 +275,7 @@ public struct PreparedMigrationTransfer: Equatable, Sendable {
     public let pczt: Data
 
     /// Creates a `PreparedMigrationTransfer`.
-    public init(id: String, txid: Data, pczt: Data) {
+    public init(id: UInt32, txid: Data, pczt: Data) {
         self.id = id
         self.txid = txid
         self.pczt = pczt
@@ -304,7 +304,7 @@ public enum MigrationTransferResult: Equatable, Sendable {
 /// Why a migration requires user attention, as carried by `MigrationState.requiresAttention`.
 public enum MigrationAttentionReason: Equatable, Sendable {
     /// The input note funding `transferId` was spent externally before its transfer broadcast.
-    case invalidTransfer(transferId: String)
+    case invalidTransfer(transferId: UInt32)
     /// A transaction's anchor/expiry elapsed before it could be broadcast.
     case transferExpired
 }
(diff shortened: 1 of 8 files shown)


Sources/ZcashLightClientKit/Model/MigrationModels.swift
```diff
@@ -93,7 +96,7 @@ public struct MigrationTransferProposal: Identifiable, Equatable, Sendable, Coda
 
     /// Creates a `MigrationTransferProposal`.
     public init(
-        id: String,
+        id: UInt32,
```
dannywillems, line 99: LGTM. It is a u32 now on the FFI.

Tests/OfflineTests/MigrationFFITests.swift
```diff
@@ -212,31 +212,40 @@ final class MigrationFFITests: XCTestCase {
...
+    /// (Before transfer ids became `UInt32` this test passed `.networkError` with an unparseable
+    /// string id, so what threw was the id decode, not the missing run. `.networkError` records
+    /// nothing by design — see `testRecordTransferResultForANetworkErrorSucceedsWithNoActiveRun`
+    /// — so it can never exercise this contract.)
     func testRecordTransferResultWithNoActiveRunThrows() async throws {
         do {
             try await rustBackend.migrationRecordTransferResult(
-                transferId: "does-not-exist",
-                result: MigrationTransferResult.networkError(retryable: true),
+                transferId: 4_294_967_295,
```
dannywillems, line 226: Note: this is the max value of u32.

Tests/OfflineTests/MigrationFFITests.swift
```diff
@@ -212,31 +212,40 @@ final class MigrationFFITests: XCTestCase {
...
-            XCTFail("Expected recording a result with no active migration run to throw")
+            XCTFail("Expected recording a success with no active migration run to throw")
         } catch ZcashError.rustMigrationRecordTransferResult {
             // expected
         } catch {
             XCTFail("Expected rustMigrationRecordTransferResult but got \(error)")
         }
     }
 
+    /// A network error is a Swift-level signal for the caller's own retry policy: the native side
```
dannywillems, line 238: LGTM for the test. No error handling, therefore passes. But wondering if it is not out of the scope of this patch.

dannywillems (approved):
utACK. Simplification of the API to use a u32 directly, instead of casting to a C string, in addition to reporting correctly the crossing values.

Issue #1892, opened 2026-07-28
[Slipstream] Alternate-server rejection is an assert: a debug crash and a release no-op


nuttycom:
## Problem

`SlipstreamEngine.pushAlternates` (`Sources/ZcashLightClientKit/Slipstream/SlipstreamEngine.swift:125`) checks the FFI's return value with:

```swift
assert(accepted, "alternate-server URIs rejected: \(lastErrorMessage(fallback: "unknown"))")
```

`assert` is compiled out in release builds, so the two behaviours are inverted from what you want: in debug a rejected list crashes the host app, and in release the failure is discarded and the engine silently keeps whatever alternate list it had (or none), with the caller believing the new list took effect.

The doc comment argues a rejection is a programming error, since the URIs are built from typed endpoints. That is a fair reading of today's call sites, but it is an argument for logging loudly rather than for a check that disappears in the configuration users actually run — and consent revocation flows through this path (`setAlternates([])` is how a host disarms failover), so a silently-dropped update has a privacy-shaped failure mode, not just a performance one.

## Target branch

Fix this on `feature/ironwood-slipstream`.

## Suggested direction

Log the rejection through the injected logger at error level and leave the stored `alternates` unchanged (or clear it, whichever the failover semantics want), so debug and release behave the same. If a caller needs to know, `setAlternates` can become throwing — but a logged no-op is probably right for an optimization path.

## Acceptance criteria

- A rejected alternate-server list behaves identically in debug and release.
- The rejection is visible in logs, with the FFI's last-error message.
- Stored state after a rejection is documented and matches what the engine actually holds.

---
Filed with Claude Code assistance during review of PR #1818.

Issue #1891, opened 2026-07-28
[Slipstream] Ironwood coverage gaps on the slipstream sync path


nuttycom:
## Problem

Three places on this path still assume a pre-NU6.3 world, where Sapling and Orchard are the only shielded pools that matter.

**1. The restore-time balance is reported as Orchard value.** The engine's recovery view is a per-account net with no pool breakdown, so `AccountBalance::override_with_recovery_net` (`rust/src/ffi.rs`) surfaces the whole clamped net in one pool and zeroes the rest — historically Orchard. Hosts gate the "Migration Required" prompt on a nonzero Orchard balance, so once NU6.3 is active a restoring wallet whose funds are entirely in Ironwood is told to migrate, on the strength of a value that carries no pool information at all. It self-corrects when the hold releases (≤120 s), but the prompt is actionable in the meantime.

**2.** `ZcashTransaction.Output.Pool` **has no** `.ironwood` **case** (`Sources/ZcashLightClientKit/Entity/TransactionEntity.swift`), so every Ironwood output decodes as `.other(4)` and an app cannot name the pool its funds are in. Pool code 4 is `zcash_client_sqlite`'s `pool_code(PoolType::Shielded(ShieldedPool::Ironwood))`.

**3. The engine's reconciliation view has no Ironwood arm.** `ext_slipstream_v_tx_reconciled` tests `spend_pool = 2` and `= 3` only, so a dangling Ironwood spend reads as reconciled. After activation essentially every dangling spend during a restore is an Ironwood spend, so the view degrades toward "always reconciled". Filed upstream as [zodl-inc/slipstream#9](<https://github.com/zodl-inc/slipstream/issues/9>); noted here because it caps what the SDK-side gate can deliver even once its own defect (zcash/zcash-swift-wallet-sdk#1890) is fixed.

## Status

(1) and (2) are fixed on `kris/merge-maint-v2.8.x` (commit `2cd058b9`, in flight as [#1884](<https://github.com/zcash/zcash-swift-wallet-sdk/issues/1884>) → `michal/slipstream-support`): the collapsed recovery net now routes to Ironwood when NU6.3 is active at the chain tip and stays in Orchard before activation, and `Output.Pool` gained `.ironwood`. This issue tracks confirmation on the target branch plus the open question below; (3) is upstream.

## Open question on (1)

[...]

Issue #1890, opened 2026-07-28
[Slipstream] The reconciliation gate reads a view name the engine no longer creates, so it silently does nothing


nuttycom:
## Problem

`TransactionSQLDAO.unreconciledTxids()` (`Sources/ZcashLightClientKit/DAO/TransactionDao.swift:126`) queries:

```sql
SELECT txid FROM slipstream_v_tx_reconciled WHERE reconciled = 0
```

The engine creates that view under a different name. At the rev this PR pins (`f86eea53`), `core/src/reconcile.rs` declares `RECONCILE_VIEW_NAME = "ext_slipstream_v_tx_reconciled"` and creates `ext_slipstream_v_tx_reconciled`; the `ext_slipstream_*` schema migration explicitly runs `DROP VIEW IF EXISTS slipstream_v_tx_reconciled` for the legacy unprefixed name (`core/src/migrations.rs:78`).

So on any engine-provisioned database the query throws "no such table", the `catch { return [] }` swallows it, and `droppingUnreconciled` (`SlipstreamSynchronizer.swift:845`) holds nothing back. The phantom-receive suppression — a transaction whose spend was scanned before its input's origin block showing as a receive during a recent-first restore — never engages.

The defect is invisible from the test suite because `Tests/OfflineTests/SlipstreamReconcileReadTests.swift:27` builds its fixture under the *wrong* name:

```swift
try setup.run("CREATE TABLE slipstream_v_tx_reconciled (txid BLOB, reconciled INTEGER)")
```

The DAO reads what the test creates, so the suite is green while production reads nothing.

## Target branch

Fix this on `feature/ironwood-slipstream`.

## Suggested direction

- Read the name the engine publishes (`ext_slipstream_v_tx_reconciled`), ideally sourced from the engine rather than duplicated as a Swift string literal, so a future rename fails loudly instead of silently.
- Rename the fixture in `SlipstreamReconcileReadTests` to match, and add coverage that exercises a database provisioned by the engine's own migration (or at minimum asserts the exact view name against the engine constant), so a stub cannot re-encode the bug.
- Narrow the `catch`. A missing view is a legitimate state on a database the engine has never opened, but it is a defect on one it has; swallowing every SQL error conflates the two. Distinguish "view absent" and log it once at debug, and surface anything else.

[...]

Issue #1889, opened 2026-07-28
[Slipstream] Clear isRunning during wipe teardown


nuttycom:
## Problem

`SlipstreamSynchronizer.wipeImpl` stops polling, stops and closes the engine, deletes wallet data, and publishes the unprepared state, but it never sets the private `isRunning` flag to false.

When a running synchronizer is wiped and then prepared again, later operations such as `importAccount`, `deleteAccount`, or `switchTo(endpoint:)` can observe the stale true value and behave as though synchronization is active. This can trigger an unexpected restart before the caller has invoked `start()`.

The flag should be cleared before any throwing file operation so both successful and failed wipe paths accurately reflect that the engine has been stopped and closed.

## Target branch

Fix this on `feature/ironwood-slipstream`.

## Acceptance criteria

- `isRunning` becomes false whenever wipe teardown stops the engine.
- The state is correct even when file deletion subsequently fails.
- After wipe followed by prepare, account or endpoint operations do not start synchronization unless the caller explicitly started it.
- Regression coverage begins with a running synchronizer and exercises wipe → prepare → another lifecycle operation.

Filed with OpenAI Codex AI assistance during review of PR #1818.

PR #1893, opened 2026-07-28
[#1806] Orchard→Ironwood pool migration, custom networks, and the Slipstream sync engine

nuttycom:
Adds the ability to move an existing Orchard balance into Ironwood, ahead of the NU6.3 turnstile closing.

Closes zcash/zcash-swift-wallet-sdk#1806.  <br>Resolves zcash/zcash-swift-wallet-sdk#1892  <br>Resolves zcash/zcash-swift-wallet-sdk#1891  <br>Resolves zcash/zcash-swift-wallet-sdk#1890  <br>Resolves zcash/zcash-swift-wallet-sdk#1886  <br>Resolves zcash/zcash-swift-wallet-sdk#1889  <br>Resolves zcash/zcash-swift-wallet-sdk#1888  <br>Resolves zcash/zcash-swift-wallet-sdk#1887

## Why

2.8.0-rc.1 already receives, syncs and holds Ironwood value. What it cannot do is get a wallet's *existing* Orchard balance across: at NU6.3 the Orchard turnstile closes, so value can leave Orchard but nothing new enters it, and anything left behind stays behind. This branch supplies the crossing.

## Migrating an Orchard balance

A real balance cannot cross in one transaction. A single sweep is one conspicuous, all-at-once event that ties every note together, and it cannot be shaped into the denominations the protocol wants. The migration is therefore staged, and the SDK owns the staging:

* the balance is split into round `{1,2,5}×10ⁿ` denominations,
* transfers are scheduled over time on ZIP 318's randomized cadence, so the pattern of crossings does not itself identify the wallet,
* each transfer is proven ahead of its window and broadcast when it arrives, over a dedicated Tor runtime with fail-closed semantics (never a silent clearnet fallback) to one explicitly chosen submission endpoint,
* sync and broadcast stay in separate sessions: a wallet-scope gate blocks sync for ten minutes after a broadcast, and broadcasting during a sync is refused.

The app talks to one 32-member migration group on `Synchronizer`; the per-account engine, broadcaster and privacy gate behind it are internal, and two accounts can migrate concurrently. The group covers the state machine and progress, note-split propose/sign/submit, gradual and immediate lanes, height-gated background delivery (which works without `prepare()`, so a background session can deliver without starting sync), overdue and expiry recovery, the sub-0.01-ZEC residual opt-in with balance locking, a run-count estimate for multi-round migrations, live per-transaction statuses, and a full external-signer path for hardware wallets — including a Keystone batch-signing bridge that carries the ceremony's PCZTs over animated QR.

Consent is verified rather than assumed: proposals re-randomize per ZIP 318, so every commit-shaped call checks what it is about to sign against the plan the user actually saw, and a mismatch surfaces as "propose again" instead of silently signing different values.

## Slipstream, an alternative sync engine

`SlipstreamSynchronizer` is a second `Synchronizer` implementation backed by the `slipstream-core` engine: non-linear Spend-before-Sync scheduling, density-adaptive concurrent fetch, sparse commitment-tree persistence, poll-model snapshots, and a stall watchdog. It shares one `data.db` with the existing engine, and hosts opt in by constructing it — `SDKSynchronizer` remains the default.

## Exercising it before activation

Custom (regtest-style) networks with configurable activation heights let the whole path — receive, migrate, spend — run against a modified-parameter chain ahead of mainnet or testnet activation.

## What is not here

Shielded voting is off on this line, as on the 2.8 maintenance line: the module is gated behind `cfg(zcash_voting)`, its dependency is commented out, and the Swift surface is removed.

## Provenance and follow-ups

Replaces zcash/zcash-swift-wallet-sdk#1885, which was cut from `feature/ironwood-support`; that line is fully contained in this one and is being deleted.

Known follow-ups, none blocking: zcash/zcash-swift-wallet-sdk#1886 (FFI handle lifetime), zcash/zcash-swift-wallet-sdk#1890 (the reconciliation gate reads a stale view name and silently does nothing), zcash/zcash-swift-wallet-sdk#1887, zcash/zcash-swift-wallet-sdk#1888, zcash/zcash-swift-wallet-sdk#1889, zcash/zcash-swift-wallet-sdk#1891 (Ironwood coverage), zcash/zcash-swift-wallet-sdk#1892, and engine-side [zodl-inc/slipstream#9](<https://github.com/zodl-inc/slipstream/issues/9>).

Draft, mirroring zcash/zcash-swift-wallet-sdk#1885's state — mark ready once those are triaged against the release.

---


nuttycom:
This should be automatically closed as complete by the public 3.0.0 release.

Issue #1894, opened 2026-07-28
[Slipstream] The engine-backed immediate migration lane is orphaned, and its commit-time reschedule is documented but absent


nuttycom:
## Problem

`zcashlc_migration_propose_immediate_transfers` has no caller. The immediate lane became an ordinary send-max sweep — `proposeSendMaxTransfer(orchardOnly: true)` followed by `recordImmediateMigration` — and the welding member that reached the extern was removed with that change, so the only path passing `immediate: true` to `plan_and_cache` is itself unreachable.

The flag it sets is never read. `migration_plan_cache::CachedPlan.immediate` is written and never consulted (``field `immediate` is never read`` on every build), while `commit_or_resume` documents that "when the cached preview came through the immediate lane, the committed transfers' scheduled heights are rewritten to the commit tip". Nothing performs that rewrite here; the block that does exists only on `feature/ironwood-support`, which the send-max redesign superseded.

So the doc promises a behaviour the line does not have, and the compiler has been pointing at the evidence.

## Why it matters

Dead code that still compiles still gets maintained. Merging maint/v2.8.x turned up that this extern passed gross funding notes to `schedule_rows`, overstating every previewed transfer by one fee buffer — found and fixed in a function no caller reaches.

## Target branch

`feature/ironwood-slipstream`.

## Acceptance criteria

- No entry point, cache field or parameter survives for a lane the SDK does not drive.
- No doc comment describes a commit-time reschedule that does not happen.
- The build is free of the dead-field warning.

---
Filed with Claude Code assistance.

nuttycom:
Closing in favor of https://github.com/zodl-inc/zodl-swift-wallet-sdk/issues/93

PR #1895, opened 2026-07-28
[#1806] Drop the orphaned engine-backed immediate migration lane

nuttycom:
Removes dead scaffolding the send-max redesign left behind, and a doc comment that promised behaviour this line does not have.

Closes zcash/zcash-swift-wallet-sdk#1894.<br>Resolves zcash/zcash-swift-wallet-sdk#1894

## What was left behind

`zcashlc_migration_propose_immediate_transfers` has had no caller since the immediate lane became an ordinary send-max sweep — `proposeSendMaxTransfer(orchardOnly: true)` followed by `recordImmediateMigration`. The welding member that reached the extern went with that change and the Swift surface never grew a replacement, so the only path that passed `immediate: true` to `plan_and_cache` was itself unreachable.

The flag it set was never read. `CachedPlan.immediate` was written and never consulted (the compiler said so on every build: `` field `immediate` is never read ``), while `commit_or_resume` documented that "when the cached preview came through the immediate lane, the committed transfers' scheduled heights are rewritten to the commit tip". No code performs that rewrite on this line — the block that does exists only on `feature/ironwood-support`, which the send-max redesign superseded.

## Why remove it rather than leave it

Dead code that still compiles still gets maintained. Merging maint/v2.8.x turned up that this extern passed gross funding notes to `schedule_rows`, overstating every previewed transfer by one fee buffer — a real bug, found and fixed in a function no caller reaches. Deleting it removes the surface that invites that.

So the extern, the `plan_and_cache` parameter, the `CachedPlan` field and the false promise all go. No behaviour changes: the immediate lane that actually ships is the send-max one, untouched.

## Also here

The entry-point census in `rust/CHANGELOG.md` had drifted independently — the bullet describes 21 `zcashlc_migration_*` entry points out of a stated total of 24, but the real total is 30 once the per-transaction statuses read, the four Keystone batch-signing members and the debug fast-reschedule are counted. Corrected while adjacent.

No CHANGELOG entry for the removal itself: the extern arrived after 2.8.0-rc.1 and never appeared in a release — which is the answer to "does this need a note?".

`cargo test` 111/111, `swift test --filter OfflineTests` 691/691, fmt and clippy clean.

---

Resolves COR-1883

### DIFF
--- a/rust/CHANGELOG.md
+++ b/rust/CHANGELOG.md
@@ -28,8 +28,10 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `zcash_client_sqlite::pool_migration`, both on librustzcash main; the family pin
   targets a plain main rev — boundary-anchor proving (#2710) and owner-keyed
   note locking (#2716) are both merged, nothing unmerged remains):
-  21 of the 24 `zcashlc_migration_*` entry points (the residual-locking pair
-  and the run-count estimate ride their own entries below) plus the
+  21 of the 30 `zcashlc_migration_*` entry points (the residual-locking pair,
+  the run-count estimate, the per-transaction statuses, the four Keystone
+  batch-signing members, and the debug fast-reschedule ride their own entries
+  below) plus the
   `zcashlc_ironwood_activation_height` helper, with their `#[repr(C)]` return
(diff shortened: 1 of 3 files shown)


Chlup (approved):

Chlup (dismissed):
utACK

PR #1896, opened 2026-07-28
[MOB-1566] Treat the decrypt-and-store FFI error sentinel as an error

Chlup:
## Problem

`zcashlc_decrypt_and_store_transaction` returns `1` on success and `-1` on error; it never returns `0`. The Swift wrapper guarded on `result != 0`, which both values satisfy, so the throw branch was dead: a failed decrypt-and-store returned the zero-initialized txid buffer as if it had succeeded.

Callers discard that return value, so a real FFI failure was recorded as completed work — transaction enhancement marked the request done without persisting the transaction, the mempool monitor looked up the all-zero txid and silently dropped the transaction, and `enhanceTransactionBy` reported success — all without surfacing any error or triggering the enhancer's retry loop.

## Fix

Guard on `result == 1` so a failure throws `ZcashError.rustDecryptAndStoreTransaction` with the underlying Rust error message.

## Tests

- New failure-path test pins the sentinel semantics against the real FFI: malformed transaction bytes must throw (previously returned an all-zero txid without throwing).
- New success-path test pins that a stored transaction returns its real txid rather than the zeroed placeholder.
- Full OfflineTests suite on this base: 695 tests, 0 failures.

This is the only `guard result != 0` occurrence in `ZcashRustBackend`; other i32-returning FFI wrappers were audited and use correct sentinels. (The one remaining `result != 0` in the file is a boolean conversion in `isSeedRelevantToAnyDerivedAccount`, guarded separately by `result >= 0`.)

## Base

Retargeted from `maint/v2.7.x` to `feature/ironwood-slipstream` and rebased onto it. The source and test changes applied unchanged; only the `CHANGELOG.md` entry was refiled into that branch's Unreleased `## Fixed` section, since the two lineages have different release histories. The sentinel contract is unchanged on this base — `rust/src/lib.rs` still returns `Ok(1)` on success and `-1` via `unwrap_exc_or` on error.


LukasKorba (approved):
utACK

PR #1898, opened 2026-07-28
[MOB-1568] Submit sequentially instead of fanning out when Tor is off

Chlup:
## Problem

`MultiEndpointSubmitter` raced the same raw transaction to every endpoint in a task group regardless of the Tor state. With Tor enabled that is safe — each attempt rides its own isolated circuit — but with Tor disabled every attempt is a clearnet gRPC connection from the device's real IP, so a single send handed the IP↔txid link to every configured server operator simultaneously, including the wallet's own sync server.

## Fix

Give `SubmissionRace` a dispatch mode chosen from `SDKFlags.torEnabled` at submission time:

- **Tor on** keeps the parallel race, first-acceptance resolution, and the post-acceptance grace window unchanged.
- **Tor off** now tries endpoints strictly one at a time in list order and stops at the first acceptance, so operators beyond the accepting one are never contacted. Rejections and transport failures still fail over to the next endpoint, the shared response timeout bounds the whole attempt, caller cancellation ends the chain, and there is no grace window because nothing runs behind a sequential acceptance.

The background resubmission path (`SubmitPlanExecutor`) already retried sequentially with stop-on-first-accept and is unchanged.

## Tests

Five new tests for the Tor-off mode (stop-at-first-acceptance with no further endpoints contacted, no overlapping submissions via a new max-concurrency probe on the mock, deterministic first-rejection outcome, timeout bounding, caller cancellation stopping the chain). All 12 pre-existing parallel-race tests pass unmodified, pinning the Tor-on behavior. Full OfflineTests suite: 401 tests, 0 failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+## Changed
+- Multi-endpoint transaction submission no longer fans the same transaction
+  out to every endpoint in parallel when Tor is disabled. With Tor off, each
+  submission is a clearnet connection from the device's real IP, so the
+  parallel race handed every contacted server operator the IP↔txid link at
+  once. Endpoints are now tried one at a time in list order, stopping at the
+  first acceptance, so at most one operator (plus any that already failed to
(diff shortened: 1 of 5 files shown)


pacu (approved):
utACK

PR #1903, opened 2026-07-28
Build every Apple architecture in one cargo invocation

### LINKED ISSUE #1901: FFI xcframework build compiles each architecture serially
`BuildSupport/Makefile` builds the five Apple targets with five separate
`cargo build --target <triple>` invocations. These cannot be overlapped with
`make -j`: concurrent cargo invocations sharing `../target` block on an
exclusive lock over the host artifact directory, reporting
`Blocking waiting for file lock on artifact directory`.

That is expensive mainly because of `[profile.release] lto = true` in
`Cargo.toml`. Fat LTO is a single-threaded whole-program pass over the
~674-crate graph. Measured on a 16-core machine, with all dependencies warm:

| Build | Wall | User CPU | Utilization |
|---|---|---|---|
| One target, LTO/link tail only | 2m02s | 156s | ~1.0x (serial) |
| All five targets, one invocation | 3m52s | 1924s | ~5.1x |

So the current shape spends roughly ten minutes of strictly single-threaded
time in the release build, one LTO pass after another.

Passing several `--target` flags to a single `cargo build` (stable since
1.64) puts every architecture into one job graph with one lock, letting those
passes run concurrently.

Two related items in the same area:

- `Scripts/prepare-release.sh` runs `make clean`, whose recipe includes
  `rm -rf ../target`. In `.github/workflows/build-ffi.yml` this runs
  immediately after `Swatinem/rust-cache` restores `target/`, so the Rust
  cache has never had any effect on a release build. Splitting the target-dir
  wipe into a separate `distclean` goal would let the restored dependency
  artifacts survive.
- `.github/workflows/swift.yml` and `.github/workflows/codeql.yml` invoke
  `make macos` and install only the two macOS rustup targets, so any change
  here must keep the per-platform goals narrow.

### LINKED ISSUE #1902: build.rs under-declares rerun-if-changed, so the generated zcashlc.h can go stale
`rust/build.rs` generates `target/Headers/zcashlc.h` with  <br>`cbindgen::Builder::new().with_crate(crate_dir)`, which parses the **entire**  <br>crate. But it declares only:

```rust
println!("cargo:rerun-if-changed=rust/src/lib.rs");
println!("cargo:rerun-if-changed=rust/src/voting.rs");
println!("cargo:rerun-if-changed=rust/src/voting");
println!("cargo:rerun-if-changed=rust/wrapper.c");
println!("cargo:rerun-if-changed=rust/wrapper.h");
```

Emitting any `rerun-if-changed` disables cargo's default behaviour of  <br>watching the whole package, so `rust/src/ffi.rs`, `rust/src/ffi/`,  <br>`derivation.rs`, `eip681.rs`, `tor.rs`, `migration.rs` and `os_log.rs` are all  <br>unwatched. Editing any of them recompiles the crate without rerunning the  <br>build script, leaving the previously generated header in place.

Reproduced by appending a new exported function to `rust/src/ffi.rs` and  <br>rebuilding without wiping `target/`:

| `build.rs` | New `extern "C"` fn present in `zcashlc.h`? |
| -- | -- |
| as committed | no (stale header) |
| watching `rust/src` as a directory | yes |

Practical effect: a newly added FFI function can be missing from the  <br>generated header, and a changed signature can go unreflected. Note that  <br>`Bindings::write_to_file` skips writing when content is unchanged, so file  <br>mtime is not a reliable way to observe this — the reproduction above needs an  <br>actual API change.

`Scripts/rebuild-local-ffi.sh` calls `cargo build` directly and is exposed to  <br>this today. The release path happens to be shielded because  <br>`Scripts/prepare-release.sh` runs `make clean`, which currently does  <br>`rm -rf ../target` and so forces regeneration — meaning this becomes  <br>load-bearing if that wipe is ever removed (see zcash/zcash-swift-wallet-sdk#1901).

Fix: watch `rust/src` as a directory, which cargo scans recursively.

nuttycom:
Closes #1901
Closes #1902
Resolves MOB-1601
Resolves MOB-1600

The FFI release build recompiled each of the five Apple architectures with its
own `cargo build --target <triple>`. Those invocations cannot be overlapped
with `make -j`, because concurrent cargo processes sharing `../target` block on
an exclusive lock over the host artifact directory:

```
[ios]    Blocking waiting for file lock on package cache
[darwin] Blocking waiting for file lock on artifact directory
```

The cost concentrates in the link step. `[profile.release] lto = true` is a
single-threaded whole-program pass over a ~674-crate graph, so the five passes
ran back to back.

## Changes

**Single multi-target cargo invocation** (`BuildSupport/Makefile`). Passing
several `--target` flags to one `cargo build` puts every architecture into a
single job graph with one lock, so those LTO passes overlap. Measured on 16
cores with dependencies warm:

| Build | Wall | User CPU | Utilization |
|---|---|---|---|
| One architecture, LTO/link tail only | 2m02s | 156s | ~1.0x (serial) |
| All five, one invocation | 3m52s | 1924s | ~5.1x |

The per-platform goals keep their own narrow architecture lists, so
`make macos` still needs only the two macOS rustup targets installed — which
`swift.yml` and `codeql.yml` rely on. `make macos` also drops from two cargo
invocations to one. A full build takes `cargo-all` as its first prerequisite,
after which the per-platform rules reduce to a fingerprint check.

**`clean` / `distclean` split.** `clean` now removes `products/` only;
`distclean` additionally wipes `../target`. `Scripts/prepare-release.sh` runs
`make clean` immediately after `build-ffi.yml` restores `Swatinem/rust-cache`,
so that cache had never affected a release build. Products are still rebuilt
from scratch, so no stale slice can survive `clean`.

**`rust/build.rs` header-staleness fix.** Removing the `target/` wipe from
`clean` exposed a pre-existing bug that wipe had been masking. `build.rs`
generates `zcashlc.h` via cbindgen, which parses the whole crate, but declared
`rerun-if-changed` for only `lib.rs`, `voting.rs`, `voting/` and the wrappers.
Emitting any `rerun-if-changed` disables cargo's default whole-package
watching, leaving `ffi.rs`, `derivation.rs`, `eip681.rs`, `tor.rs`,
`migration.rs` and `os_log.rs` unwatched. Verified by appending a new exported
function to `rust/src/ffi.rs` and rebuilding without wiping `target/`:

| `build.rs` | New `extern "C"` fn present in `zcashlc.h`? |
|---|---|
| before | no (stale header) |
| after | yes |

`Scripts/rebuild-local-ffi.sh` calls `cargo build` directly and was exposed to
this already.

## Verification

- `make macos` builds only the macOS architectures and emits a single cargo
  invocation naming just those two triples.
- A full `make xcframework` produces all three slices with the expected
  architectures (`ios-arm64` → `arm64`, `ios-arm64_x86_64-simulator` →
  `x86_64 arm64`, `macos-arm64_x86_64` → `x86_64 arm64`), with `Headers`,
  `Modules/module.modulemap` and both `Info.plist` files in place.
- Header regeneration confirmed in both directions (function added → appears;
  reverted → disappears).

## Notes for review

- Releases now build incrementally rather than from scratch. Switching
  `Scripts/prepare-release.sh` to `make distclean` would restore the previous
  from-scratch behaviour at the cost of the cache benefit — happy to change it
  if that hermeticity is wanted for tagged releases. Note that
  `Swatinem/rust-cache` strips workspace-local artifacts before saving, so
  `libzcashlc` itself and its LTO link always rerun regardless; the saving is
  on the dependency crates.
- Both workflows hash `BuildSupport/Makefile` into their FFI cache key, so this
  invalidates that cache once.
- Not addressed here: `lto = "thin"` would cut the per-architecture cost
  itself, but it changes codegen for scanning and proving, so it wants a
  runtime benchmark rather than a build-time one.

### DIFF
--- a/BuildSupport/Makefile
+++ b/BuildSupport/Makefile
@@ -10,6 +10,18 @@ IOS_DEVICE_ARCHS = aarch64-apple-ios
 IOS_SIM_ARCHS_STABLE = x86_64-apple-ios  aarch64-apple-ios-sim
 MACOS_ARCHS = x86_64-apple-darwin aarch64-apple-darwin
 IOS_SIM_ARCHS = $(IOS_SIM_ARCHS_STABLE)
+ALL_ARCHS = $(IOS_DEVICE_ARCHS) $(IOS_SIM_ARCHS) $(MACOS_ARCHS)
+
+# Cargo serialises concurrent invocations behind a lock on the shared host
+# artifact directory (../target/release), so running one `cargo build --target X`
+# per architecture cannot overlap them — `make -j` does not help. That matters
+# most for the link step: `lto = true` is a single-threaded whole-program pass
+# costing roughly two minutes per architecture, so five of them run back to back.
+#
+# Passing several `--target` flags to one invocation puts every architecture in a
+# single job graph, letting those passes run concurrently.
+CARGO_BUILD = RUSTUP_TOOLCHAIN=stable cargo build --manifest-path ../Cargo.toml --release
+cargo_targets = $(addprefix --target ,$(1))
 
 RUST_SRCS = $(shell find ../rust -name "*.rs") ../Cargo.toml
 STATIC_LIBS = $(shell find ../target -name "libzcashlc.a")
@@ -23,15 +35,26 @@ install:
 release: clean xcframework
 .PHONY: release
 
+# Discard assembled build products. ../target is deliberately preserved so that
+# cargo's fingerprints (and CI's restored Rust cache) still apply; cargo decides
+# what genuinely needs rebuilding. Every framework under products/ is rebuilt
+# from scratch, so no stale slice can survive this.
 clean:
 	rm -rf products
-	rm -rf ../target
 .PHONY: clean
+
+# Additionally discard the cargo target directory, forcing a from-scratch build.
+distclean: clean
+	rm -rf ../target
+.PHONY: distclean
 
 xcframework: products/libzcashlc.xcframework
 .PHONY: xcframework
 
-products/libzcashlc.xcframework: $(PLATFORMS)
+# cargo-all is listed first so a full build compiles every architecture in one
+# invocation. The per-platform cargo rules below then find their work already
+# done and reduce to a fingerprint check.
+products/libzcashlc.xcframework: cargo-all $(PLATFORMS)
 	rm -rf $@
 	mkdir -p $@
 	cp -R products/ios-device/frameworks $@/ios-arm64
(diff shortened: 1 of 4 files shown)


Chlup (approved):
utACK

PR #1905, opened 2026-07-28
Restore the minedTransaction event and surface setTransactionStatus failures

pacu:
## Summary

Three related fixes around how the SDK reports that a sent transaction has been mined.

### `SynchronizerEvent.minedTransaction` was dead

`BlockEnhancerImpl.enhance` takes a `didEnhance` callback and has not invoked it since transaction data requests were adopted (e5e1fdf3), so the emit site in `EnhanceAction` was unreachable and the event never fired in production. Client apps therefore had no signal that a pending sent transaction had been mined, and had to rediscover it by other means.

The callback is now invoked exactly once per unmined→mined transition of a sent transaction:

- the `.enhancement` arm captures whether the wallet already knew the transaction as mined *before* `decryptAndStoreTransaction` writes, and reports only on a nil→mined transition (this is the scanning-originated path);
- the `.getStatus` arm reports on a mined answer — such a request only exists for a transaction whose mined height is unknown, so a mined answer is by definition a transition.

`EnhanceAction` and the event plumbing are unchanged.

### `setTransactionStatus` failures were silently discarded

`zcashlc_set_transaction_status` returned `void`, so every error — including the early return when the chain tip is unknown — vanished. It now returns `bool`, and `ZcashRustBackend.setTransactionStatus` throws the new `ZcashError.rustSetTransactionStatus` (`ZRUST0111`), which feeds the block enhancer's existing retry and logging. No `repr(C)` struct layout changes.

The cbindgen-generated `target/Headers/zcashlc.h` is a build artifact rather than a checked-in file, so the new signature does not appear in this diff; #1902 on this branch guarantees it is regenerated. Verified locally after a rebuild:

```
bool zcashlc_set_transaction_status(const uint8_t *db_data,
```

### Server validation for Ironwood

`ValidateServerAction` now throws the new `ZcashError.compactBlockProcessorServerMissingIronwoodSupport` (`ZCBPEO0024`) when the chain is on the NU6.3 consensus branch but the connected server's tip tree state carries no `ironwoodTree` frontier. Compact-block scanning is what detects the wallet's shielded transactions, and a server that omits Ironwood data would otherwise pass silently: absent Ironwood chain metadata reads as zero and satisfies the scanner's tree-size consistency check, so the wallet would simply never detect anything in that pool. The check is block-level and adds one `GetTreeState` call per server validation.

I checked the current default servers before adding it — `zec.rocks`, `us.zec.stardust.rest` and `testnet.zec.rocks` (all lightwalletd v0.5.1) all serve a non-empty `ironwoodTree`, so this only fires against a server that is genuinely out of date.

## Tests

New `BlockEnhancerImplTests`: a mined `GetStatus` answer reports exactly one newly-mined transaction; an unrecognized txid reports nothing; enhancing an already-mined transaction produces no duplicate report; a nil→mined enhancement reports once; and a throwing status write neither crashes nor produces a report.

`swift test --filter OfflineTests`: 331 tests, 0 failures, against the rebased base (`zcash_client_sqlite 0.22.0-rc.5`).

## Review updates

- Rebased onto the current `maint/v2.7.x` (resolves the conflict; picks up the rc.5 dependency bump, #1902 and #1901).
- The error cases are now generated: the two cases were added to `ZcashErrorCodeDefinition.swift` and `generateErrorCode.sh` was run with sourcery 2.3.0. The earlier hand-edits to the generated files have been reverted.


rust/src/lib.rs
```diff
@@ -3213,7 +3213,7 @@ pub unsafe extern "C" fn zcashlc_set_transaction_status(
     txid_bytes: *const u8,
     txid_bytes_len: usize,
     status: ffi::TransactionStatus,
-) {
+) -> bool {
```
schell, line 3216: The PR description says "the header is regenerated by cbindgen" but the regenerated header is not in the diff. Shouldn't it be?

Sources/ZcashLightClientKit/Error/ZcashError.swift
```diff
@@ -383,6 +383,11 @@ public enum ZcashError: Equatable, Error {
     /// sourcery: code="ZRUST0064"
     /// ZRUST0064
     case rustTransactionDataRequests(_ rustError: String)
+    /// Error from rust layer when calling ZcashRustBackend.setTransactionStatus
+    /// - `rustError` contains error generated by the rust layer.
+    /// sourcery: code="ZRUST0111"
+    /// ZRUST0111
+    case rustSetTransactionStatus(_ rustError: String)
```
schell, line 389: I think running `sourcery` would result in a different file than what these hand-edits have produced, and the top of the file says "DO NOT EDIT". You should run `generateErrorCode.sh` after editing `ZcashErrorCodeDefinition.swift` to generate the edits - bummer, I know.

schell (changes requested):
Just a few changes, then CI needs to pass.

rust/src/lib.rs
pacu: Good catch on the description — that wording was wrong and I've fixed it. The header isn't a checked-in artifact: `build.rs` writes it to `target/Headers/zcashlc.h` via cbindgen, and `/target/` is gitignored, so there's nothing for the diff to carry. `git ls-files` in this repo only has `rust/wrapper.h` (the bindgen input).

Your #1902 on this branch ("Watch all of rust/src so the generated header cannot go stale") is what makes that safe for this change in particular: `zcashlc_set_transaction_status` lives in `lib.rs`, which the old `rerun-if-changed` list already covered, but I've rebased onto it regardless so the regeneration is guaranteed. I verified the rebuilt header locally carries the new signature:

```
bool zcashlc_set_transaction_status(const uint8_t *db_data,
```

Sources/ZcashLightClientKit/Error/ZcashError.swift
pacu: You were right, and the regenerated files differ from my hand-edits exactly as you predicted (sourcery emits the description as a single line; I had wrapped it). Fixed properly: installed sourcery 2.3.0, reverted both generated files to the base branch, added the two cases to `ZcashErrorCodeDefinition.swift` only, and ran `generateErrorCode.sh`. The generated files are now untouched by hand.

I also shortened the `compactBlockProcessorServerMissingIronwoodSupport` description while I was in there, so the generated line stays in range with the rest of the file.

pacu:
Both review points are addressed and the branch is rebased onto current `maint/v2.7.x`, so the conflict is gone and CI should have a clean base to run on. Details in the replies above; `swift test --filter OfflineTests` is green (331 tests) against rc.5.

pacu:
Rebased again onto current `maint/v2.7.x` — the 2.7.0-rc.3 release and the rc.6 dependency bump had re-conflicted the changelog. The conflict was confined to `rust/CHANGELOG.md`; the code diff is unchanged (same 10 files).

Re-verified against `zcash_client_backend 0.24.0-rc.6` / `zcash_client_sqlite 0.22.0-rc.6` after rebuilding the local FFI: `swift test --filter OfflineTests` passes 323 tests with 0 failures, and the new `BlockEnhancerImplTests` are 5/5. (`BroadcasterTests` is skipped in that count only because my sandbox cannot read `~/Documents/sapling-spend.params` — a local permissions issue, unrelated to this change.)

pacu:
Verified both points on the pushed branch rather than just asserting them:

**Generated header** — `git ls-files` tracks no `zcashlc.h` (only `rust/wrapper.h`, the bindgen input); `build.rs` writes the cbindgen output to `target/Headers/`, which is gitignored. The freshly built framework header carries the new signature:

```
bool zcashlc_set_transaction_status(const uint8_t *db_data,
```

**Generated error files** — re-running `generateErrorCode.sh` on the pushed tree now produces **zero diff**, so the committed files are byte-identical to sourcery's output and nothing is hand-edited. The only file I touched by hand is `ZcashErrorCodeDefinition.swift`, which carries no "DO NOT EDIT" banner.

While verifying, I found one thing worth fixing that your comment led me to: I had written one of the two annotations as `/// sourcery: code="ZRUST0111"` instead of `// sourcery: ...`. That form leaks the annotation into the generated public API documentation:

```swift
/// Error from rust layer when calling ZcashRustBackend.setTransactionStatus
/// - `rustError` contains error generated by the rust layer.
/// sourcery: code="ZRUST0111"   // <- leaked into the docs
/// ZRUST0111
case rustSetTransactionStatus(_ rustError: String)
```

`ZcashErrorCodeDefinition.swift`'s own header calls this out ("it's important to use `///` for the documentation and only `//` for the sourcery command"), and the file is 209:12 in favour of `//`. Fixed and regenerated; the docs are now clean and the code still maps to `ZRUST0111`.

`swift test --filter OfflineTests` is green (323 tests) against rc.6 after the change — `BroadcasterTests` is excluded from that count only because my local sandbox can't read `~/Documents/sapling-spend.params`.

schell (approved):

PR #1906, opened 2026-07-29
Release zcash-swift-wallet-sdk 2.8.0-rc.2

nuttycom:
Release zcash-swift-wallet-sdk version 2.8.0-rc.2

- **rust: handle non_exhaustive BirthdayError match**
- **[#1902] Watch all of rust/src so the generated header cannot go stale**
- **[#1901] Build every Apple architecture in one cargo invocation**
- **Update to latest librustzcash releases.**
- **Prepare SDK release 2.8.0-rc.2**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ ff = "0.13"
 sapling = { package = "sapling-crypto", version = "0.7", default-features = false }
 transparent = { package = "zcash_transparent", version = "0.10", default-features = false }
 zcash_address = { version = "0.13" }
-zcash_client_backend = { version = "0.24.0-rc.4", features = [
+zcash_client_backend = { version = "0.24.0-rc.5", features = [
     "lightwalletd-tonic-tls-webpki-roots",
     "orchard",
     "pczt",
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -93,6 +93,16 @@ bindgen = "0.72"
 cbindgen = "0.29"
 cc = "1.0"
 
+[patch.crates-io]
+# pczt = { git = "https://github.com/zcash/librustzcash.git", rev = "REPLACE_ME" }
```
Chlup, line 97: This doesn't seem right. But I am not sure what is intention here.

Chlup (changes requested):

rust/CHANGELOG.md
```diff
@@ -23,6 +27,20 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Fixed
+- `build.rs` now watches the whole `rust/src` directory. It previously named
+  only `lib.rs`, `voting.rs` and `voting/`, and emitting any `rerun-if-changed`
+  disables cargo's default whole-package watching, so edits to any other module
+  (`ffi.rs` in particular) recompiled the crate without rerunning the build
+  script. The cbindgen-generated `target/Headers/zcashlc.h` could therefore go
+  stale, leaving a newly added FFI function absent from the header or a changed
+  signature unreflected.
+>>>>>>> maint/v2.7.x
```
Chlup: Artefact after merging

Cargo.toml
nuttycom: This is correct, we no longer need patch revisions, because we are now on published release dependencies. I just left the stanza in here so that it's easy for someone doing development to switch to a patch when they need to.

Chlup (dismissed):

PR #1915, opened 2026-07-30
Vendor proposal.proto from librustzcash by script, and check it in CI

dannywillems:
Ports the fix from zcash-android-wallet-sdk#2093. This repository has the same
gap, and the same drift.

`Sources/ZcashLightClientKit/Modules/Service/GRPC/ProtoBuf/proto/proposal.proto`
is a hand-copied snapshot of `zcash_client_backend/proto/proposal.proto`.
Nothing recorded where it came from and nothing checked it. It is currently
missing `ValuePool.Ironwood`, `Proposal.confirmationsPolicy`,
`Proposal.proposedVersion`, `TransactionBalance.dummyOutputs` and
`ChangeValue.isEphemeral`. libzcashlc keeps encoding proposals with the pinned
crate schema, so every field the copy lacks is one Swift cannot read.

## What this adds

- **`Scripts/update-proposal-proto.sh`** re-derives the file, resolving the
  upstream ref from `Cargo.lock` rather than `Cargo.toml` so it follows the
  *resolved* dependency: a crates.io release maps to the `<crate>-<version>`
  tag, a git dependency to the exact revision Cargo recorded.
- **`--check`** re-derives to a temp file and diffs, printing the offending
  hunks and failing on drift. A new `Proto Sync` workflow runs it.
- **A `DERIVED FILE` banner** naming the upstream path, ref and script,
  matching what `Scripts/update-lightwallet-protocol.sh` writes for the
  lightwallet protos.

The sole deviation from upstream is the injected `option swift_prefix`, the
same shape as the `java_package` injection on the Android side.

## Scope: the .proto body is deliberately NOT updated here

`proposal.pb.swift` is committed and must be regenerated with the
`protoc-gen-swift` version pinned in `Package.resolved`. Updating the `.proto`
without regenerating it would leave the two out of step, which is worse than
the current state. I did not have that toolchain available (`nix` is not
installed locally), so rather than commit codegen I could not verify, this PR
adds only the mechanism plus the banner.

**Consequence: the `Proto Sync` check is red on this branch.** That is the
check correctly reporting the real, pre-existing drift. The follow-up is to run
the script, regenerate `proposal.pb.swift`, and commit both; the check then
goes green.

## Verification

- `shellcheck` clean.
- `--check` fails with exit 1, resolving the ref to
  `zcash_client_backend-0.24.0-rc.6` from `Cargo.lock` and printing exactly the
  five missing items.
- Re-deriving and stripping the banner produces a file byte-identical to
  upstream with `option swift_prefix` injected, and `protoc` compiles it.
- The workflow YAML parses and the job registers.
- The workflow follows the repo zizmor conventions (SHA-pinned action with a
  version comment, workflow-level `permissions: {}`, job-scoped
  `contents: read`, `persist-credentials: false`, no `${{ }}` inside `run`).
  I could not run `zizmor` locally to confirm, as it is not installed.

Based on `maint/v2.7.x`, matching the Android PRs.


Chlup (approved):
utACK

PR #1916, opened 2026-07-30
Expand the Makefile and route the CI build jobs through it

dannywillems:
The root Makefile held five convenience targets while CI repeated its own
swift and FFI commands inline, so reproducing a CI failure locally meant
reading the workflow and retyping the steps, and the two drifted. This expands
the Makefile to own every command the project needs and has the workflows call
the targets.

Mirrors zcash-android-wallet-sdk#2094. Independent of, and not stacked on, the
`proposal.proto` PR (#1915).

## Nine steps now run a target

- `swift.yml`: `ffi-macos`, `verify-ffi`, `configure-local-ffi`, `build`,
  `test-offline`
- `codeql.yml`: `ffi-macos`, `configure-local-ffi`, `resolve`, `build`

Each expands to exactly what the step ran before. The verbose build is
preserved via `make build SWIFT_BUILD_FLAGS=-v`, so CI keeps its full log while
a local `make build` stays quiet.

## Compatible with what was there

Every previous target name still works; `swift-build` is now an alias for
`build`. The additions are the standard set (`build`, `test`, `check`, `clean`,
`setup`, `info`) plus the FFI graph the workflows needed: `ffi-macos`
reproduces the CI step of building the macOS slice and assembling the
XCFramework around it, and `configure-local-ffi` points `Package.swift` at the
result. `BuildSupport/Makefile` remains the FFI build graph; the root targets
delegate to it with `$(MAKE) -C`.

## What is deliberately not converted

SwiftLint, zizmor and the CodeQL analysis step are GitHub Actions rather than
commands, so converting them would change what they do. A local `make lint` is
provided for SwiftLint. The FFI cache-key hashing and the release workflow
(`build-ffi.yml`, which already calls `Scripts/prepare-release.sh`) are also
left alone.

## Portability

Targets GNU Make **3.81**, the version macOS ships, so none of the 4.x-only
features are used. The `build` and `test` targets work anywhere SwiftPM does;
the `ffi-*` targets produce an Apple XCFramework, so they check for macOS via
`require-macos` and say so, rather than failing obscurely partway through a
cargo build.

## Verification

- 0 lines over 80 chars; parses under GNU Make 3.81; all targets register.
- `make info` ran for real and correctly reported Swift 6.2.3, cargo, the Xcode
  path, and that no XCFramework is built yet.
- `make help` renders; `require-macos` passes on macOS.
- Dry-ran each converted target and compared against the original step, e.g.
  `make build SWIFT_BUILD_FLAGS=-v` -> `swift build -v`, `test-offline` ->
  `swift test --filter OfflineTests`, `ffi-macos` -> `make -C BuildSupport
  macos` followed by the same three assembly commands.
- All five workflow files still parse with their jobs intact.

I did not run a full Xcode or FFI build, only verified the commands issued.

Based on `maint/v2.7.x`, matching the Android PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,39 +1,200 @@
-# Convenience development targets for this repository. Run from the repo root.
-# See BuildSupport/Makefile for the full FFI build graph.
+# Makefile for the Zcash Swift wallet SDK. Run from the repo root.
+#
+# Every command the project needs goes through a target here, so that local
+# runs and CI runs are identical: .github/workflows/ invokes these targets
+# rather than repeating the commands.
+#
+# Run `make help` for the target list, and `make info` for the resolved
+# toolchain.
+#
+# Layout
+# ------
+# The Rust FFI build graph lives in BuildSupport/Makefile; the ffi-* targets
+# here delegate to it and then assemble the XCFramework exactly as CI does.
+# The Swift package itself is built with SwiftPM from the repo root.
+#
+# Portability
+# -----------
+# Written for GNU Make 3.81, which is what macOS ships, so it avoids the
+# 4.x-only features (`!=`, `$(file ...)`, `.ONESHELL`). The swift build and
+# test targets work anywhere SwiftPM does; the ffi-* targets produce an Apple
+# XCFramework and therefore require macOS with Xcode, which they check for
+# rather than failing obscurely.
+
+SWIFT ?= swift
+CARGO ?= cargo
+BUILD_SUPPORT := BuildSupport
+SCRIPTS := Scripts
+
+# The XCFramework the Swift package links against when built locally.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,39 +1,200 @@
...
+
+.PHONY: clean
+clean: ## Clean the SwiftPM build directory
+	$(SWIFT) package clean
 
-.PHONY: help init-ffi rebuild-ffi reset-ffi swift-build test-offline
+.PHONY: clean-all
+clean-all: clean clean-ffi ## Clean SwiftPM and FFI build artifacts
 
-help:
```
dannywillems, line 10: See the new implementation of help that builds the list at runtime.

Chlup (approved):
utACK

PR #1917, opened 2026-07-30
AGENTS.md: only v_transactions and v_tx_outputs may be read directly

dannywillems:
Stacked on #1919. Mirrors zcash/zcash-android-wallet-sdk#2095.

`dataDb` is owned by Rust. `zcash_client_sqlite` treats its `v_*` views as the
interface and its tables as an implementation detail, so a migration that keeps
a view's columns stable is still free to rename, split or drop the tables
beneath it. A Swift query naming a table therefore breaks on a dependency bump
with **no compile error and no test failure** until it runs against an upgraded
wallet.

## The rule

Swift may read `v_transactions` and `v_tx_outputs` directly. Every other query
goes through the FFI, and nothing in Swift writes to the database at all.

The other `v_*` views upstream defines serve Rust's own scanning and
note-commitment machinery rather than this SDK, so they are treated like tables.

## Where direct access lives

All of it is under `Sources/ZcashLightClientKit/DAO/`:

| File | Reads | Status |
|---|---|---|
| `TransactionDao.swift` | `v_transactions`, `v_tx_outputs` | views, fine |
| `BlockDao.swift` | `blocks` | table, pre-existing exception |

I surveyed rather than assumed: a repo-wide search for `Table("` / `View("`
returns only those two files. `PagedTransactionDao.swift` and
`UnspentTransactionOutputDao.swift` name no entities. Re-run against the rebased
base, with the same result.

The exception is documented as an exception, not endorsed; removing it needs an
FFI accessor for block-lookup-by-height that does not exist yet.

## Placement

Rebased onto the reconciled `AGENTS.md` in #1919. The section now sits directly
after **Security-critical API rules**, since both are MUST rules about the code
rather than the workflow. The earlier revision placed it between *Working
documents are not committed* and *Other notes*, two sections #1919 removed as
duplicates of `main`'s conventions list. The section text is unchanged and still
+51 lines.

## Also noticed, not fixed here

`BlockDao.swift` names the table twice with different capitalisation, `:28`
`Table("blocks")` and `:38` `Table("Blocks")`. SQLite ASCII identifiers are
case-insensitive so both resolve and it is harmless today, but it is two
sources of truth for one name, and the instance property shadows the static
one. Worth collapsing when this file moves to the FFI.

Docs only, so no CHANGELOG entry and no behaviour change.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -117,6 +117,57 @@ Every public API parameter that carries a domain value — key material, seeds, 
 - When logging errors, log `error.message` or `localizedDescription` (generic, code-prefixed, safe). NEVER log errors via `String(describing:)`, `"\(error)"`, or `dump(...)` — default reflection prints the associated values, and that output ends up in logs users mail to support.
 - New error cases added to `Error/ZcashErrorCodeDefinition.swift` must keep raw input out of their `message` text; carry context via the error code, not the offending value.
 
+## Database access: views only, everything else through the FFI
+
+`dataDb` is owned by Rust. `zcash_client_sqlite` defines both its tables and a
+set of `v_*` views, and only the views are a supported interface. The tables
+are an implementation detail that upstream reshapes freely, and a schema
+migration that leaves a view's columns intact can still rename, split or drop
+the tables underneath it.
+
+**Swift may read `v_transactions` and `v_tx_outputs` directly. Every other
+query goes through the FFI.** Never read a table from Swift, and never write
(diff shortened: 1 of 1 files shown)


nuttycom (approved):

dannywillems:
I'll rebase after https://github.com/zcash/zcash-swift-wallet-sdk/pull/1927 is merged, as 1927 also fixes the CI.

PR #1918, opened 2026-07-30
Delete CLAUDE.md

dannywillems:
Stacked on #1917.

`AGENTS.md` is the tool-neutral convention and, since the backport in #1919,
carries the guidance this repository expects an agent to follow. Keeping a
second file named for one tool invites the two to drift, so this removes it.

Deletion only: `CLAUDE.md +0/-106`, no other file touched.

## What this removes

`CLAUDE.md` had six sections, and **none of them is covered by the backported
`AGENTS.md`**: Project overview, Build and test, Rust FFI development, Release,
Architecture, and Conventions and gotchas. I checked each rather than assuming
overlap.

#1920 would have folded that content into `AGENTS.md`; it was closed
deliberately, so merging this drops the content rather than relocating it.
Flagging it here so the tradeoff is visible in the PR rather than only in
review comments.


AGENTS.md
```diff
@@ -2,8 +2,97 @@
 
 Guidance for AI coding agents (and anyone else) working in this repository.
 
-See `CLAUDE.md` for the build, test, FFI and architecture overview. This file
-holds rules that must not be broken; where the two overlap, this one wins.
+## Project overview
+
+`ZcashLightClientKit` is an iOS/macOS Swift Package that implements a Zcash lightwallet client. The Swift layer wraps a Rust core (in `rust/`) via an `libzcashlc` XCFramework. Most day-to-day SDK work happens in Swift only — SPM auto-downloads a pre-built XCFramework from GitHub Releases.
```
dannywillems: I would like to use the Makefile targets after.

Chlup (changes requested):
I am not sure if we can do this. AFAIK claude code doesn't read AGENTS.md. So either CLAUDE.md must be file with `@AGENTS.md` content or it should be at least link to AGENTS.md.

dannywillems:
> I am not sure if we can do this. AFAIK claude code doesn't read AGENTS.md. So either CLAUDE.md must be file with `@AGENTS.md` content or it should be at least link to AGENTS.md.

Oh, nice catch! I didn't know!

dannywillems:
Closing to mimick main.

PR #1919, opened 2026-07-30
Backport AGENTS.md from main, with CLAUDE.md as a symlink

dannywillems:
`maint/v2.7.x` had no `AGENTS.md`, and a standalone `CLAUDE.md`. This adds the
first and makes the second a symlink to it, matching `main`.

Per @pacu's review: the first revision of this PR took the file from
`feature/ironwood-slipstream`, which was written before #1832 landed and shares
no section with the `AGENTS.md` that #1832 put on `main`. Its opening paragraph
also points at `CLAUDE.md` as a separate document. Merging it would have put a
second, contradictory convention on this line. Reconciled onto `main`'s file
instead.

## What this is

`main`'s `AGENTS.md` verbatim, including the **Security-critical API rules**
(semantic types; key material in errors and logs) that the earlier revision
lacked entirely, plus four deltas:

1. `init-local-ffi.sh` documented with `--macos-only`. `main` describes
   `--arm-macos` / `--arm-ios` / `--arm-all`; the script on this branch
   (`Scripts/init-local-ffi.sh:3-5`) has only `--cached` and `--macos-only`.
2. **Pre-push validation**, **Worktree layout** and **CHANGELOG discipline**
   appended from `feature/ironwood-slipstream`. `main` has no equivalent.
3. Two facts from that branch folded into sections `main` already has rather
   than repeated: release artifacts never come from a PR build, and generated
   output is checked in so a hand edit is silently reverted by the next
   regeneration.
4. The `Breaking API changes` bullet dropped, because the CHANGELOG discipline
   section states the same `MIGRATING.md` rule in full.

`diff` against `main`'s file is exactly those four and nothing else.

## What was dropped, and where it went

Four sections of the earlier revision were duplicates. Each fact was checked to
have a home before removing it:

| Dropped | Covered by |
|---|---|
| Commit and PR conventions | `main`'s Conventions and gotchas bullets |
| Working documents are not committed | the `Design docs and plans` bullet |
| Other notes | Conventions bullets, and Architecture for the async-adapter note |
| Regenerate, never hand-edit | `### Generated code` |
| What cannot be run locally, or at all, in CI | `## Build and test` and `## Release` |

## Corrected against this branch

The pre-push section came from a branch predating #1915 and #1916, so it was
rewritten against what CI here actually does:

- Local equivalents are `make` targets, because `swift.yml` now drives every
  build and test step through the `Makefile`.
- `proto-sync.yml` added as the fourth PR gate, and the vendored
  `proposal.proto` marked script-generated in the networking section.
- `codeql.yml` corrected: merges to `main` and weekly, **not** per PR.
- The `paths-ignore` list enumerated. `AGENTS.md` is not on it, so a change to
  this file alone still triggers a full build, as this PR demonstrates.

Every `make` target, workflow and script named in the file was verified to
exist on this branch.

## Also here

- `CLAUDE.md` becomes a symlink; the blob is `47dc3e3`, identical to `main`'s.
- `.gitignore` gains `.plans/` and `.worktrees/`, which `main` and
  `feature/ironwood-slipstream` both already ignore and which the new sections
  assume are ignored.

#1917 is stacked on this. #1918 (delete `CLAUDE.md`) is closed: the symlink
makes it unnecessary and it would have contradicted #1832.


AGENTS.md
pacu, line 1: Please check https://github.com/zcash/zcash-swift-wallet-sdk/pull/1832 and reconcile

pacu (changes requested):

AGENTS.md
dannywillems: This match now main, but `maint/v2.8.x` will conflict.
As we merge `maint/v2.7.x` into `maint/v2.8.x`, I suggest we solve the conflicts with `maint/v2.8.x` later.

pacu (approved):

PR #1922, opened 2026-07-30
Surface the v_transactions columns the SDK was dropping

dannywillems:
Mirrors zcash/zcash-android-wallet-sdk#2098.

`v_transactions` carries three columns the SDK never read.

## What is added

| Column | Exposed as | |
|---|---|---|
| `spent_note_count` | `spentNoteCount: Int` | notes of the account this tx spent |
| `pool_crossing_value` | `poolCrossingValue: Zatoshi?` | see below |
| `trust_status` | `isTrusted: Bool` | outputs spendable at the trusted confirmation count |

`pool_crossing_value` is non-NULL **exactly** when a transaction is a
wallet-internal transfer between shielded pools, such as an Orchard to Ironwood
migration, and its value is the amount that crossed. For such a transaction the
balance delta is just the negated fee, which is why a migration currently
displays as a send of the fee rather than of the migrated amount. Upstream says
so directly: "this is the amount to present to a user rather than the balance
delta".

## A separate bug found on the way

`ZcashTransaction.Output.Pool` had no `ironwood` case, so pool `4` fell through
to `.other(4)` and an Ironwood output was reported as an unknown pool. Since
NU6.3 every payment to an Orchard receiver is delivered in the Ironwood bundle,
so this affected **ordinary received payments**, not only migrations. Added.

## Not included, deliberately

`v_tx_outputs` also has `tx_mined_height`, `tx_trust_status` and
`recipient_key_scope` that Swift does not model. I left them out rather than
add properties nothing reads. `transaction_id` is excluded on principle:
upstream documents it as a database-internal join key that "should not ever be
exposed to end-users".

## Verification

`swift build` clean, and the full `OfflineTests` suite passes: **326 tests, 0
failures**. The compiler found every fixture needing the new arguments, across
five test files, which is what proves the plumbing is complete rather than
merely declared.

### DIFF
--- a/Sources/ZcashLightClientKit/Entity/TransactionEntity.swift
+++ b/Sources/ZcashLightClientKit/Entity/TransactionEntity.swift
@@ -63,6 +63,19 @@ public enum ZcashTransaction {
         public let isExpiredUmined: Bool?
         public let totalSpent: Zatoshi?
         public let totalReceived: Zatoshi?
+        /// Number of the account's own notes this transaction spent.
+        public let spentNoteCount: Int
+        /// The value that crossed shielded pools when this transaction is a
+        /// wallet-internal transfer between them, such as an Orchard to
+        /// Ironwood migration; `nil` when it is not such a transfer.
+        ///
+        /// For such a transaction `value` is just the negated fee, so this is
+        /// the amount to present to a user rather than the balance delta.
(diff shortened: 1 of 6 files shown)


Chlup (approved):
utACK
LGTM

PR #1923, opened 2026-07-30
Add support for latest librustzcash updates (adopt the upstream state machine, reachable prove sweep, estimated tip, invalidation reconciliation)

Chlup:
## Summary

Minimal-change parity-and-fixes pass over the Orchard → Ironwood migration layer, replacing SDK-invented machinery with upstream `zcash_pool_migration 0.1.0-rc.5` calls wherever upstream provides the capability, and closing behavior-parity gaps against the Android SDK.

### Fixes
- **The prove sweep is now reachable**: `finalizeReadyMigrationTransfers` is public, and `executeNextPendingMigrationTransfer` returns a `nothingDue` / `awaitingProof(id)` / `executed(result)` attempt enum and is strictly broadcast-only. Previously the sweep had zero call sites, so gradual migrations stalled at awaiting-proof forever.
- **Keystone signing-session math**: the count-based estimate (wrong for mixed runs — 6 preparations + 1 transfer is 99 actions = 2 rounds under the 96-action budget) is replaced by upstream's action-weighted `MinRounds`, plus real `NextFit` PCZT batching (`batchMigrationPcztsForSigning`, `MigrationSigningBudget.keystone`).
- **Transient prove-error classification** aligned with Android (`UnknownSpentNote` and shard-tree query races transient; `IronwoodTreeUnavailable` hard).
- **Privacy sync buffer** network-scaled: mainnet 600 s, testnet/regtest 180 s.

### Upstream state machine (breaking)
The SDK-derived 5-state `MigrationState` is deleted. `migrationAdvanceStep(accountUUID:)` is a **verbatim conduit** of upstream `MigrationState::next_step` — `nil` for no stored run, `.prove(id:kind:)` / `.broadcast(id:)` / `.rebuild(id:)` / `.waiting` / `.complete`, broadcast-first per librustzcash PR zcash/librustzcash#2867 (pre-adopted via two small shims deleted at the next crate bump; the public contract is identical before and after). Also newly wired from upstream: `migrationSyncWakeups` (sync-and-prove wake heights), per-status `dependsOn` / `anchorBoundaryHeight`, and schedule `preparations`. `MIGRATING.md` carries the full replacement mapping for the removed states.

### Android-parity additions
- `reconcileMigrationInvalidations`: the 3-pass reconcile (mined promotion, submit-crash probe with a persisted 120 s in-flight guard, foreign-spend nullifier check) — with the nullifier extraction fixed to scan **all** Orchard actions, so it actually works on production 2-action transfers.
- Estimated chain tip: measured block-rate estimator, `estimatedMigrationChainTip` / `estimatedMigrationSecondsPerBlock`, and `useEstimatedTip:` on the due-ness members (estimates accelerate due-ness only; expiry and invalidity always evaluate on the scanned tip).

### Cleanup
Deletes the orphaned engine-immediate FFI (the unreachable `zcashlc_migration_propose_immediate_transfers` — see #1894), renames the read-only `rescheduleOverdueMigrationTransfer` to `pendingMigrationTransferProposal`, fixes stale ZIP 318 timing docs, and pins `zcash_pool_migration = "0.1.0-rc.5"`.

## Verification
- `cargo test`: 145 passed, 0 failed; `cargo fmt` clean.
- `swift test --filter OfflineTests`: **732 passed, 0 failures** (the CI suite), re-verified against a full 5-architecture `init-local-ffi.sh` build.
- No new SwiftLint findings.

Breaking changes are documented in `MIGRATING.md`; user-visible changes in `CHANGELOG.md` and `rust/CHANGELOG.md`.


Chlup:
Review-fix pass pushed (4 commits). Highlights:

- **Root-cause fix upstream**: the engine now owns failure states and estimate-aware due-ness — zcash/librustzcash#2873 — and this branch pins the librustzcash family to that rev (re-pin to crates.io at the next release). The SDK's marks side-table, both #2867 pre-adoption shims, and the four dual-target "twin" predicates are deleted; the advance-step conduit is verbatim, now including the engine's native attend-first step (`.requiresAttention`).
- **Sync-gate policy corrected**: the gate blocks on a READY broadcast (proved, due, unexpired, valid — new `has_ready_broadcast` predicate), never on due-but-unproved rows, dissolving the review's interlock/deadlock cluster (findings 1, 2, 4, 5). Disclosed in MIGRATING.md.
- All 15 reported findings and the 14 unreported ones addressed (reconcile rewind-evidence + single-parse memoization, `Tree(Query(_))` narrowing with logged deferrals, trapping-conversion guards, in-flight marker re-arm/clear/clock-clamp, gate-file blind-spot fallback, schedule `encode(to:)` handle omission, saturating heights, leak-order hoist, account-free estimate members, shared injected-clock estimator, and the rest).

Verification: librustzcash patch — 180 engine + 122 store tests, clippy/doc clean; SDK — `cargo test` 158/158, `swift test --filter OfflineTests` **761/761**, full 5-architecture FFI build green.

nuttycom (dismissed):
Requesting changes to block this PR on the resolution of its upstream dependency: zcash/librustzcash#2873, whose failure-state half is being superseded by zcash/librustzcash#2871 (see the analysis at https://github.com/zcash/librustzcash/pull/2873#issuecomment-5143402057). This PR's `[patch.crates-io]` pin targets the #2873 branch, which will be closed rather than merged, so the pin — and the three API surfaces adopted on top of it — will need to re-target.

Direction for the re-adoption, so work here can be planned (a fuller handoff document adapting this PR's approach to the #2871 API will follow once that branch stabilizes):

- **The invalid-marks side-table fold becomes unnecessary rather than re-targeted.** #2871's engine detects a funding note spent outside the migration *itself*, from the wallet's own note/nullifier tables, via a store-side satisfiability oracle consulted before any step is surfaced — no consumer report, no `mark_invalid`. Existing side-table rows for spent-funding cases can simply be dropped; the engine re-derives the fact on the first drive call after sync.
- **Network-rejection reporting maps to a report/reevaluate channel, not a state mark.** #2871 deliberately does not let a node's rejection classification write a failure state (a rejection is testimony from another observer, not evidence in the wallet's scanned view — the same-seed double-spend case makes the store actively contradict it until the next sync). The replacement, landing in #2871 shortly: the app reports the failure plus the tip it observed (`report_broadcast_failure`); the engine withholds the transaction, surfaces an unconditional `Reevaluate` step instructing sync-on-next-wake, and adjudicates through its own oracle once the scanned height reaches the reported tip — marking on now-visible evidence, or clearing the report and re-offering if the rejection was transient.
- **The `Attend` conduit maps to `Replan` + statuses.** #2871 surfaces per-transaction death as a `TransactionStatus` blocker (with a persisted *kind* explaining why), and run-level remediation as a threshold-gated `AdvanceStep::Replan` — live work keeps draining rather than the whole run stalling behind an attention step.
- **The `has_ready_broadcast` gate predicate maps to `transaction_statuses`** (`ready && action == Broadcast`) — the queue queries it was built on are internal in #2871, whose public drive API verifies every step against the oracle before surfacing it.
- **`DuenessTargets` / estimate-aware due-ness is the one piece of #2873 that survives on its own merits** — it will arrive as a follow-up PR on top of #2871 (reshaped to thread the estimate through the drive API's config rather than public queue queries). The deletion of this SDK's split twins should wait for that follow-up rather than re-implementing against #2871's internals.

*Prepared with Claude Code assistance.*

nuttycom:
# Handoff — adopting zcash/librustzcash#2871 (unsatisfiable-step detection) in the mobile SDKs

**Date:** 2026-07-31
**Upstream PR:** https://github.com/zcash/librustzcash/pull/2871 (open, ready for review; branch `feat/migration_unsatisfiability` @ `92c5321272`, rebased on `main` @ `25354d417a`)
**Supersedes:** PR zcash/librustzcash#2873 (`michal/pool-migration-failure-states`) — see the subsumption analysis in the comments on zcash/librustzcash#2873.
**Audience:** the Swift SDK work in zcash/zcash-swift-wallet-sdk#1923 (and, identically, a future Kotlin adoption). This document maps zcash/librustzcash#2873's surface — which #1923 currently pins — onto zcash/librustzcash#2871's, so the SDK wrapper can be re-adopted with the same deletions it already made.

---

## Part 1 — What changed at the root, and why the shape differs from zcash/librustzcash#2873

Both PRs close the same two consumer gaps (unreportable event-based transaction death; the drive loop issuing undischargeable instructions). zcash/librustzcash#2871 closes them with a different division of labor, and that division is the main thing to internalize before re-adopting:

**The engine now discovers most death itself.** zcash/librustzcash#2873 asked the consumer to *diagnose* (`mark_invalid(id, FundingSpent | RejectedInvalid | RejectedExpired)`). zcash/librustzcash#2871 instead gives the store a satisfiability **oracle** (`PoolMigrationRead::check_step_satisfiability`) that answers from the wallet's own scan database — nullifier evidence, expiry, anchor displacement — and the public drive API adjudicates with it. A funding note spent by a same-seed wallet is found by the engine *from scan data*, at a stamped observation height, without the SDK saying anything. Because `zcash_client_sqlite` implements the oracle, the SDKs get this for free through the existing store.

**The consumer reports only what only it can know: a node's rejection.** The one verb is

```rust
MigrationState::report_broadcast_failure(id: MigrationTransferId, observed_tip: BlockHeight)
```

and it records **testimony, not a verdict** — no mark is written from a report. The engine withholds the reported transaction from the broadcast queue and, from `advance_migration`, returns the new `AdvanceStep::Reevaluate` until the wallet has scanned to at least `observed_tip`. Then it adjudicates: if the scan now shows the inputs spent/invalidated, the transaction is marked unsatisfiable (with dependency-closure propagation and, past a committed threshold, `AdvanceStep::Replan`); if the scan shows nothing wrong, the report is cleared and the broadcast is re-offered in the same call. A still-real failure will simply be re-reported by the app on the next rejection — that loop is the intended discipline, and it is what makes a *wrong* or stale node verdict self-healing.

Why this shape: a rejection can be raced (the same seed's spend may not be scanned yet), and a consumer-chosen reason can be wrong. Evidence-stamped marks derived from scanned data survive reorg truncation correctly (`truncate_to_height` clears exactly the marks whose evidence is above the truncation height); consumer verdicts cannot carry that property. This is also why `RejectedExpired` has no analogue — expiry is already the engine's own height-derived judgment, and the rebuild path handles it.

## Part 2 — Surface mapping (zcash/librustzcash#2873 → zcash/librustzcash#2871)

| zcash/librustzcash#2873 (pinned by #1923) | zcash/librustzcash#2871 |
|---|---|
| `MigrationState::mark_invalid(id, reason)` | `MigrationState::report_broadcast_failure(id, observed_tip)` — call it whenever a broadcast attempt is rejected by the node; `observed_tip` is the chain tip the app observed at rejection time (the node's height, not the wallet's scanned height). Records iff the row is `Proved`; re-reporting overwrites; no-op otherwise. |
| `InvalidReason { FundingSpent, RejectedInvalid, RejectedExpired }` | `UnsatisfiableKind { InputsSpent, InputsInvalidated, AnchorInvalidated, Inherited }` (`#[non_exhaustive]`, snake_case wire codec) — produced by the engine's adjudication, never supplied by the caller. `Inherited` = dead only through a dependency. |
| `MigrationTxState::Invalid { reason }` (a state variant) | An **orthogonal persisted mark**: `MigrationTransaction::unsatisfiable() -> Option<(BlockHeight, UnsatisfiableKind)>` (also projected as `unsatisfiable_at()` / `unsatisfiable_kind()`). Lifecycle state is unchanged — a marked row keeps its `Proved`/`Broadcast` state, and the mark clears under reorg truncation while the state survives. |
| `AdvanceStep::Attend { id }` | Two variants with sharper semantics: `AdvanceStep::Reevaluate` ("sync to at least the reported tip, then call `advance_migration` again" — surfaced ahead of all actionable work, after terminal `Complete`) and `AdvanceStep::Replan` (enough value is dead that the migration plan must be rebuilt). Per-transaction attention detail lives in `transaction_statuses`. |
| `Blocker::Invalid` | `Blocker::Unsatisfiable` (with the kind exposed on the status; `Inherited` when the row is dead only via a dependency) and `Blocker::AwaitingReevaluation` (an open report). Precedence: `Unsatisfiable` > `AwaitingReevaluation` > `Expired` > `ExpiryImminent` > schedule/dependency blockers. |
| `mark_mined` overwrites `Invalid` | Same principle, stronger: `mark_mined` clears the unsatisfiability mark **and** any open broadcast-failure report — chain inclusion outranks every prior verdict. |
| `RebuildError::Invalid(id, reason)` | `RebuildError::Unsatisfiable(MigrationTransferId)` — rebuild refuses a marked transfer (or one with a marked direct dependency); the remedy is the migration-level replan, not a fresh single-transaction rebuild. |
| verbatim `next_step` conduit | `next_step` is no longer public. The conduit target is the public drive API: `satisfiability::advance_migration(&mut store, &mut state, targets: DuenessTargets, config: &AdvanceConfig) -> Result<AdvanceStep, St::Error>` (`satisfiability::AdvanceConfig::new(reorg_settle_depth)`). It runs the oracle sweep, adjudication, persistence, and step selection in one call; every determination it makes is persisted before it returns. `ReorgSettleDepth::new(10)` is the recommended default. |
| `DuenessTargets` / `next_provable_at` / `next_broadcastable_at` | **In zcash/librustzcash#2871** (landed in `7fbb16011e`, below the current pin). ZIP 318 forbids contacting lightwalletd before the next-step check, so a wall-clock estimate *must* be usable at drive time: `satisfiability::DuenessTargets` carries both heights (accessors `scanned()` / `effective()`; the fields are private), in the crate's `tip + 1` target convention — `DuenessTargets::new(scanned, estimated)` clamps `effective = max(scanned, estimated)`, and `DuenessTargets::at(target)` is the degenerate pair for a caller whose estimate *is* its chain view. `scanned` governs every durable or destructive judgment (expiry as a determination, dead-set seeding, `expired_transactions`/`Rebuild` eligibility, the drain-time `Replan` gate, `record_satisfiability`'s dependency closure, anchor-boundary settledness, the statuses `Expired` blocker); `effective` governs only re-ordering and protective withholding (the prove and broadcast schedules, the doomed-broadcast withhold, the prove-side expiry skip). A transfer in the doomed window (`scanned <= expiry_height < effective`) reports the new `Blocker::ExpiryImminent` — precedence `Unsatisfiable` > `AwaitingReevaluation` > `Expired` > `ExpiryImminent` > schedule/state — and is neither `ready` nor carrying an `action`, so the statuses-based sync gate (a broadcast session is warranted iff some status has `ready && action == Some(NextAction::Broadcast)`) now respects the kernel's withhold automatically, with no SDK-side estimate handling. |
| new sqlite migration `orchard_ironwood_migration_invalid_reason` | One (still-unshipped) migration `orchard_ironwood_migration_unsatisfiability`, adding `unsatisfiable_at`, `unsatisfiable_kind`, `broadcast_failure_at` to the transactions table and `replan_threshold` to the migrations table, plus the new child table `orchard_ironwood_migration_spend_nullifiers` (one 32-byte nullifier per row, ordered by `ordinal`) and its backfill from the stored PCZTs of in-flight rows. Unshipped means: a dev database that ran an earlier draft of it must be recreated. |

Also new, no zcash/librustzcash#2873 analogue: `MigrationState::mark_superseded()` (terminal status for a replanned-away migration), `replan_required()` (the committed dead-value threshold, default 20%), and the typed prove channel (`ProveOutcome` / `ProveFailure::InputNotAvailable { nullifier, as_of }`) the wallet adapter uses to report prove-time unavailability as evidence. That channel's error type also tightened: `WalletProveError::{NoRealSpend, MalformedNullifier}` are gone, replaced by the single `WalletProveError::RealSpends(RealSpendError)` carried from the now-public `zcash_pool_migration::pczt_spends` module (`orchard` feature), whose `real_spend_nullifiers(&Pczt) -> Result<Vec<(usize, orchard::note::Nullifier)>, RealSpendError>` states the real-spend rule once — so a store or wrapper reconstructing the nullifier cache from stored PCZT bytes reuses it rather than restating it, and a corrupted PCZT now surfaces `MalformedNullifier` / `NoRealSpends` instead of silently reading as "this transaction spends nothing".

## Part 3 — Adoption notes

1. **Pin.** Re-pin the `[patch.crates-io]` family to the zcash/librustzcash#2871 head commit hash (`92c5321272` today; expect amendments while the PR is in review — a hash pin fails loudly on rebase, which is what you want). Re-pin to crates.io at the first release containing the PR.
2. **The invalid-marks side-table migration changes meaning.** #1923's one-time migration folded side-table rows into `mark_invalid`. Under zcash/librustzcash#2871: drop the `FundingSpent`-shaped rows entirely — the oracle rediscovers them from scan data on the first `advance_migration` call, with correct evidence stamping (this is strictly better than replaying a stored verdict). For rejection-shaped rows, replay each as `report_broadcast_failure(id, tip)` using the wallet's current scanned height as `tip` if the original observation height was not stored; the next advance will adjudicate immediately.
3. **The conduit stays verbatim.** `.requiresAttention`-style surfacing maps from `Reevaluate` (migration-level: sync, then re-advance) plus statuses (per transaction: `Blocker::AwaitingReevaluation`, or `Blocker::Unsatisfiable` read together with the sibling `unsatisfiable_kind` field — the blocker carries no payload). Nothing SDK-private is needed for precedence — the engine orders blockers and steps itself.
4. **Report on every rejection.** The engine tolerates (and expects) re-reports; do not debounce them SDK-side. Do not report scan-discovered spends at all — that is the oracle's job.
5. **Dueness discipline.** Construct `DuenessTargets::new(scanned_target, estimated_target)` at each wake-up: `scanned_target` is the wallet's `block_fully_scanned` height + 1, `estimated_target` its wall-clock estimate of the tip + 1. Both are available without touching the network, which is the point — the drive decision precedes any lightwalletd contact. Do not try to keep the estimate out of the engine: the engine anchors every destructive judgment to `scanned` internally, and the constructor's clamp (`effective >= scanned`) means an estimate that lags the wallet's own observations can never withhold work the chain data already justifies.

## Status and risk

- zcash/librustzcash#2871 is open and under review (reviewer: dannywillems); names and representations may shift, but the testimony/oracle division is the settled design. The wrapper-marshals-and-app-enum-stays-stable strategy from #1923 absorbs this the same way it absorbed zcash/librustzcash#2873 review risk.
- Estimate-aware scheduling is no longer a follow-up: the `DuenessTargets` split landed in zcash/librustzcash#2871 itself (`7fbb16011e`), after review established that ZIP 318 session separation makes a wall-clock estimate the only tip information available at drive time. An earlier revision of this document described the opposite contract (see the addendum comment on #1923); it is superseded by the row above.

*Prepared with Claude Code assistance.*

nuttycom:
Addendum correcting one point in [the handoff comment above](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1923#issuecomment-5144757923).

That comment said the dueness split was deliberately deferred out of zcash/librustzcash#2871 into a follow-up, and that the SDK must pass only a chain-view height into the engine, keeping its wall-clock estimate outside. The reverse is the case, and the split has now landed in zcash/librustzcash#2871 itself: ZIP 318 separates sync wake-ups from broadcast wake-ups, so no lightwalletd query may precede the next-step check, and a wall-clock estimate is the only tip information available at drive time. Estimates are therefore first-class: `advance_migration` takes `DuenessTargets::new(scanned_target, estimated_target)` — scanned from the wallet's fully-scanned height + 1, estimated from wall clock, no network needed — and anchors every destructive judgment (expiry determinations, dead-set seeding, rebuild eligibility, the durable dependency closure) to the scanned target internally, with the constructor clamping the effective target up so a transposed pair is harmless. A transaction in the doomed window surfaces the new `Blocker::ExpiryImminent`, is neither `ready` nor actionable, and so drops out of the `ready && action == Broadcast` sync-gate predicate automatically — no SDK-side estimate handling is needed.

New pin: see the [main handoff comment](https://github.com/zcash/zcash-swift-wallet-sdk/pull/1923#issuecomment-5144757923), which tracks the current zcash/librustzcash#2871 head. The full corrected surface mapping rides in the updated handoff document.

nuttycom:
# ✅ Resolved — both items landed

*(This handoff has been executed. The original text is preserved verbatim below.)*

| | |
|---|---|
| **Item 1** (rename) | [`de4a7d84`](https://github.com/zcash/zcash-swift-wallet-sdk/commit/de4a7d84) — *Rename the Swift unrecorded-broadcast repair to match the engine* |
| **Item 2** (mocks + estimate coverage) | [`4d4fbd8f`](https://github.com/zcash/zcash-swift-wallet-sdk/commit/4d4fbd8f) — *Cover the advance step's tip estimate* |
| **Branch tip** | `4d4fbd8f`, pushed to `upstream/michal/migration-parity-fixes` |
| **Verification** | `swift build` clean; `swift test --filter OfflineTests` → **763 tests, 0 failures**. `de4a7d84` was verified in isolation (761 tests, 0 failures), so each commit builds on its own. |

The rename went in exactly as specified below, and `ZRUST0145` is unchanged — the sourcery `code=` annotation stayed attached to the renamed case, confirmed in the regenerated `ZcashErrorCode.swift`.

## Four things differed from the plan, all resolved

**1. The local FFI was stale and had to be rebuilt first.** `LocalPackages/` dated Jul 27, predating both Rust commits — its header carried neither `zcashlc_migration_reconcile_unrecorded_broadcasts` nor `zcashlc_migration_advance_step`, so nothing would have compiled. Fixed with `./Scripts/rebuild-local-ffi.sh macos`. **Only the macOS slice is fresh**; the iOS slices were preserved as-is and are still stale, so iOS work needs `init-local-ffi.sh --arm-all` first.

**2. Three of the "doc-comment references only" sites were stale in substance, not just in name** — a pure rename would have left them factually wrong:

- `Synchronizer.swift:622` — the `.requiresAttention` discharge still routed through this method; rewritten to "sync, advance again, and only then cancel and re-plan".
- `MigrationModels.swift:81` — `MigrationAdvanceStep.requiresAttention` described a "compatibility flow" through this method, contradicting its own type doc 40 lines above.
- `ZcashRustBackendWelding.swift:553` — `migrationHasInvalidTransfers` still credited this method with foreign-spend detection; corrected against the Rust doc on `zcashlc_migration_has_invalid_transfers`.

**3. The MIGRATING.md paragraph needed rewriting, not just trimming.** Its opening clause was "because its name no longer describes it" — the very thing the rename fixes. Reframed to keep the note's purpose (the method is narrower than a repair entry point sounds) without invoking a name that never shipped.

**4. Item 2's diagnosis was exactly right, and regenerating the mocks broke 4 tests across 3 files** — `SDKSynchronizerMigrationTests`, `SlipstreamSynchronizerMigrationTests`, and `OrchardMigrationHostTests` had all been stubbing the one-argument form and silently exercising the estimate-discarding default. Repointed at `migrationAdvanceStepForEstimatedTip*`.

## Notes on the added coverage

The two new tests live in `OrchardMigrationCompositionTests.swift`, mirroring the wiring already pinned for `executeNextPendingTransfer` and `isSyncBlocked`. The first pins the *exact* projected height rather than merely non-nil: one block-rate sample 150 s (two 75 s target-spacing blocks) before the frozen test clock projects to height + 2, a value neither a dropped estimate nor a wall-clock leak can produce. The second pins the other half of the gating contract — a block-rate-samples read that throws degrades to scanned-tip behaviour instead of failing the advance step.

Mutation-checked: passing `estimatedTip: nil` from `OrchardMigration.advanceStep()` fails the first test with `nil is not equal to Optional(3000002)`.

SwiftLint is not installed in this environment, so added lines were checked against `.swiftlint.yml` / `.swiftlint_tests.yml` by hand (150-char limit, `ignores_comments: true`). The only two over the limit were the lint-excluded generated `ZcashError.swift` and one test line, which was wrapped.

*Update written with Claude Code assistance.*

---

# Handoff — Swift-side work blocked on a macOS toolchain

**Branch:** `michal/migration-parity-fixes` @ `37ea1c0b`
**Blocked because:** the last two commits were made on Linux, where there is no Swift toolchain and no Sourcery. Everything below needs `swift build` / `swift test` and `sourcery 2.3.0` (the mock script hard-checks that exact version). The Rust side of both items is already done and its tests pass.

There are two independent items. The first is the substantive one; the second is a loose end from the commit before it.

---

## Item 1 — Rename `reconcileMigrationInvalidations`

### Why

`37ea1c0b` moved foreign-spend detection into the engine's satisfiability oracle and mined-transaction promotion into `advance_migration`. What is left of this method is one narrow repair: **a transaction this process submitted to a node, which mined, but whose broadcast was never recorded** — a crash, or a failed persist, between submitting and recording, leaving a `Proved` row whose transaction is already on chain. The engine's own promotion sweep keys on the recorded txid, so nothing else would ever promote it.

It records no invalid marks, and it is no longer part of the `.requiresAttention(id:)` discharge (that is now "sync, advance again, and only then cancel and re-plan"). The name is the last thing still describing the old behaviour. Its documentation is already corrected everywhere and marked as due to change; this is purely the rename.

The Rust entry point was already renamed in `37ea1c0b` and is `zcashlc_migration_reconcile_unrecorded_broadcasts`.

### Suggested names

Following the existing `<verb><qualifier>Migration<noun>` shape of `finalizeReadyMigrationTransfers` / `refreshStaleMigrationTransfers`:

| Layer | From | To |
|---|---|---|
| Public `Synchronizer` | `reconcileMigrationInvalidations(accountUUID:)` | `reconcileUnrecordedMigrationBroadcasts(accountUUID:)` |
| `OrchardMigration` actor | `reconcileInvalidations()` | `reconcileUnrecordedBroadcasts()` |
| Welding | `migrationReconcileInvalidations(for:)` | `migrationReconcileUnrecordedBroadcasts(for:)` |
| `ZcashError` case | `rustMigrationReconcileInvalidations` | `rustMigrationReconcileUnrecordedBroadcasts` |

The return type stays `Bool`; it now means "this call repaired a row".

### Sites

Declarations and call sites:

- `Sources/ZcashLightClientKit/Synchronizer.swift` — protocol requirement (`:693`) and the unimplemented-default extension (`:1329`)
- `Sources/ZcashLightClientKit/Synchronizer/SDKSynchronizer.swift:1225`
- `Sources/ZcashLightClientKit/Slipstream/SlipstreamSynchronizer.swift:1157`
- `Sources/ZcashLightClientKit/Migration/OrchardMigration.swift:318` (declaration), `:325` (welding call)
- `Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift:480`
- `Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift:1551` (declaration), `:1562-1563` (thrown error + fallback message string)
- `Sources/ZcashLightClientKit/Error/ZcashErrorCodeDefinition.swift:722-725`

Doc-comment references only:

- `Sources/ZcashLightClientKit/Synchronizer.swift:622`
- `Sources/ZcashLightClientKit/Model/MigrationModels.swift:38`, `:81`
- `Sources/ZcashLightClientKit/Migration/MigrationSyncGate.swift:23`, `:376`
- `Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift:479`, `:553`

Tests:

- `Tests/OfflineTests/SynchronizerMigrationDefaultsTests.swift:50`

### One thing to be careful about

In `ZcashErrorCodeDefinition.swift`, keep the `// sourcery: code="ZRUST0145"` annotation attached to the renamed case. The code is explicit rather than positional, so preserving the annotation keeps `ZRUST0145` stable across the rename — renaming the case while dropping or moving the annotation would silently renumber a published error code.

### Regenerate, then

```
./Sources/ZcashLightClientKit/Error/Sourcery/generateErrorCode.sh
./Tests/TestUtils/Sourcery/generateMocks.sh     # sourcery 2.3.0 exactly
```

Then drop the "name is due to change" notes, which exist only to cover this gap:

- `MIGRATING.md` — the `reconcileMigrationInvalidations` paragraph under **New members** ("Expect the name to change before release")
- `CHANGELOG.md` — the trailing sentence of that entry
- `Sources/ZcashLightClientKit/Synchronizer.swift` — the `- Note:` on the method
- `Sources/ZcashLightClientKit/Rust/ZcashRustBackendWelding.swift` — the sentence beginning "The Swift name predates that split"

`ClosureSynchronizer` / `CombineSynchronizer` do not expose the migration API, so no adapter changes.

---

## Item 2 — Regenerate mocks for the estimate-aware advance step (independent of the rename)

`2c391bb8` added `migrationAdvanceStep(for:estimatedTip:)` to `ZcashRustBackendWelding` with a protocol-extension default that forwards to the one-argument form, but **the mocks were never regenerated**. `ZcashRustBackendWeldingMock` therefore implements only `migrationAdvanceStep(for:)`, so every test silently exercises the default that *discards the estimate*. Nothing covers the estimate path, and a regression that dropped the estimate in production would pass CI.

Regenerating the mocks makes the new member mockable. Worth adding alongside it: a test that `OrchardMigration.advanceStep()` passes the gating tip estimate through, which is currently unassertable.

---

## Verification

```
swift build
swift test --filter OfflineTests
```

The Rust side needs nothing: `cargo test` in `rust/` is green at 138 passing on this branch.

---

*Written with Claude Code assistance.*

nuttycom:
## Adoption note: zcash/librustzcash#2892 (prove-time wallet persistence)

When this PR's `zcash_pool_migration` pin advances past zcash/librustzcash#2892 (currently on the `feat/migration_unsatisfiability` line), the prove seam changes shape. Summary of what the FFI layer and product code should expect:

### API changes to adopt

- `ProveOutcome::Proved` now carries a `#[must_use]` `ProvedTransaction` (the proven PCZT bytes + row id), and `prove_transfer` / `prove_preparation` no longer move the state to `Proved` themselves. The only discharge path is the new required `PoolMigrationWrite::store_proved_transaction`:

  ```rust
  match engine::prove_transfer(&params, &mut prover, &mut state, id, scanned_tip, &mut rng)? {
      ProveOutcome::Proved(proven) => {
          // The prover's wallet borrow must end before the store is used — the proof
          // travels as a value precisely so the two borrows stay sequential.
          store.store_proved_transaction(&mut state, proven)?;
      }
      outcome => {
          // NotYetProvable / MarkedUnsatisfiable: persist via replace_migration, as today.
          store.replace_migration(&state)?;
      }
  }
  ```

  For this PR that means the prove sweep behind `finalizeReadyMigrationTransfers` is the discharge site.

- `zcash_client_sqlite`'s store constructor gains wallet context: `PoolMigrations::for_account(params, clock, conn, account)` (pass the network parameters and `zcash_client_sqlite::util::SystemClock`). The `PoolMigrationWrite` impl requires real `Parameters`/`Clock` values; `()` is accepted only for read-only use.

### Behavioral changes product code should expect

- From the moment a transaction is proved, the wallet database holds its finalized record — raw transaction, fee, sent outputs, and input notes marked spent — written atomically with the migration store. Spendable balance therefore drops at *prove* time rather than at mine time, and the wallet's own spends can no longer consume a migration input during the (deliberately long) window before the scheduled broadcast. Any "available balance" or note-selection expectations in tests and UI should be checked against this.
- A broadcast-only session (`executeNextPendingMigrationTransfer`) no longer needs to extract the transaction from the stored PCZT: the raw transaction is available from the wallet DB by txid from prove time onward.
- Proving now surfaces extraction problems early: `store_proved_transaction` re-verifies proofs and signatures while finalizing, so a PCZT that would not survive broadcast fails at the prove step rather than at submission.
- **Do not auto-rebroadcast**: proved migration transfers now sit in the wallet DB as stored, unmined, deliberately not-yet-broadcast transactions, possibly for days. Any generic resubmission logic that treats "stored unmined sent transaction" as "submit now" must exclude them (e.g. skip txids present in the migration store) or it will front-run the ZIP 318 schedule and collapse the timing privacy. See the follow-up note on zcash/librustzcash#2891.

---

PR #1925, opened 2026-07-31
[#1924] Unify the release scripts into two phases

### LINKED ISSUE #1924: Unify start-release.sh and prepare-release.sh into a two-phase release script
`Scripts/start-release.sh` and `Scripts/prepare-release.sh` solve adjacent halves of a release without collaborating.

`start-release.sh` cuts `release/X.Y.Z` from the previous release tag and `review/X.Y.Z` from the revision being released, promotes the CHANGELOG, and then stops and prints a `gh` invocation for the operator to run by hand. The pull request — the artifact the whole branch topology exists to produce — is not opened by the tool that created the branches.

`prepare-release.sh` is unaware of any of that. It cuts a *third* branch, `release/ffi-X.Y.Z`, builds the XCFramework, uploads a draft release, and prints instructions to edit `Package.swift`, commit, push that third branch, and merge it into the release branch. So a release arrives as two pull requests that must be reconciled, and the branch the operator was told to work on in the first phase is not the branch the second phase commits to.

`release.sh` then calls `prepare-release.sh`, writes `Package.swift`, tags and publishes — while checking that the operator is on `main`, which the branch model in `CONTRIBUTING.md` says they must not be.

Two version numbers are also unmanaged. Nothing updates `Cargo.toml` at release time, so the crate version is whatever the last person to touch it by hand made it. `BuildSupport/platform-Info.plist` has drifted to `0.8.1` and is stamped into every framework slice of an SDK that ships `2.7.0-rc.3`.

## Proposal

Unify the first two scripts into a single two-phase `prepare-release.sh`, with a third subcommand exposing the artifact build alone so CI can call it without any git or pull-request authority.

| Command | Owns |
| -- | -- |
| `prepare-release.sh start` | Cut the branches, promote the CHANGELOG, open a **draft** PR |
| `prepare-release.sh build` | Bump versions, produce artifacts, write `Package.swift`, extend the candidate branch, flip the PR to ready |
| `prepare-release.sh artifacts` | Build, zip and upload the draft release. No git, no PR |
| `release.sh` | Post-merge: verify the checksum, sign the tag, push it, publish |

`start-release.sh` and the `release/ffi-X.Y.Z` branch both go away. The release PR's head branch is renamed `review/X.Y.Z` -> `candidate/X.Y.Z`.

Phase two runs as four steps — bump, artifacts, `Package.swift`, PR — of which only the artifact step varies between building locally and dispatching CI. Each step derives whether it has already run and skips if so, so a re-run after a failed upload resumes rather than rebuilding.

Version bumping covers `Cargo.toml`, `Cargo.lock` (via `cargo update --workspace`, so dependency versions are untouched), and `platform-Info.plist` — the latter with any pre-release suffix stripped, since Apple rejects a `CFBundleShortVersionString` that is not a dotted numeric string.

## Relationship to zcash/zcash-swift-wallet-sdk#1745

zcash/zcash-swift-wallet-sdk#1745 proposes a different decomposition of overlapping ground: CI-driven, keeping `release/ffi-<version>`, and splitting the work into `update-package-swift.sh`, `stamp-changelog.sh`, `upload-ffi-draft.sh` and `finalize-release.sh`.

This issue deliberately diverges on one point. Because the two paths must coexist — the legacy path exists to support iOS deployment targets CI runners cannot build — the artifact build is factored out as the shared unit, and *all* git and pull-request mutation stays with the maintainer. `build-ffi.yml` keeps its single `contents: write` permission and its `persist-credentials: false` checkout; it never pushes a commit or touches a PR. The CI path is then reachable through one flag, `build --artifacts ci`, and is exercised by the same four steps as the local path.

The pre-flight verification zcash/zcash-swift-wallet-sdk#1745 asks for — building the SDK and running `OfflineTests` against the freshly built FFI before publishing — is not covered here and remains worth doing on top of either decomposition.

---

*Filed with [Claude Code](<https://claude.com/claude-code>) assistance.*

nuttycom:
Closes #1924 Resolves MOB-1614

`start-release.sh` and `prepare-release.sh` solved adjacent halves of a release without collaborating. The first cut the branches but stopped short of opening the pull request the whole branch topology exists to produce. The second was unaware of any of that: it cut a *third* branch, `release/ffi-X.Y.Z`, so a release arrived as two pull requests that had to be reconciled, and the branch the operator was told to work on in phase one was not the branch phase two committed to.

## What this does

`prepare-release.sh` now owns both phases, with a human reading the pull request in between:

| Command | Owns |
| --- | --- |
| `start --issue <N> <remote> <version>` | Cut `release/X.Y.Z` from the previous release tag and `candidate/X.Y.Z` from the revision being released, promote the CHANGELOG, open a **draft** PR between them |
| `build <remote> <version>` | Bump the recorded version, build and verify the XCFramework, upload it, point `Package.swift` at it, mark the PR ready |
| `artifacts <version>` | The build alone — no git, no PR |
| `release.sh <remote> <version>` | Post-merge: verify the checksum, sign and push the tag, publish |

`start-release.sh` and the `release/ffi-*` branch are gone. The release PR's head branch is renamed `review/X.Y.Z` → `candidate/X.Y.Z`.

Pass `--dry-run` to any of them to see what would happen without changing anything.

## Notes for review

**`artifacts` is deliberately free of git and pull-request state.** That is what lets `build-ffi.yml` keep `contents: write` as its only permission and `persist-credentials: false` on checkout. It gained a `force` input so `build --artifacts ci --rebuild` has something to forward; the input reaches the script through `env:` rather than `${{ }}` interpolation into the `run:` body, so `zizmor` stays clean.

**`release.sh` now cross-checks the checksum against the asset actually attached to the release**, downloading and hashing it rather than trusting `release.env` (gitignored, absent in a fresh clone). SwiftPM validates that checksum at fetch time, so a mismatch would not fail during the release — it would fail in every consumer's build, after the tag was already public and immutable. It also no longer checks for `main`, which contradicted the branch model in `CONTRIBUTING.md`.

**Artifacts are no longer uploaded until the Swift package builds and `OfflineTests` pass against the newly built framework.** This is the pre-flight verification asked for in #1745, placed between the build and the upload — the last point at which a broken FFI boundary can be caught while the artifact is still private. A signature mismatch across that boundary compiles on both sides and surfaces only at link or run time. Because the check lives inside `produce_artifacts`, the local path, `--artifacts ci` and the CI workflow all inherit it. `--skip-verify` opts out locally.

**Version numbers are now bumped as part of the release.** `Cargo.toml`, `Cargo.lock` (via `cargo update --workspace`, leaving dependency versions untouched) and `BuildSupport/platform-Info.plist` — the last of which had drifted to `0.8.1` and was being stamped into every framework slice of an SDK shipping `2.7.0-rc.3`. The plist gets the version with any pre-release suffix stripped, since Apple rejects a non-numeric `CFBundleShortVersionString`.

**Phase two is resumable.** Each step derives whether it has already run, so a failure after the ten-minute build resumes rather than restarting.

## Testing

Text transforms and version comparisons live in `Scripts/lib/release-lib.sh`, covered by a dependency-free bash runner: `make test-scripts`, wired into `swift.yml` so it fails in seconds rather than after the FFI build. 68 assertions.

What that does **not** cover, and what a reviewer should weigh: the `gh` interactions and the resume paths have no automated coverage — they were exercised by hand via `--dry-run` and, for the partial-failure paths, by isolated scratch repositories with stubbed `gh`. A rehearsal against a fork before the first real release is still worth doing. `.plans/` in my working tree carries the full checklist, including deliberately breaking an `extern "C"` signature to confirm the new verification actually blocks an upload.

`CONTRIBUTING.md`, `docs/ci.md` and `CLAUDE.md` are updated to match.

---
*Opened with [Claude Code](https://claude.com/claude-code) assistance.*


nuttycom:
The `release.sh` script should also ensure that the release branch is properly merged back to the appropriate maintenance branch.

zcash-harry (changes requested):
A few additional points before merging:

1. PR #1802 already added automatic GitHub prerelease marking on `main`, but this rewrite does not preserve it. Versions such as `2.8.0-rc.1` should continue passing `--prerelease` when the release is created, updated, and published.

2. In `release.sh`, verify that local `release/X.Y.Z` exactly matches `<remote>/release/X.Y.Z` before tagging, then push only the tag. This prevents tagging an outdated or unreviewed local commit.

3. Restrict `--force-overwrite-existing-release` to draft releases. Replacing a published XCFramework would invalidate the checksum in `Package.swift`.

4. Skip GitHub authentication and the signing-key check during `--dry-run`, since no PR or tag is actually created.

_Changes requested by Codex Sol 5.6_

nuttycom:
Addressed in 94300650.

**1. Pre-release marking.** Confirmed, and the mechanism is worth stating: #1802 landed on `main`, this branch is cut from `maint/v2.7.x`, which never carried it — so the regression would have arrived silently at the forward merge rather than showing up in this diff.

It is now set at all three points where a release is written (`gh release create`, the `--clobber` re-upload, and `release.sh`'s publish), and in the recovery instructions the scripts print when they abort part-way, so a hand-finished release ends up with the same bit an automated one would have had.

One deliberate difference from #1802: the flag is stated in **both** directions, `--prerelease=true|false`, rather than added only for a pre-release. `gh release upload` replaces assets but not release properties, so a draft created before its version gained or shed a suffix would otherwise keep the stale bit — which is the same failure mode the `gh release edit` call in #1802's overwrite path was added to cover, just in the other direction.

**2. Tag only what merged.** `release.sh` now requires `release/X.Y.Z` to be sha-identical to `<remote>/release/X.Y.Z` before it tags anything, reports ahead and behind counts separately when it is not, and pushes `refs/tags/<version>` alone. The two properties are connected: once the branches are known equal there is nothing left for a branch push to contribute except, on a checkout that had drifted, commits the pull request never carried.

**3. Overwrite restricted to drafts.** `produce_artifacts` reads `isDraft` and refuses a published release, with the reason stated in the error — `Package.swift` pins the checksum of the asset and SwiftPM verifies it on every fetch, so the breakage is permanent and lands on consumers rather than on the operator. The check runs before the build, not after it.

**4. `--dry-run`.** Split, and I would like your read on the second half.

The signing-key check is now advisory under `--dry-run`: it warns and continues, so a rehearsal reports the problem *and* still shows the rest of the plan. That seemed better than skipping it silently, which would let a dry run print "a real run would sign and push the tag" for an operator who has no key. Same treatment for `gh` authentication in `start`, whose dry-run path reaches GitHub nowhere.

I kept the authentication check fatal in `build`, `artifacts` and `release.sh`, because their dry runs *do* make authenticated calls. `release.sh --dry-run` makes three — `release view`, `release view --json isDraft`, and `release download` — and downloading and hashing the asset to check it against `Package.swift` is the main thing that dry run exists to do. Skipping the up-front check would not make it work without a token; it would fail further along at `gh release view` and report "there is no `<version>` release on `<repo>`", blaming a missing release for a missing credential. Say the word if you would still rather have the uniform skip.

**Also fixed:** the PR body claimed `CLAUDE.md` was updated. It is a symlink to `AGENTS.md`, which was not in the diff — its Release section still documented the two scripts this branch removes.

## Verification

`make test-scripts` is at 74 assertions, up from 68.

The seven changed `gh` and git paths have no automated coverage, so I exercised them against a scratch repository with a stubbed `gh` and an ssh-signed tag: dry run with no signing key, a local-only commit, a checkout behind the remote, a real run (tag pushed, branch ref untouched), stable versus release-candidate flag values, and the draft/published split on the overwrite flag.

---
*Reply written with [Claude Code](https://claude.com/claude-code) assistance.*

zcash-harry (approved):
utACK

PR #1926, opened 2026-07-31
Add Proposal.spendsLegacyOrchardFunds

Chlup:
## Summary

Adds a public `Proposal.spendsLegacyOrchardFunds` property so wallets can detect — before creating the transactions — that a payment proposal will spend notes from the legacy Orchard pool. Post-NU6.3, spending legacy Orchard notes crosses the Orchard turnstile, publicly revealing the value that leaves the pool; wallets use this to warn the user before sending (ZODL's send-confirmation screen is the first consumer).

## Semantics

- `true` iff any proposal step has an input whose oneof value is a `receivedOutput` with `valuePool == .orchard`.
- Prior-step references never count: they point at outputs created by an earlier step of the same proposal, so any legacy-Orchard value entering the proposal necessarily appears as a `receivedOutput` in some step (this also keeps ZIP 320 TEX two-step proposals correct).
- Ironwood (ZIP 2005) inputs — FFI pool value 4, currently decoding as `UNRECOGNIZED(4)` under the vendored proto — never count as Orchard. Refreshing the vendored `proposal.proto` to name the Ironwood case is deliberately out of scope here.
- The property describes the proposal's inputs only; it is activation-agnostic. Callers that warn should apply their own network context if they support pre-activation networks.

`Proposal.testOnlyFakeProposal(totalFee:)` gains a defaulted `spendsLegacyOrchardFunds:` parameter (source-compatible) so client apps can build fixtures for both cases; its pre-existing fee behavior (fakes report a zero `totalFeeRequired()`) is intentionally unchanged and now pinned by a test in both branches.

## Testing

- New `Tests/OfflineTests/ProposalSpendsOrchardTests.swift`: orchard input, sapling/transparent-only, Ironwood raw-4 input (built via `FfiValuePool(rawValue: 4)` so the test survives a future proto regeneration), prior-step-only inputs, orchard in a later step, empty proposal, factory both variants, and the factory fee pin.
- `swift test --filter OfflineTests`: 405 tests, 0 failures.

No linked issue by agreement for this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+## Added
+- `Proposal.spendsLegacyOrchardFunds` — whether the proposal spends notes from
+  the legacy Orchard pool, so wallets can warn before a turnstile-crossing send.
+  `Proposal.testOnlyFakeProposal(totalFee:spendsLegacyOrchardFunds:)` gained a
+  defaulted parameter for building test fixtures.
+
 # 2.8.0-rc.3 - 2026-07-29
 
 ## Changed
--- a/Sources/ZcashLightClientKit/Model/Proposal.swift
+++ b/Sources/ZcashLightClientKit/Model/Proposal.swift
@@ -27,6 +27,27 @@ public struct Proposal: Equatable {
             acc + Zatoshi(Int64(step.balance.feeRequired))
         }
     }
+
+    /// Whether any transaction in this proposal spends notes received in the legacy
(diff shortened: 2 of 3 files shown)


Chlup:
Superseded by #1923.

That PR already carries these two commits (`Add Proposal.spendsLegacyOrchardFunds` and `Refine spendsLegacyOrchardFunds docs and tests`), based on `feature/ironwood-slipstream`, with byte-identical `Sources/ZcashLightClientKit/Model/Proposal.swift` and `Tests/OfflineTests/ProposalSpendsOrchardTests.swift`. Closing here rather than retargeting, to avoid two open PRs adding the same API to the same base.

PR #1927, opened 2026-07-31
Surface the v_transactions ZIP 318 classification

dannywillems:
Stacked on `dw/views-columns-up-to-date`, which surfaces the other
`v_transactions` columns the SDK was dropping. Review that one first.

## What this adds

Upstream librustzcash adds a `zip318_kind` column to `v_transactions`
(zcash/librustzcash#2874). It records how a transaction classifies against
ZIP 318, the Orchard to Ironwood pool migration, so a wallet can label a
migration transaction in its history without a migration plan, which does not
survive a seed restore.

`ZcashTransaction.Overview.ZIP318Kind` carries the five values, and
`Overview.zip318Kind` exposes it.

| code | case | meaning |
| --- | --- | --- |
| 0 | `notClassified` | the column default; nothing has looked at this transaction yet |
| 1 | `nonconforming` | classified, and it is not a ZIP 318 transaction |
| 2 | `preparation` | a note-preparation self-send a migration run makes before crossing |
| 3 | `transfer` | a pool crossing paying the account's own internal address |
| 4 | `canonicalCrossingPayment` | the same shape, paying an external third party |

## Three things a consumer must get right

**`notClassified` is not a decision.** It means "we never looked", not "this is
not a migration". Render it as no label at all. Presenting it as "not a
migration" would claim the wallet examined a transaction it never looked at.
That is why 0 and 1 are distinct cases rather than one falsy value.

**Only `preparation` and `transfer` are the wallet's own migration.**
`canonicalCrossingPayment` has the identical on-chain shape by design, because
the ordinary send path builds such payments deliberately so that they join the
migration anonymity set. It is a payment to a third party and must not be shown
as a migration.

**This names a conformance class, never a provenance.** ZIP 318's privacy
argument is that every migration transaction looks like every other one, so
recognizing the shape places a transaction in the anonymity set and can never
establish which wallet's run produced it.

## Decoding

`init(rawValue:)` is total and non-failable: an unrecognized code yields
`notClassified`, never a trap and never `nonconforming`. A future librustzcash
may add codes, and an SDK that meets one has learned nothing about the
transaction, so saying so is the honest answer.

This follows `ZcashTransaction.Output.Pool` rather than a Swift raw-value enum.
The one raw-value enum in this SDK fed from Rust, `ScanRange.Priority`, turns an
unknown value into `fatalError`, which is exactly the failure mode to avoid
here. A total initializer makes the rule structural rather than something a call
site can forget. `Pool`'s `.other(Int)` escape hatch was not copied, because
preserving the unknown integer would create a sixth state that is neither a
decision nor the documented default.

## Dependency, and an open question

The last commit pins the librustzcash crates to the revision of
zcash/librustzcash#2874, so the column exists and `Overview.init(row:)` decodes
it rather than throwing. **That pin must be reverted before this merges**, and
it must not outlive the upstream release carrying the column. It is pinned to a
specific revision rather than a branch so the build stays reproducible while the
upstream branch moves.

Without the pin, `zcash_client_sqlite 0.22.0-rc.6` has no such column and
`Overview.init(row:)` throws for every transaction against a real database, so
this PR and the upstream release must land in order.

The Android counterpart takes the other approach: it guards the column index and
treats an absent column as `notClassified`, so it keeps working before the bump.
Worth deciding whether this SDK should match. The argument for matching is the
same rule that already governs unknown codes, that an SDK which cannot read the
column has learned nothing. The argument against is that tolerating a missing
column masks a genuine schema mismatch after the bump, and that throwing is
consistent with how the other columns on the base branch behave.

## Testing

`make check` (`swift build` plus `swift test --filter OfflineTests`) passes: 326
tests, 0 failures. SwiftLint was not available in the environment used, so
`make lint` was not run. The offline tests do not exercise a real
`v_transactions` row, so they give no signal on the decode path against a live
database.


Chlup (dismissed):

Cargo.toml
```diff
@@ -92,14 +92,14 @@ cbindgen = "0.29"
...
 [patch.crates-io]
-# pczt = { git = "https://github.com/zcash/librustzcash.git", rev = "REPLACE_ME" }
-# zcash_client_backend = { git = "https://github.com/zcash/librustzcash.git", rev = "REPLACE_ME" }
-# zcash_client_sqlite = { git = "https://github.com/zcash/librustzcash.git", rev = "REPLACE_ME" }
-# zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "REPLACE_ME" }
-# zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "REPLACE_ME" }
-# zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "REPLACE_ME" }
-# zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "REPLACE_ME" }
-# zcash_transparent = { git = "https://github.com/zcash/librustzcash.git", rev = "REPLACE_ME" }
+pczt = { git = "https://github.com/zcash/librustzcash.git", rev = "8755958580c6a3d5913ac60fd91ca0174cb1ebe0" }
```
dannywillems: We will update these commits later to point to `main` of librustzcash, when https://github.com/zcash/librustzcash/pull/2874 is merged.

dannywillems (commented):
self-utACK

Chlup (dismissed):

Tests/OfflineTests/CompactBlockProcessorActions/BlockEnhancerImplTests.swift
```diff
@@ -47,7 +47,11 @@ final class BlockEnhancerImplTests: XCTestCase {
             value: value,
             isExpiredUmined: false,
             totalSpent: nil,
-            totalReceived: nil
+            totalReceived: nil,
```
dannywillems, line 50: These were missing on `maint/v2.7.x`, and was making the CI fails.

Chlup (approved):

PR #1931, opened 2026-08-03
[#1930] Clear the compiler warnings in the test targets

### LINKED ISSUE #1930: Test targets emit around 40 compiler warnings
Building the test targets on `maint/v2.7.x` (`swift build --build-tests`, which
is what the `Build and Run Offline Tests` workflow does) emits around 40
compiler warnings. None of them are in production code, but they bury the real
diagnostics: the compile error that broke #1927 was only visible after
scrolling past a screen of them.

They fall into three groups.

1. `DarksideTests` (and `AdvancedReOrgTests`, `BalanceTests`,
   `PaymentURIFulfillmentTests`), around 32 warnings. These tests have had
   their `sendToAddress` / `shieldFunds` calls commented out for a while. What
   is left behind still compiles but is dead: bindings only the commented-out
   call ever read, `var`s only it ever assigned, and `do` blocks whose sole
   remaining statement cannot throw, so the `catch` is unreachable.

2. `Tests/NetworkTests/TorClientTests.swift:20`, one warning. `try` on
   `TorClient.init(torDir:)`, which does not throw.

3. `Tests/TestUtils/proto/darkside.pb.swift`, seven warnings. The checked-in
   file was generated by an older `protoc-gen-swift` and still emits its name
   maps through `_NameMap.init(dictionaryLiteral:)`, which SwiftProtobuf
   1.35.1 deprecates with an explicit "regenerate your .pb.swift files"
   message.

Two further warnings are out of scope here because they are a real API
question rather than a cleanup: `Zatoshi+SQLite.swift:10` (retroactive
`Sendable` conformance) and `Zatoshi.swift:29` (`_amount` has non-Sendable
type `Clamped<Int64>`). Both are Swift 6 language mode errors and want their
own issue.

Likewise `Sources/ZcashLightClientKit/Tor/TorClient.swift:86`, which warns
about passing a non-Sendable `OpaquePointer` into an actor-isolated context.
The pointer from `zcashlc_tor_isolated_client` is uniquely owned and stored
nowhere else, so the hand-off is a genuine transfer that Swift 6.0 had no way
to express at that call site. `sending` (SE-0430) is the semantically correct
fix but needs a Swift 6 compiler to parse, which `swift-tools-version:5.6`
cannot assume. Also its own issue.

dannywillems:
Closes #1930.
Resolves MOB-1635.

`swift build --build-tests` on `maint/v2.7.x` emits around 40 warnings, none
of them in production code. They bury the diagnostics that matter: the compile
error that broke #1927 was only visible after a screen of them. This clears
every one of them in the test targets.

Three commits, one per group.

### Darkside tests

Those tests have had their `sendToAddress` / `shieldFunds` calls, and several
of the assertions that followed them, commented out for a while. What was left
behind still compiled but was dead: bindings only the commented-out call ever
read, `var`s only it ever assigned, and `do` blocks whose sole remaining
statement cannot throw, so the `catch` is unreachable.

The commented-out code is deleted rather than kept as a record. Git history is
the record, and a commented-out block that no compiler checks drifts out of
step with the API it calls, as these already had. That is 457 lines across
eight files.

The `TODO: [#1247]` and `FIXME:` comments stay, because they name the issue
tracking the restoration of those checks, which is the part worth keeping.
Where a binding is still read further down, it becomes an explicitly-nil `let`
rather than a never-mutated `var`, so the test still expresses that it has
nothing to assert on yet.

No behavioural change: every test touched is already disabled and none of the
removed statements were reachable.

### TorClientTests

`TorClient.init(torDir:)` does not throw, so the `try` covered no throwing
call.

### darkside.pb.swift

Regenerated, not hand-edited. The checked-in file came from an older
`protoc-gen-swift` and still emits its name maps through
`_NameMap.init(dictionaryLiteral:)`, which SwiftProtobuf 1.35.1 deprecates
with an explicit "regenerate your .pb.swift files" message. The plugin was
built from the swift-protobuf revision pinned in `Package.resolved`, mirroring
what `Scripts/update-lightwallet-protocol.sh` does for the SDK's own generated
sources; the exact commands are in the commit body.

One thing to know for next time: the new plugin drops `@testable import
ZcashLightClientKit`, which is not generated but is required, because
`DarksideSubtreeRoots` refers to the internal `SubtreeRoot` from the SDK's own
`service.pb.swift`. It is re-added by hand and now carries a comment saying
so. There is no script covering `Tests/TestUtils/proto`; adding one would be
worthwhile.

### Deliberately out of scope

- `TorClient.swift:86`, passing a non-Sendable `OpaquePointer` into an
  actor-isolated context. The pointer from `zcashlc_tor_isolated_client` is
  uniquely owned and stored nowhere else, so the hand-off is a genuine
  transfer that Swift 6.0 had no way to express at that call site. `sending`
  (SE-0430) is the semantically correct fix but needs a Swift 6 compiler to
  parse, which `swift-tools-version:5.6` cannot assume. It also does not
  reproduce under Swift 6.2, so it could not be verified here.
- `Zatoshi+SQLite.swift:10` and `Zatoshi.swift:29`. Fixing them means deciding
  whether `Clamped` should be `Sendable`, which is an API call rather than a
  cleanup.

Both are noted in #1930 and want their own issues.

### Verification

Ran locally against a macOS FFI slice built from `rust/`, the same way the
workflow does:

- `make build`: exit 0
- `make test-offline`: 331 tests, 0 failures
- `swift build --build-tests`: zero warnings under `Tests/`
- `swiftlint lint`: 0 errors, and 0 violations under `Tests/DarksideTests`
  with `.swiftlint_tests.yml`


Tests/DarksideTests/AdvancedReOrgTests.swift
```diff
@@ -323,26 +323,26 @@ class AdvancedReOrgTests: ZcashTestCase {
...
-        var testError: Error?
-        let sendAmount = Zatoshi(10000)
+        let pendingEntity: ZcashTransaction.Overview? = nil
+        let testError: Error? = nil
 
         /*
         4. create transaction
         */
-        do {
+//        let sendAmount = Zatoshi(10000)
```
dannywillems: I would be in favor of removing this code if it is documented.

Tests/DarksideTests/AdvancedReOrgTests.swift
```diff
@@ -323,26 +323,26 @@ class AdvancedReOrgTests: ZcashTestCase {
...
 //                toAddress: try Recipient(Environment.testRecipientAddress, network: self.network.networkType),
 //                memo: try Memo(string: "test transaction")
 //            )
 //            pendingEntity = pendingTx
-            sendExpectation.fulfill()
-        } catch {
-            testError = error
-            XCTFail("error sending to address. Error: \(String(describing: error))")
-        }
+//        } catch {
```
dannywillems: I would prefer to remove the code if it is commented.

Tests/DarksideTests/AdvancedReOrgTests.swift
```diff
@@ -822,25 +822,24 @@ class AdvancedReOrgTests: ZcashTestCase {
...
-        
+        let pendingEntity: ZcashTransaction.Overview? = nil
+
         /*
         2. send transaction to recipient address
         */
-        let recipient = try Recipient(Environment.testRecipientAddress, network: self.network.networkType)
-
-        do {
+//        let recipient = try Recipient(Environment.testRecipientAddress, network: self.network.networkType)
```
dannywillems: I would prefer to delete the code if commented.

Tests/DarksideTests/AdvancedReOrgTests.swift
```diff
@@ -1272,23 +1271,23 @@ class AdvancedReOrgTests: ZcashTestCase {
...
         let sendExpectation = XCTestExpectation(description: "send expectation")
-        var pendingEntity: ZcashTransaction.Overview?
-        
+        let pendingEntity: ZcashTransaction.Overview? = nil
+
         /*
         2. send transaction to recipient address
         */
-        do {
+//        do {
```
dannywillems: I would prefer to remove the code if commented.

Tests/DarksideTests/BalanceTests.swift
```diff
@@ -96,25 +96,25 @@ class BalanceTests: ZcashTestCase {
         XCTAssertTrue(verifiedBalance > network.constants.defaultFee())
         XCTAssertEqual(verifiedBalance, totalBalance)
         
-        let maxBalance = verifiedBalance - Zatoshi(10000)
-        
+//        let maxBalance = verifiedBalance - Zatoshi(10000)
```
dannywillems: I would prefer to remove the code if commented.

nullcopy (approved):
Yay for deleting code!!

utACK d7c57807534c3798e60a09eef440c23a58265d72

PR #1933, opened 2026-08-03
[#1932] Add host Rust build targets and make the aggregates compose

### LINKED ISSUE #1932: Makefile: add host Rust build/clean targets and make the aggregate targets compose
## Summary

The `Makefile` has no way to build the Rust crate on its own, and the aggregate
targets do not compose: `clean-all` and the build targets are not inverses, so
a full clean leaves state behind that later builds silently link against.

## Problems

1. **No host Rust build.** The only way to compile the Rust is `make ffi-macos`
   / `make ffi-all`, which always build release, always cross-compile to Apple
   triples, and then assemble an XCFramework. There is no fast debug build for
   the edit-typecheck-link loop. The Android SDK has exactly this split
   (`build-rust` for the host, `build-rust-android` for the ABIs).

2. **No way to clean the Cargo artifacts.** `clean-ffi` deliberately preserves
   `target/`, and nothing else removes it, so `distclean` in
   `BuildSupport/Makefile` is the only route and it is not reachable from the
   root `Makefile`.

3. **Nothing builds everything.** Building from source takes three commands
   (`make ffi-macos && make configure-local-ffi && make check`), documented in
   `AGENTS.md` but not expressed in the target graph.

4. **`clean-all` is not the inverse of any build target.** It removes
   `.build/` and `BuildSupport/products/` but leaves both `target/` and
   `LocalPackages/`. Since `Package.swift` switches to local-FFI mode purely on
   the presence of `LocalPackages/Package.swift`, a stale copied XCFramework
   survives a full clean and keeps being linked.

5. **`configure-local-ffi` fails obscurely.** It copies `$(XCFRAMEWORK)` with
   no existence check, so a missing XCFramework surfaces as a raw
   `cp: No such file or directory` rather than the clear message `verify-ffi`
   already prints.

## Proposed change

Root `Makefile` only. No Swift, Rust, or CI changes.

New targets:

- `build-rust` / `build-rust-release`: build the crate for the host, debug and
  release. Mirrors the Android SDK. `Cargo.toml` is at the repo root here, so
  cargo runs from there with no directory to change into.
- `clean-rust`: `cargo clean`.
- `build-all`: `ffi-macos` then `configure-local-ffi` then `build`.
- `build-all-release`: same, but `build-release`.
- `check-all`: `build-all` then `test`.

Changed targets:

- `clean-all` gains `clean-rust` and `reset-ffi`, so it is now the inverse of
  `build-all`.
- `configure-local-ffi` depends on `verify-ffi`, for the clear error.

## Note

`-all` ends up meaning three different things across the file: "including
Rust" (`build-all`, `check-all`, matching the Android SDK convention), "all
architectures" (`ffi-all`), and "including networked tests" (`test-all`).
Worth reconciling separately; renaming `test-all` is out of scope here.

dannywillems:
Closes #1932.
Resolves MOB-1636

`Makefile` only. No Swift, Rust, or CI changes.

## What and why

The only way to compile the Rust was `ffi-macos` / `ffi-all`, which always
build release, always cross-compile to Apple triples, and then assemble an
XCFramework. There was no host build for the edit-typecheck-link loop, and no
way to clean the Cargo artifacts, since `clean-ffi` preserves `target/` on
purpose.

The aggregate targets also did not compose. Building from source took three
separate commands, and `clean-all` was the inverse of none of them: it left
both `target/` and `LocalPackages/` behind. Because `Package.swift` selects the
local FFI purely on the presence of `LocalPackages/Package.swift`, a stale
copied XCFramework survives a full clean and keeps being linked. That is not
hypothetical, see the test plan below.

## Targets

New:

| Target | Does |
|---|---|
| `build-rust` | `cargo build` (host, debug) |
| `build-rust-release` | `cargo build --release` (host) |
| `clean-rust` | `cargo clean` |
| `build-all` | `ffi-macos` then `configure-local-ffi` then `build` |
| `build-all-release` | same, but `build-release` |
| `check-all` | `build-all` then `test` |

`build-rust` / `build-rust-release` mirror the Android SDK's targets of the
same name. `Cargo.toml` is at the repo root here rather than in a crate
subdirectory, so cargo runs from the root with no directory to change into.

Changed:

- `clean-all` gains `clean-rust` and `reset-ffi`, making it the inverse of
  `build-all`.
- `configure-local-ffi` now depends on `verify-ffi`, so a missing XCFramework
  reports `Error: BuildSupport/products/libzcashlc.xcframework not found`
  instead of a bare `cp: No such file or directory`.

## Test plan

Ran on macOS with Xcode 16, Swift 6.2.3, cargo 1.97.1.

- `make build-rust` builds the crate for the host: 41s against a warm cache,
  producing `target/debug/libzcashlc.a`.
- `make build-rust-release` produces `target/release/libzcashlc.a` (56MB) in
  2m16s. The `[profile.release] lto = true` link pass dominates.
- `make clean-rust`, `make clean-all`, `make build-all`, `make
  build-all-release` and `make check-all` each verified with `make -n` to
  expand to the intended chain. `clean-all` expands to `swift package clean`,
  `make -C BuildSupport clean`, `cargo clean`, `./Scripts/reset-local-ffi.sh`
  in that order.
- `make help` lists every new target.
- `configure-local-ffi` guard confirmed by running it with
  `BuildSupport/products/` absent: it now fails on `verify-ffi` with the clear
  message rather than a `cp` error.
- The stale-FFI problem reproduced directly. With a `LocalPackages/` copy from
  an older commit still in place, `make build` fails with
  `ZcashRustBackend.swift:1352:15: error: cannot convert value of type 'Void'
  to expected condition type 'Bool'`, because the stale `zcashlc.h` declares
  `zcashlc_set_transaction_status` as returning `void`. `make build-all`
  rebuilds the FFI from source and resolves it; `make clean-all` now clears
  the stale copy rather than leaving it to be linked.

No manual testnet verification applies: this changes no runtime behaviour.

## CI impact

The one change CI touches is `configure-local-ffi` gaining a `verify-ffi`
dependency. `swift.yml` already runs `make verify-ffi` immediately before
`make configure-local-ffi`, and the XCFramework exists at that point, so the
added dependency is a no-op there.

## Note, not addressed here

`-all` now means three different things across the file: "including Rust"
(`build-all`, `check-all`, matching the Android SDK convention), "all
architectures" (`ffi-all`), and "including networked tests" (`test-all`). So
`check-all` runs the offline suite while `test-all` runs the networked one.
Worth reconciling, but renaming `test-all` is out of scope.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,12 +95,25 @@ test: test-offline ## Run the offline test suite
 .PHONY: check
 check: build test ## Build and run the offline tests
 
+# `build` links whatever FFI is already in place; these build the Rust from
+# source first, so the XCFramework matches the tree.
+.PHONY: build-all
+build-all: ffi-macos configure-local-ffi build ## Build the FFI from source, then the Swift package
+
+.PHONY: build-all-release
+build-all-release: ffi-macos configure-local-ffi build-release ## Build the FFI and the Swift package in release mode
+
+.PHONY: check-all
+check-all: build-all test ## Build everything from source and run the offline tests
+
 .PHONY: clean
 clean: ## Clean the SwiftPM build directory
 	$(SWIFT) package clean
 
+# reset-ffi last: it removes LocalPackages/, which would otherwise survive as a
+# stale copy that Package.swift keeps linking after everything else is gone.
 .PHONY: clean-all
-clean-all: clean clean-ffi ## Clean SwiftPM and FFI build artifacts
+clean-all: clean clean-ffi clean-rust reset-ffi ## Clean all artifacts and reset the FFI mode
 
 .PHONY: setup
 setup: ## Set up the development environment
@@ -143,6 +156,27 @@ format: ## Apply SwiftLint autocorrections
 swift-build: build ## Alias for `make build`
 
 # ---------------------------------------------------------------------------
+# Rust crate
+# ---------------------------------------------------------------------------
+#
+# Cargo.toml is at the repo root, so cargo runs from here. These build for the
+# host only; the ffi-* targets below cross-compile for Apple platforms.
+
+.PHONY: build-rust
+build-rust: ## Build the Rust crate for the host (debug)
+	$(CARGO) build
+
+.PHONY: build-rust-release
+build-rust-release: ## Build the Rust crate for the host (release)
+	$(CARGO) build --release
+
+# Removes target/ entirely, including the cross-compiled slices the ffi-*
+# targets build. clean-ffi preserves those on purpose.
+.PHONY: clean-rust
+clean-rust: ## Clean the Cargo build artifacts (target/)
+	$(CARGO) clean
+
+# ---------------------------------------------------------------------------
 # Rust FFI (BuildSupport/)
 # ---------------------------------------------------------------------------
 #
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
utACK, nice regularization.

PR #1938, opened 2026-08-04
Update to the published zcash_client_backend 0.24.0-rc.7 crates

nuttycom:
Replaces the librustzcash git pin with the published zcash_client_backend 0.24.0-rc.7 and zcash_client_sqlite 0.22.0-rc.7, which carry the zip318_kind column the SDK reads, and with it zcash_protocol 0.10.4, pczt 0.9.2 and zcash_pool_migration 0.1.0-rc.6. The patch table is kept, commented out, so that a future pin needs only a revision.

Records the one fix these releases bring that this SDK can observe: extract_and_store_transaction_from_pczt now stores a transaction's Ironwood outputs, so a post-NU6.3 PCZT payment delivered through Ironwood keeps its recipient address and memo, which were previously dropped and are not recoverable afterwards. The rust CHANGELOG's migrated-to line was still naming the 0.24.0-rc.5 crates; it now names what the manifest pins.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,8 @@ and this library adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   SDK that does not know the code has learned nothing about the transaction.
 
 ## Changed
-- Updated the librustzcash crates to `zcash_client_backend 0.24.0-rc.6` and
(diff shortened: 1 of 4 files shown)


dannywillems (approved):
utACK fb0c63462564944334b8105899f5c68ee44d283c

PR #1945, opened 2026-08-04
[MOB-1466] Interim librustzcash pin + NULL trust_status decode fix

Chlup:
Three commits, two concerns:

## Interim librustzcash pin (8f962d47, da94c695)

Pins the librustzcash family via the `[patch.crates-io]` template to **main at the zcash/librustzcash#2909 merge** (`b49e2cb3`). What it delivers over the published rc.7/rc.6 line:

- **zcash/librustzcash#2909** — migration transactions are ZIP 318-classified at store-at-prove time, not only after they mine. This is what activates the app's unmined-migration Activity filter for the whole prove→broadcast window (M3).
- **zcash/librustzcash#2907** — the engine's advance-path selection picks prove/broadcast steps by scheduled height instead of internal id (M2's engine half). Note: the SDK's own next-due delivery query still selects by id — tracked separately as the M2 residual.

Interim by design: the Cargo.toml comment and CHANGELOG both say to recomment the block at the first published rc containing both PRs. Pinned to the merge-commit hash (permanently reachable), never a branch.

## NULL `trust_status` decode fix (bc63f42f)

`trust_status` is an opt-in marker (`set_tx_trust`) with no default, no backfill, and no production caller anywhere yet, so it is NULL on every row of every real wallet. The `Overview` decode read it as a strict non-optional `Bool`: the first row threw, the entire `getAllTransactions` failed, and the app rendered an **empty transaction list over a fully-populated wallet** (field, 2026-08-04) — silently, because the only live tests on this path were COUNT queries (the `_testFind…` family has been disabled since #1518).

NULL now decodes as untrusted — the same interpretation librustzcash's own readers apply with `IFNULL(trust_status, 0)`. The new regression test decodes every row of the real migrated fixture through `v_transactions` (the suite's only live `Overview` decode against the real schema) and reproduced the exact field error red before the fix. The upstream view-side fix is zcash/librustzcash#2919; this decode tolerance protects builds regardless of pin generation.

## Verification

- SDK offline suite: **796/796** (795 + the new regression test), on both pin revs.
- All three FFI slices rebuilt against the pin; full zodl suite on top of this SDK: **1037 tests / 157 suites green**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,15 @@ Changes are relative to `2.8.0-rc.3`.
 
 ## Changed
 
-- The librustzcash family rides an interim git pin (librustzcash PR #2909) so a wallet's own
-  scheduled migration transactions carry their ZIP 318 classification (`Overview.zip318Kind`)
-  from the moment they are STORED at proving time, instead of only after they mine and are
-  scanned. This is what lets a wallet hold stored-but-unmined migration transactions out of its
-  activity list. The pin reverts to published crates at the first rc containing that PR.
+- The librustzcash family rides an interim git pin (librustzcash main at the #2909 merge) so a
+  wallet's own scheduled migration transactions carry their ZIP 318 classification
+  (`Overview.zip318Kind`) from the moment they are STORED at proving time, instead of only after
+  they mine and are scanned. This is what lets a wallet hold stored-but-unmined migration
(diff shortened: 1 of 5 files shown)


LukasKorba (commented):
utACK

LukasKorba (approved):

PR #1946, opened 2026-08-04
Move the librustzcash interim pin to main at the #2910 merge

Chlup:
## Summary

Moves the interim librustzcash pin from main@`b49e2cb3` (the #2909 merge) to main@`e8f5b1f4` (the #2910 merge) — the two PRs that landed in between:

- **zcash/librustzcash#2910** — a wallet that sleeps through part of its ZIP 318 broadcast schedule no longer finds every missed transfer due at once on resume (which would broadcast the backlog as a cluster, defeating the drawn inter-arrival delays). The schedule re-spreads with the gaps intact; at most the one overdue transfer is released immediately.
- **zcash/librustzcash#2898** — scheduling test bounds derived from ZIP 318 constants (test-side only).

#2910 threads an rng through `advance_migration` (the re-spread draws fresh inter-broadcast gaps), so the FFI wrapper in `rust/src/migration.rs` now passes `&mut OsRng` — the file's established idiom. That is the only code change; everything else is the pin swap (11 `[patch.crates-io]` lines + 13 `Cargo.lock` source entries, no registry churn) and the CHANGELOG.

**Slipstream needs no rewiring**: `slipstream-core` (pin unchanged at `ab89ada4`) compiles cleanly against the new rev through the `[patch.crates-io]` override, which redirects the librustzcash crates for the whole cargo graph.

Note: zcash/librustzcash#2919 (the `v_transactions` `trust_status` IFNULL fix) is still open upstream and therefore **not** in this rev — the tolerant decode from #1945 keeps covering it.

## Verification

- All three local FFI slices rebuilt against the new rev (ios-device, ios-sim universal, macOS).
- `swift test --filter OfflineTests`: **796 tests, 0 failures**, exit 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,15 +10,17 @@ Changes are relative to `2.8.0-rc.3`.
 
 ## Changed
 
-- The librustzcash family rides an interim git pin (librustzcash main at the #2909 merge) so a
+- The librustzcash family rides an interim git pin (librustzcash main at the #2910 merge) so a
   wallet's own scheduled migration transactions carry their ZIP 318 classification
   (`Overview.zip318Kind`) from the moment they are STORED at proving time, instead of only after
-  they mine and are scanned. This is what lets a wallet hold stored-but-unmined migration
-  transactions out of its activity list. The rev also carries librustzcash #2907: the engine's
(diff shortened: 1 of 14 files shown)


LukasKorba (approved):
utACK

PR #1947, opened 2026-08-05
TEMPORARY: pin librustzcash to 59ed8930 and adapt to its API

nuttycom:
Adopts `librustzcash` revision `59ed8930b8c6d96ce7c0c44a756d480f7311be8a` and fixes the resulting
compilation errors.

Part of #1806 — deliberately closes nothing, since the pin is meant to be replaced by released
version numbers rather than merged and left standing.

Depends on https://github.com/zcash/librustzcash/pull/2929 and the associated PR stack. Counterpart
to zcash/zcash-android-wallet-sdk#2134, which adapts the same revision on the Android side.

## Why a pin

The `zcash_pool_migration` changes this adapts to are in no published crate, and a version
requirement cannot reach them: the pinned crate calls itself `zcash_pool_migration` 0.1.0-rc.6, and
so does the latest published rc.6 — but the published one still has the three-parameter
`WalletProveError`, no `MigrationProver::lock_spent_notes`, and the two-argument
`set_transaction_proved`. Only a revision distinguishes the two. This pins a specific revision
rather than a branch, so the build stays reproducible while the upstream branch moves.

**The pin must not outlive the upstream release that carries these changes.** Replace it with the
released version numbers before merging; the source adaptation stays.

## What changed upstream

**`wallet::WalletProveError` gains a fourth type parameter** (the lock store's error) and a
`Lock(LockError<LE>)` variant, reporting a note another flow has already reserved.

`ProveErrorClass::is_transient` classifies a lock **conflict** as transient. It resolves without the
prove sweep's help either way: the competing flow releases or expires its lock, after which a later
attempt proves; or its transaction mines, after which the next attempt reports `UnknownSpentNote`,
already transient here. Deferring is also the only response available, since the notes were fixed by
the transaction's signature and nothing can re-select around the conflict. A lock-store **write
failure** is not a conflict, does not self-resolve, and still surfaces as hard — the same reading
this impl already applies to `Tree(ShardTreeError::Storage(_))`.

**`satisfiability::advance_migration` takes a `CryptoRng`**, for the anchor-boundary redraws it now
performs when re-spreading a broadcast schedule the wallet slept through. `zcashlc_migration_advance_step`
passes `OsRng`, the source every other engine entry point in this crate is given: a still-unproved
transfer whose schedule shifts forward but whose boundary did not would broadcast an anchor older
than any honest draw by exactly the shift — and every deferred transfer of the wallet older by the
*same* amount, a linkable fingerprint. These draws are privacy-bearing, not incidental.

**`MigrationState::set_transaction_proved` takes the lock owner** the proof's spent notes were
reserved under, and `MigrationTransaction::from_parts` now types that field as
`MigrationLockOwner` rather than `[u8; 32]`. The one production call site is `due_assuming_proving`'s
scratch simulation, which re-states the row's own stored token so the virtual proof stays a pure
lifecycle flip.

**`PoolMigrationRead::get_migration` is now PENDING-ONLY**: a migration persisted with a terminal
status is retained as history and no longer returned there. This one is silent — it compiles
unchanged and changes behavior — and it broke two tests.

Reviewing all fourteen `get_migration()` call sites, the `None` answer and the `is_terminal()` answer
already coincided at every one except two, which now read through the new history accessor
(`PoolMigrations::latest_migration`, exposed on this SDK's `Backend`):

- `zcashlc_migration_advance_step` reports the `Complete` step verbatim for a terminal run. Read
  through the pending-only accessor it would answer NULL — "no run was ever stored" — for the one
  case the function exists to name.
- `reconcile_mined`, which backs the read-only entry points. `zcashlc_migration_transaction_statuses`
  lists the transactions a finished run mined, and nothing else would ever surface them again.

Every caller keeps its own `is_terminal()` branch; the entry points that *act* on a run (committing,
signing, proving, broadcasting, replanning) still read `get_migration`, so the store's own
pending-only filter is what keeps them off a record nothing will drive.

## Verification

- `cargo check --all-targets` — clean; the two remaining warnings are pre-existing dead-code ones.
- `cargo test` — 138 passed, 1 failed. The failure is pre-existing and unrelated; see below.
- `cargo fmt` — applied.
- `./Scripts/update-proposal-proto.sh --check` — `proposal.proto` matches the new revision.
- `make ffi-macos` then `make check` — the documented gate for a Rust-internals change. Exit 0;
  the offline suite executed 796 tests with 0 failures. No FFI signature changed, so `zcashlc.h`
  regenerates identically and the Swift layer is untouched.

## Noted, not fixed

**The wallet-side record seam moved from prove to broadcast, and this SDK has not followed it.**
Upstream's `store_proved_transaction` no longer writes the finalized transaction into the wallet's
own tables; the new `PoolMigrations::take_transaction_for_broadcast` does, atomically with handing
out the broadcastable bytes. This SDK's `serve_proved` extracts those bytes from the stored proven
PCZT in memory and never calls the new seam, so as things stand a migration transaction is never
recorded in the wallet's tables at all — no spend marks, no status-retrieval queue entry, no
`v_transactions` row ever. Adopting the seam is a change to the delivery lane's shape rather than a
compile fix, so it is left out of this PR.

**The unmined-migration Activity filter has nothing to show until broadcast.** The same upstream
change means a pending migration transaction no longer appears in `v_transactions` before broadcast
— which is the feed `TransactionDao` reads and the feature the previous interim pin
(`8f962d47`) was taken for. Upstream's answer is the new `v_transactions_with_pending_migrations`
view, a union that leaves `v_transactions` itself unchanged, so a consumer opts in by switching view
names. That is a deliberate choice about this SDK's read surface (and about the AGENTS.md
views-only rule, which currently sanctions exactly two views), not a mechanical swap.

**The hand-rolled cancel is now upstream's job.** `zcashlc_migration_schedule` cancels a pending run
by writing `MigrationStatus::Failed` directly. Upstream's `cancel_migration` additionally releases
the note reservations the run's never-broadcast transactions hold and works without deserializing
the state; its companion schema migration exists specifically to clean up the rows this SDK's
hand-rolled version left in the field. Worth adopting, separately.

**One pre-existing test failure, unrelated to this bump.**
`migration_keystone::tests::build_sign_batch_qr_parts_retains_the_dummy_orchard_spend_auth_sig`
asserts that a dummy Orchard `spend_auth_sig` survives `redact_pczt_for_batch_signer`. Upstream
clears it, deliberately ("deployed batch Signers reject requests that already carry signatures"), and
`build_sign_batch_qr_parts`'s own doc comment in this repo says the same. The function is
byte-identical between the previous pin and this one; checked out at the previous pin the test fails
identically, so this bump neither caused nor could fix it. There is a worktree on
`fix/pczt-v6-wallet-spend-derivations` that looks like where it belongs.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1680,7 +1680,7 @@ dependencies = [
 [[package]]
 name = "equihash"
 version = "0.3.0"
-source = "git+https://github.com/zcash/librustzcash?rev=b49e2cb3f96e77cd97a0b350cbf0eadfaa0ef05c#b49e2cb3f96e77cd97a0b350cbf0eadfaa0ef05c"
+source = "git+https://github.com/zcash/librustzcash?rev=59ed8930b8c6d96ce7c0c44a756d480f7311be8a#59ed8930b8c6d96ce7c0c44a756d480f7311be8a"
 dependencies = [
  "blake2b_simd",
  "corez",
@@ -1716,7 +1716,7 @@ dependencies = [
 [[package]]
 name = "f4jumble"
 version = "0.1.1"
-source = "git+https://github.com/zcash/librustzcash?rev=b49e2cb3f96e77cd97a0b350cbf0eadfaa0ef05c#b49e2cb3f96e77cd97a0b350cbf0eadfaa0ef05c"
+source = "git+https://github.com/zcash/librustzcash?rev=59ed8930b8c6d96ce7c0c44a756d480f7311be8a#59ed8930b8c6d96ce7c0c44a756d480f7311be8a"
 dependencies = [
  "blake2b_simd",
 ]
@@ -3610,7 +3610,7 @@ checksum = "2ee67f1008b1ba2321834326597b8e186293b049a023cdef258527550b9935b4"
 [[package]]
 name = "pczt"
 version = "0.9.2"
-source = "git+https://github.com/zcash/librustzcash?rev=b49e2cb3f96e77cd97a0b350cbf0eadfaa0ef05c#b49e2cb3f96e77cd97a0b350cbf0eadfaa0ef05c"
+source = "git+https://github.com/zcash/librustzcash?rev=59ed8930b8c6d96ce7c0c44a756d480f7311be8a#59ed8930b8c6d96ce7c0c44a756d480f7311be8a"
 dependencies = [
  "blake2b_simd",
  "bls12_381",
@@ -7797,7 +7797,7 @@ dependencies = [
 [[package]]
 name = "zcash_address"
 version = "0.13.0"
-source = "git+https://github.com/zcash/librustzcash?rev=b49e2cb3f96e77cd97a0b350cbf0eadfaa0ef05c#b49e2cb3f96e77cd97a0b350cbf0eadfaa0ef05c"
+source = "git+https://github.com/zcash/librustzcash?rev=59ed8930b8c6d96ce7c0c44a756d480f7311be8a#59ed8930b8c6d96ce7c0c44a756d480f7311be8a"
 dependencies = [
  "bech32",
  "bs58",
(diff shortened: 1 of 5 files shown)


nuttycom:
Subsumed by #1806

PR #1948, opened 2026-08-05
Delegate migration selection and preview to the engine's exported reads; retire the debug reschedule endpoint

Chlup:
Stacked on `michal/bump-librustzcash-pin-2910` — four commits replacing the FFI's hand-rolled migration logic with the pinned engine's own answers, per the delegation audit of this SDK against librustzcash.

## Why

The migration FFI predates the modern engine and still carried private re-implementations of decisions the engine now provides. Two of them were live correctness hazards, not just duplication: the delivery/prove read lanes selected the FIRST ready row in commit order while the engine's own queues (since the height-ordered-selection change upstream) pick the schedule-earliest — so `migrationNextDueTransfer`/`migrationHasReadyBroadcast` could name a different transaction than the advance step would offer in the same state, and the prove sweep (one proof per FFI call) proved in commit order instead of oldest-anchor-first, defeating the engine's checkpoint-retention rationale. The plan-preview numbering hand-mirrored the engine's commit-time id assignment ("mirrors EXACTLY", per its own doc) with silent-desync risk.

## The four commits

1. **Pin move** — the librustzcash interim pin advances to the head of `michal/pool-migration-public-next-due` (zcash/librustzcash#2938, stacked on the spacing-floor PR): the spacing-floor lineage plus two public read-only selection methods on `MigrationState` (`next_due_broadcast`, `next_provable`), exported precisely so consumers can stop reimplementing selection. (A follow-up commit on this branch re-pins to that branch's amended head after its own review round.)
2. **Selection delegation** — the FFI's `next_ready_action`/`next_provable_excluding` are deleted; the delivery lane, the overdue/ready queries, the virtual-prove composition, and the prove sweep all select through the engine's exported reads. Two new pins hold the fix: a drive-vs-read agreement fixture where id-order and height-order disagree, and an oldest-anchor-first prove-order fixture. The note-split ceremony picks engine-first (due-`Proved` before `Signed`) with its documented ahead-of-schedule fallback preserved byte-for-byte.
3. **Preview delegation** — `preparation_steps_from_plan` and the transfer-id offset now read `MigrationPlan::planned_transactions()` / `preparation_tx_count()`; the whole-previous-layer dependency edges remain local (the commit path's own rule, not expressed by the planned enumeration). A safety-net equality test plus the existing propose-vs-re-serve agreement test pin the swap; output is byte-identical.
4. **Debug endpoint retirement** — `zcashlc_migration_debug_reschedule_transfers` and its whole Swift chain (welding, backend, `OrchardMigration`, both synchronizers, error case ZRUST0139 retired per the ZRUST0098 tombstone precedent, mocks and tests) are deleted. It was the only code writing raw SQL against engine-owned tables, and its testing purpose is superseded engine-side (compressed test-network scheduling + the spacing floors). It never shipped in an rc; MIGRATING.md carries the note for branch consumers.

## Behavior notes (deliberate)

- The read lanes may now legitimately name an earlier-scheduled row than before — that is the fix, and it makes the read lanes consistent with the advance lane by construction.
- Mainnet/testnet engine behavior is untouched (the engine commit is additive public API only).
- App-side debug callers of `migrationDebugRescheduleTransfers` must drop the call when adopting this branch.

## Gates

- Rust: `cargo test` — 139 passed / 1 failed. The one failure is PRE-EXISTING and inherited from the base branch (`migration_keystone.rs:494`, Keystone QR redaction — proven byte-identical on the untouched baseline; tracked separately, unrelated to this change).
- Swift: `swift build` clean; `swift test --filter OfflineTests` — 799 passed / 0 failed, consistent across repeated runs (DarksideTests are non-functional in this environment and excluded by instruction).
- FFI slices rebuilt against the final pin (ios-device, ios-sim universal, macos universal): pinned-rev source-path hits 150/300/299 across the three binaries, zero stale-rev hits, and the retired debug symbol absent from all binaries and headers (`zcashlc_migration_advance_step` intact at 7 parameters).

## Care and feeding

The engine chain (#2927 ← #2932 ← the selection-export branch) is force-push-managed while under review: any amend upstream re-pins this branch and rebuilds the slices, same discipline as the prior pin moves.


LukasKorba (approved):
utACK

PR #1950, opened 2026-08-05
Serve the migration UI reads from read-only connections (no reconcile on the read paths)

Chlup:
Stacked on #1951 (the D1+D2 + pin-move branch; carries the librustzcash pin at the rebased michal/pool-migration-public-next-due head, 41a1e17c).

The six migration read entry points — statuses, progress, has_overdue, has_invalid,
has_ready_broadcast, sync_wakeups — were "read-shaped but WRITE": each ran `reconcile_mined`
first, whose conditional whole-run rewrite pinned them to the Swift write actor, so every UI
and gate read queued behind in-flight proof chunks. They now open through a new `open_read()`
(both connections SQLITE_OPEN_READ_ONLY, preamble writers skipped) and report the persisted
run as-is.

Correctness: Broadcast→Mined promotion is still persisted on every write-lane call —
`advance_migration` sweeps mined-ness itself (see the advance_step wrapper's comment), and the
prove/serve lanes still reconcile — so a read trails a just-mined broadcast by at most one
drive pass. Read-only-ness is machine-enforced (a write down a read path fails
SQLITE_READONLY, pinned by test); fresh-DB error-path answers are pinned before and after the
switch; the progress path tolerates the lazily-created `sdk_immediate_runs` table being absent.

Swift side: the six methods move to the DB-READ side of the actor split (no signature
changes); DBActor's header documents the machine-enforced exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,8 @@ Changes are relative to `2.8.0-rc.3`.
   `migrationSyncWakeups`) no longer take the database write actor: the FFI now serves them from
   read-only connections without reconciling first, so they answer in milliseconds even while a
   proof is being generated. A just-mined broadcast can trail in these answers by at most one
-  drive pass (≤30 s foregrounded); checkmark/"done" rendering is unaffected (it derives from the
+  write-lane pass (the platform's next advance-step, prove, or delivery call — typically its next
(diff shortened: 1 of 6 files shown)


LukasKorba (approved):
utACK

PR #1953, opened 2026-08-06
Move the librustzcash pin to the exact-funding head (batch Prove + outlook + adapter snapshot)

Chlup:
References #1806.

Moves the librustzcash `[patch.crates-io]` pin from `41a1e17c` to `4d654982c4` — the `michal/tmp-next-due-plus-exact-funding-note-count` head: the rebased public next-due selection plus the exact-funding-note-count merge (#2949), over a main that now carries #2939 (batch Prove), #2946 (adapter spendable-note snapshot), #2943 + #2945 + #2947 (preparation/denomination rework), and #2916/#2915/#2911 (ZIP 318 classification at store time, idempotent sent-transaction stores) — and adapts the SDK to the two API generations this brings, plus one SDK-side parity fix the pin alone does not deliver.

## What changed

**1. `63ebafeb` — pin move + the batch Prove generation (librustzcash #2939).** Upstream's `AdvanceStep::Prove` now serves the WHOLE provable set (`Vec<ProveTarget>`) instead of one transaction: proving emits nothing on-chain, so a synced session proves everything provable at once, while broadcast stays one-at-a-time on the privacy schedule. The FFI struct grows a `FfiProveTarget` array (the per-step `kind_*` fields moved onto the entries), and the Swift case becomes `MigrationAdvanceStep.prove(transactions: [MigrationProveTarget])` — earliest-ready first, never empty. The discharge path is unchanged: `finalizeReadyMigrationTransfers(accountUUID:)` already proves every ready row in one pass.

**2. `e9c819c8` — the advance outlook surfaced (librustzcash #2936).** `advance_migration` has returned `Advance { step, next }` since the previous pin; the conduit used to discard `next`. It now marshals both: `migrationAdvanceStep(accountUUID:)` returns `MigrationAdvance?` — the step plus an advisory outlook (`MigrationNextWork`: the earliest target height at which the migration next has serviceable work, and its `MigrationStepKind`), letting a host register its next wake-up from the engine's own plan (a `.broadcast` outlook needs no sync session; a `.prove` one is sync-bound). `migrationSyncWakeups` remains the proving-schedule authority. Rider: the conduit's `ReorgSettleDepth::new(10)` now names upstream's exported `PROVABLE_ANCHOR_DEPTH` (same value).

**3. `a08bdc49` — spendable-note snapshot for the SDK's own adapter (parity with librustzcash #2946).** Upstream's fix landed in THEIR `WalletMigration` adapter, which this SDK cannot use (it requires an unconditional spending key; our `Backend` exists to serve hardware-wallet accounts holding only a viewing key). Our adapter had the same quadratic shape — `resolve_wallet_note` re-ran the full note selection per input during commit-time signing — which on a large wallet stalled plan-accept for ~25 minutes until the app was killed, and could shift the selection index space mid-commit when locks landed. The adapter now snapshots the selection per instance (`RefCell`, upstream's exact pattern); a regression test pins snapshot stability, fresh-adapter freshness, and index-space agreement.

**4. `61e275a8`** — aligns the remaining doc prose (MIGRATING.md, `Synchronizer` protocol docs, changelog Added bullet, one FFI field doc) with the batch-and-outlook generation.

Consumer-facing detail is in `CHANGELOG.md` (`## Changed`), `rust/CHANGELOG.md`, and `MIGRATING.md` (before/after snippets for both breaking shapes). The Sourcery mock regeneration in `e9c819c8` also swept two stubs of a protocol member the base branch had already removed — expected, not scope creep. Known nit left in place: one singular "prove step's kind" phrase in MIGRATING.md's intro prose (the batch framing is fully spelled out in the same file's discharge list).

## Test plan

- `cargo test`: 147 passed / 1 pre-existing documented failure (`migration_keystone.rs`, see f83a5226) — the only tolerated failure.
- `cargo fmt --check` clean; `cargo clippy` 13 warnings (lib) / 47 (all-targets), exactly the branch-point baseline.
- `make check` (macOS local FFI): 798/798 offline Swift tests, including new coverage: batch-Prove marshal (multi-entry order + kinds, empty batch → nil), outlook marshal (height+kind, `-1` sentinel, unknown-kind nil), and the adapter-snapshot regression test.
- `make ffi-all`: all five architectures build; XCFramework assembles.
- `./Scripts/update-proposal-proto.sh --check`: no drift from the moved `zcash_client_backend` pin.
- Manual (testnet): accept a migration plan on a large wallet — commit completes without the multi-minute stall; drive a run and observe the advance log line now carrying the batch and the outlook.

### DIFF
--- a/rust/src/migration_engine.rs
+++ b/rust/src/migration_engine.rs
@@ -19,6 +19,7 @@
 //! [`crate::migration_finalize`].
 
 use anyhow::anyhow;
+use core::cell::{Ref, RefCell};
 use incrementalmerkletree::Position;
 use orchard::keys::{FullViewingKey, SpendAuthorizingKey};
 use orchard::note::Note as OrchardNote;
@@ -60,6 +61,13 @@ pub(crate) struct Backend<'a> {
     account: AccountUuid,
     usk: Option<UnifiedSpendingKey>,
     store: PoolMigrations<&'a mut rusqlite::Connection, NetworkParams, SystemClock>,
+    /// The spendable-note snapshot every read is served from, filled on first use — the same
+    /// contract upstream's `WalletMigration` adapter adopted in #2946: the engine addresses a
+    /// note by its index into this sequence, so every read through one adapter must see the same
+    /// set, and resolving each spent note through a fresh selection made signing a large plan
+    /// quadratic in the wallet's note count. Wallet changes are observed by constructing a fresh
+    /// adapter (every FFI entry point does).
+    spendable: RefCell<Option<Vec<SpendableNote>>>,
 }
 
 impl<'a> Backend<'a> {
@@ -86,6 +94,7 @@ impl<'a> Backend<'a> {
                 account,
             )
             .map_err(|e| anyhow!("opening the account-scoped migration store failed: {e}"))?,
+            spendable: RefCell::new(None),
         })
     }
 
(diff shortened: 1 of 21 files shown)


rust/src/migration_engine.rs
nuttycom, line 1: The changes in this file look heavily duplicated from the backend, I need to understand better why the LLM perceived this as needed.

nuttycom:
Recording that #1957 subsumes this PR: it branches from this PR's head, retains the batch-Prove/outlook arc unchanged as its base commits, and then reworks the lanes per the rulings recorded there (every migration operation drives through `advance_migration`; the read-only selection exports this branch pinned are no longer consumed — see the discussion on zcash/librustzcash#2938). Leaving this PR's disposition to its owner.

*Filed with Claude Code assistance.*

pacu (approved):

nuttycom:
Marking as draft until we decide whether to close it in favor of #1957

PR #1954, opened 2026-08-06
Merge `feature/ironwood-slipstream` changes back to `main`

nuttycom:
This PR brings `main` up to date with the `feature/ironwood-slipstream` work. Note that work for the 3.0.0 SDK release will continue to be based on and merge to `feature/ironwood-slipstream`; the purpose of this is to integrate all outstanding work so that we can begin working on coinholder polling on `main` without duplicating or conflicting with ongoing work on the feature branch.


rust/src/voting/db.rs
```diff
@@ -30,14 +33,37 @@ pub struct VotingDatabaseHandle {
...
+        // follow the registered base network — a modified-mainnet chain votes
+        // with mainnet hotkeys and HRPs — so derive it from CUSTOM_PARAMS via
+        // `parse_network` (which also errors if the custom network was never
+        // configured) rather than assuming Regtest.
+        let network = match network_id {
+            crate::NETWORK_ID_TESTNET => voting::types::Network::Testnet,
+            crate::NETWORK_ID_MAINNET => voting::types::Network::Mainnet,
+            crate::NETWORK_ID_REGTEST => {
+                use zcash_protocol::consensus::{NetworkType, Parameters};
+                match crate::parse_network(network_id)?.network_type() {
```
nullcopy, line 52: This doesn't make sense to me. Why would we represent a `network_id` such that it first matches as regtest, and then we parse it to learn that it's not actually regtest?

rust/src/migration_keystone.rs
nullcopy, line 1: I'm getting an unexpected diff in this file, when comparing against `feature/ironwood-slipstream`:
```
diff --git a/rust/src/migration_keystone.rs b/rust/src/migration_keystone.rs
index ebc28722..8eaffbf1 100644
--- a/rust/src/migration_keystone.rs
+++ b/rust/src/migration_keystone.rs
@@ -440,14 +440,16 @@ mod tests {
         base.serialize().expect("serialize")
     }

-    /// The batch request RETAINS a pre-existing Orchard `spend_auth_sig` rather than stripping it.
-    /// The only pre-signed actions in a wallet PCZT are the protocol padding dummies, which carry
-    /// no ZIP 32 derivation and whose `dummy_sk` the compact view has already cleared — stripping
-    /// their signature would leave an action neither the signer nor the wallet could ever
-    /// authorize. Upstream's `redact_pczt_for_batch_signer` states and enforces that contract; this
-    /// pins that we get the redacted request it produces.
+    /// The batch request STRIPS a pre-existing Orchard `spend_auth_sig` rather than retaining it.
+    /// The only pre-signed actions in a wallet PCZT are the protocol padding dummies, and the
+    /// Keystone batch-signing firmware rejects any request that carries an Orchard spend
+    /// authorization signature at all. Stripping strands nothing: the request is a signing
+    /// request, not the authoritative PCZT, and `apply_batch_signatures` puts the device's
+    /// signatures back onto the caller's retained unredacted bytes, which still hold the dummy's.
+    /// Upstream's `redact_pczt_for_batch_signer` states and enforces that contract; this pins that
+    /// we get the redacted request it produces.
     #[test]
-    fn build_sign_batch_qr_parts_retains_the_dummy_orchard_spend_auth_sig() {
+    fn build_sign_batch_qr_parts_strips_the_dummy_orchard_spend_auth_sig() {
         let unsigned = build_single_pool_orchard_pczt();

         // Sanity-check the premise: the IO-finalized PCZT already carries a dummy spend_auth_sig
@@ -490,14 +492,27 @@ mod tests {
         let request = BatchSignRequest::parse(batch.get_data()).expect("parse batch sign request");

         assert_eq!(request.pczts().len(), 1);
-        // The pre-signed dummy keeps its signature; the signer needs nothing further for it.
+        // No action reaches the device carrying a signature; the firmware rejects the whole
+        // request if one does.
         assert!(
             request.pczts()[0]
+                .orchard()
+                .actions()
+                .iter()
+                .all(|action| action.spend().spend_auth_sig().is_none()),
+            "redaction must strip every Orchard spend_auth_sig from the batch request",
+        );
+
+        // The caller's retained bytes are untouched, so the dummy's signature is still there to
+        // be combined with whatever the device returns.
+        assert!(
+            pczt::parse(&unsigned)
+                .expect("re-parse retained pczt")
                 .orchard()
                 .actions()
                 .iter()
                 .any(|action| action.spend().spend_auth_sig().is_some()),
-            "the dummy's signature must survive redaction",
+            "redaction must not disturb the caller's retained PCZT",
         );
     }
```

The final result looks good to me, but I wouldn't expect this to differ. Is this an improvement that should be backported to `feature/ironwood-slipstream`?

rust/src/lib.rs
nullcopy, line 1: I can't load the diff in UI, but on ~L107 you remove the feature gate around the voting module. Do we not want to leave the gate, and just add it to default features?

nullcopy (dismissed):
I'd be curious to know the answers to a few questions, but no blockers.

I diff'd against `feature/ironwood-slipstream` to check that the merge result will effectively be the slipstream branch + voting updates.

utACK 4021cd01ba2bd807a0f329b12aef069c5fb1f601

nuttycom:
Converted to draft; we will need another merge to bring in the released slipstream crate as the private pins are no longer resolvable.

nuttycom:
Re-pointed pins to slipstream-internal.

dannywillems (approved):
utACK 0076d56af31e84c8e3d2c7b1c8a6bd0d7ebbab1d. I checked the diff after @nullcopy approval. It is mostly pinning to our slipstream-internal.

PR #1957, opened 2026-08-07
Advance-driven migration lanes + librustzcash dedup

nuttycom:
**Draft — at reviewable-final state.** Part of #1806. The alternative to #1953 discussed on librustzcash #2938: every migration lane drives through `advance_migration`, and the SDK's duplication of `zcash_pool_migration` is removed.

### The architecture (each stage owner-ruled during review)

- **`advance_migration` is the sole top-level call.** It appears exactly once in the crate, inside `drive_advance`, whose only caller is the `zcashlc_migration_advance_step` conduit. Payload-taking executors (`zcashlc_migration_prove_transactions`, `zcashlc_migration_take_broadcast_transaction`; Swift `proveMigrationTransactions` / `performMigrationBroadcast`) execute instructions and never ask. The delivery serve rides the store's atomic broadcast seam, so the #2927 overdue re-spread fires on the delivery path and the wallet record binds at bytes-handout.
- **No capabilities for semantic goals.** The Swift surface is the conduit, instruction executors, and reads. Instruction payloads are opaque (no public initializers — possession proves the caller cranked). The kind-filtered queue peeks, the semantic drive-wrappers, and their reporting types are removed, not deprecated.
- **Read-only queries ride `transaction_statuses`**, preserving the engine's broadcast-outranks-proving precedence (two-tier `due_assuming_proving`).
- **The sync gate is behavior-based.** The post-broadcast time buffer is deleted — a fixed delay is a correlation *signature*, not protection. Only the in-flight broadcast marker survives, as crash-recovery liveness. Old gate files load gracefully.
- **The wallet-adapter fork is gone.** `Backend`'s 350-line near-verbatim copy of `WalletMigration` collapsed onto upstream; no stateful spending key exists anywhere — the signing entry points take `&orchard::keys::SpendingKey` per call, FVK-validated upstream against the account's UFVK (librustzcash #2951's `WrongSpendAuthority` boundary). The plan preview reads the engine's planned-transaction graph (single derivation, consumed by commit).
- **Keystone**: the ZIP-32 re-stamp the engine already applies is deleted (proven at the pre-annotation seam); the suite's long-standing keystone failure was root-caused (stale pin of an older redaction contract) and fixed.

### Status

- Suites: Rust **156/0** (the historical baseline failure is gone), Swift offline **788/0**. Every commit builds and tests green on its own.
- Pin: `kris/tmp-respread-plus-adapter-2` — librustzcash main + #2927 + #2951's head, deliberately excluding #2938's selectors. Reverts to a published rc once #2927 and #2951 land.
- App adoption: the handoff document (call-site-cited, in the zodl-ios working tree under `.plans/`) enumerates the ~5-file production impact and the staged behavior-gate cluster.
- Open: the `migrationExtractBroadcastTx` keep-or-remove call (no in-SDK production consumer behind its documented retention rationale); minor follow-ups listed in the final review.


Sources/ZcashLightClientKit/Migration/OrchardMigration.swift
```diff
@@ -298,14 +347,49 @@ actor OrchardMigration {
...
+    /// `executeNextPendingTransfer`'s own default judging at the scanned tip, an estimate-advanced
+    /// sweep inside the estimate's lead window sees `.broadcast` and proves nothing while the
+    /// delivery lane sees nothing due — and neither lane progresses until the scan catches up.
     func finalizeReadyTransfers() async throws -> Int {
-        try await welding.migrationProvePending(for: accountUUID)
+        var totalProved = 0
+        while true {
+            guard case .prove(let transactions)? = try await advanceStep(useEstimatedTip: false)?.step else {
+                return totalProved
+            }
```
nuttycom: This is wrong. `advanceStep`'s return value *may* result in the need prove pending transactions, but the caller should not drive the engine seeking such a result. I think `finalizeReadyTransfers` should be removed.

Chlup (approved):

PR #1958, opened 2026-08-11
Span-rate tip estimator + bare Replan/Reevaluate discriminants

Chlup:
Part of #1806. Two migration hardening changes from the gardening pass; each commit builds and tests green on its own, and the branch merges cleanly into the current base tip.

### Tip estimator: span rate over an absence-scale window (f2798af0)

The measured seconds-per-block was the mean of consecutive header-time deltas, each clamped to [5, 150] s, over the last 100 scanned blocks. Two structural failures, both field-hit on 2026-08-08 (testnet run data37):

- The clamped-delta mean is biased **low** on bimodal spacing (testnet's min-difficulty bursts + gaps): bursts dominate by count, so the tip projection overshoots the real chain.
- The 100-block window (~33 min on a fast chain) measures whatever regime preceded sleep and extrapolates it across an hours-long absence; the incident window ran 20 s/block against a night that averaged 31 s/block.

Since upstream #2927 the projection is not advisory: `advance_migration` sizes **and persists** its overdue re-spread at the estimate. The incident compounded two wake-time re-spreads into ~2,690 blocks of phantom deferral (real overnight advance: ~900), parking a live run ~15 wall-hours past the chain and rendering as the 7h/17h/12h ETA wobble.

`secondsPerBlock()` is now the **span** rate — total header-time span over total height span, clamped once — immune to the bimodal bias and free of the adjacent-heights assumption; `sampleWindow` widens 100 → 1000 so the sampled span covers absence-scale regime changes. The incident's exact 100-row window ships as a regression fixture (header times that regress 140 s included), and the stale "an over-estimate costs nothing worse than an early due answer" contract doc is rewritten to the post-#2927 reality.

Constants heritage: clamps + fallback still mirror the Android SDK; the method and window deliberately diverge — the Kotlin estimator has the same bias and needs the same fix.

### Split Replan and Reevaluate onto their own bare discriminants (93a11081)

The collapsed ATTEND projection is retired end-to-end: the FFI marshals upstream's `AdvanceStep::Replan`/`Reevaluate` as `ZCASHLC_ADVANCE_STEP_REPLAN=6` / `REEVALUATE=7` (5 stays a documented hole — a stale header must decode to nil, never to another step), `MigrationAdvanceStep` gains `.replan`/`.reevaluate` and loses `requiresAttention(id:)`, and the Blocker-scanning id synthesis is deleted. Neither case carries an id: upstream names no transaction on either. Docs and the four test files follow; a new pin holds the retired discriminant hole at nil.

### Gates

- f2798af0: OfflineTests 807/807 (iPhone 17 Pro sim); zodl-testnet BUILD SUCCEEDED against this checkout
- 93a11081 (branch tip): cargo test 162/0; iOS-sim suite 806/806
- Plain `swift build` fails on this branch with 28 pre-existing errors from the stale macOS FFI slice, with or without these changes.

### DIFF
--- a/Sources/ZcashLightClientKit/Migration/ChainTipEstimator.swift
+++ b/Sources/ZcashLightClientKit/Migration/ChainTipEstimator.swift
@@ -10,27 +10,45 @@ import Foundation
 /// (`ZcashRustBackendWelding.migrationBlockRateSamples(window:)`).
 ///
 /// The projection feeds the migration delivery lane's `estimatedTip` inputs
-/// (`migrationHasOverdueTransfers(for:estimatedTip:)` / `migrationAdvanceStep(for:estimatedTip:)`),
-/// where it may only ACCELERATE scheduled-height due-ness — the rust side takes
-/// `max(scanned, estimated)` and always evaluates expiry against the SCANNED tip — so an
-/// over-estimate here costs nothing worse than an early "due" answer, and an under-estimate
-/// degrades to the scanned-tip behavior.
+/// (`migrationHasOverdueTransfers(for:estimatedTip:)` / `migrationAdvanceStep(for:estimatedTip:)`)
+/// AND — since upstream #2927 — the overdue re-spread inside `advance_migration`, which is
+/// **judged and sized at the estimate and PERSISTED**: an over-estimate is no longer "at worst an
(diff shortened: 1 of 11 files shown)


LukasKorba (approved):
utACK

PR #1964, opened 2026-08-13
[MOB-1737] Vendor ZODL Slipstream as an opt-in AGPL variant; default artifact goes MIT-clean

pacu:
## Why

`zodl-slipstream` (AGPL-3.0-only, dual-licensed by Znewco) is an unconditional dependency of `libzcashlc`, so every released xcframework ships AGPL code to every SDK consumer — regardless of whether they ever construct `SlipstreamSynchronizer`. Runtime opt-in is irrelevant to AGPL (distribution triggers it), and ZODL's `LICENSE-EXCEPTIONS.md` §2(b) explicitly withholds the AGPL §7 App Store permission from third parties. Classic Spend-before-Sync does not depend on the engine, so the SDK can be MIT-clean by default without losing it.

## What

Every release ships **two variants** (ffmpeg-kit pattern; full design in `docs/ZODL_SLIPSTREAM_VENDORING_PLAN.md`):

| | `X.Y.Z` (default) | `X.Y.Z-zodl-slipstream` (opt-in, `exact:` only) |
|---|---|---|
| Artifact | `libzcashlc.xcframework` — zero AGPL code | `libzcashlc-zodl-slipstream.xcframework` (superset) |
| Products | `ZcashLightClientKit` | + `ZODLSlipstream` (`SlipstreamSynchronizer`) |

- **Rust**: new `slipstream` cargo feature (`gpu` implies it); the whole `zcashlc_slipstream_*` FFI surface moves to `rust/src/slipstream_ffi.rs` behind one `cfg`; cbindgen wraps the declarations in `#if defined(ZCASHLC_FEATURE_SLIPSTREAM)` and packaging specializes the header with `unifdef`.
- **SwiftPM**: tools-version 5.9; exactly one `libzcashlc` binaryTarget per resolved graph, URL/checksum switched by `slipstreamVariantPinned` (flipped only on variant tags — the flag must live in the manifest bytes because SwiftPM's manifest cache is keyed on manifest content) or the local-dev marker file. The Swift Slipstream layer moves to `Sources/ZODLSlipstream/` behind ~220 `internal → package` promotions (never `public`); protobuf regeneration now passes `Visibility=Package`.
- **Guardrails**: `prepare-release.sh` and the clean CI job both hard-fail on any `zcashlc_slipstream` symbol or header residue in the clean artifact. SwiftPM's resolver categorically excludes pre-release tags from version ranges, so `from:` can never resolve a variant tag (verified against resolver source; see `docs/Versioning.md`).
- **CI**: clean/zodl-slipstream matrix; CodeQL builds the superset so every Swift source is scanned; the `authorize-slipstream` GitHub App machinery is removed — see the note below, this is now *provably* dead rather than merely believed dead.
- **Docs**: `Sources/ZODLSlipstream/NOTICE.md` (AGPL terms, App Store caveat, licensing@zodl.com, trademark/naming conditions), README variant guidance, `docs/Versioning.md`.

## The private-repo pin on `main` is inert — and this PR proves it

`main` carries `[patch.crates-io] slipstream-core = { git = "…/slipstream-internal.git", rev = … }` (from 0076d56a), which is why `authorize-slipstream` mints a token for that private repo on every cargo job. That patch **matches nothing in the dependency graph**: it is keyed `slipstream-core`, while the actual dependency is `zodl-slipstream` (public crates.io), and `zodl-slipstream` does not depend on `slipstream-core` either. Its only live effects were forcing a private-repo clone on every resolve and leaving two orphan packages (`slipstream-core 0.6.4`, `slipstream-gpuhash 0.0.1`) in `Cargo.lock`.

Removing the dead line: `cargo metadata` resolves with **zero private-repo access**, and the lock drops exactly those two entries (742 → 740 packages, no version changes). That is what makes the auth machinery removable. **Please sanity-check this claim** — it is the one change here that touches how CI authenticates.

## Reviewing this diff

75 files, +4,876 / −4,109 — but the reviewable surface is far smaller:

- **~2,834 lines are `Cargo.lock`** reordering from pruning those two stale entries.
- **~2,140 lines are a verbatim code move** (`lib.rs` → `slipstream_ffi.rs`). Review with `git diff --color-moved=dimmed-zebra` and only the genuine edits (`use crate::` rewiring, `pub(crate)` promotions) stay bright.
- **~220 changes are one-word `internal → package` promotions**, compiler-driven; none is `public`, so the SDK's public API is unchanged. Swift file moves are 100%-similarity renames.

That leaves roughly **400 lines of real design surface**, concentrated in three mechanisms: the cargo feature (the licensing boundary), the cbindgen `#ifdef` + `unifdef` header specialization, and the `Package.swift` variant graph.

## Verification

- `cargo test`: **198 passing** (default) / **234 passing** (`--features slipstream`); `cargo tree` shows zero `zodl` crates for default features.
- Swift: all targets compile in variant mode; clean-variant checkout (marker removed, caches purged) builds with zero errors and zero slipstream targets. Offline suites pass in both modes.
- `unifdef -U` strips the clean header to zero slipstream references; the `nm` purity gate was validated against a genuinely contaminated artifact — it detects the symbols and aborts.
- Downstream: zodl-ios adopts with a 2-file change ([zodl-inc/zodl-ios#1991](https://github.com/zodl-inc/zodl-ios/pull/1991)) — clean build, 1,225 tests, 0 failures.

## Notes for reviewers

- Stacked on **#1985** (the `ext_slipstream_v_*` view-name fix), which targets `main` and is reviewable on its own. Merge #1985 first; this then retargets to `main`. This PR's diff against #1985 carries no behavioural change — it is purely the code move.
- Identifiers were renamed to the full product name after review, so the manifest constants are now `zodlSlipstreamFFIURL` / `zodlSlipstreamFFIChecksum`.
- `zodlSlipstreamFFIURL/Checksum` in `Package.swift` are intentionally empty until the first dual release uploads the variant artifact; the manifest fails loudly if the variant is selected without a local FFI build.
- Darkside suites have not been run for either variant.

🤖 Drafted and opened by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

https://claude.ai/code/session_01AkdBfgXpNvWmnNQXWjfA9G


rust/src/lib.rs
```diff
@@ -102,7 +102,10 @@ mod migration_finalize;
 mod migration_keystone;
 mod migration_plan_cache;
 mod migration_turnstile;
+#[cfg(feature = "slipstream")]
 mod retained_marks;
+#[cfg(feature = "slipstream")]
+mod slipstream_ffi;
```
nuttycom: These should be done using `cfg(all(zodl_slipstream, feature = "slipstream"))` flags or something of the sort, so that `--all-features` does not enable it.

rust/build.rs
```diff
@@ -30,6 +30,10 @@ fn main() {
     if let Ok(b) = cbindgen::Builder::new()
         .with_crate(crate_dir)
         .with_language(cbindgen::Language::C)
+        // The slipstream FFI surface (AGPL engine) is cfg-gated; emit it inside
+        // an #ifdef so one deterministic header serves both artifact variants.
+        // BuildSupport/Makefile specializes the shipped copy with unifdef.
+        .with_define("feature", "slipstream", "ZCASHLC_FEATURE_SLIPSTREAM")
```
nuttycom: This will change with the use of a non-`feature` cfg flag.

Scripts/init-local-ffi.sh
```diff
@@ -96,21 +105,32 @@ build_arm_xcframework() {
...
         esac
 
-        echo "Building $rust_target -> $slice ..."
+        echo "Building $rust_target -> $slice ...${SLIPSTREAM_LABEL}"
 
         # Ensure the Rust target is available (idempotent), then build it.
         # cargo is incremental, so repeat builds after small edits are fast.
         rustup target add "$rust_target"
-        cargo build --target "$rust_target" --release
+        cargo build --target "$rust_target" --release "${CARGO_FEATURES[@]}"
```
nuttycom: This will change when using a cfg-flag.

Sources/ZODLSlipstream/SlipstreamEngine.swift
```diff
@@ -288,15 +289,9 @@ public actor SlipstreamEngine {
...
-/// `max(bundled checkpoint, birthday+1)` fallback), `treeState` nil (the host keeps its
-/// birthday checkpoint). NEW: the reorg-safe recent server tree state, or nil when
-/// offline (the host keeps its bundled checkpoint defaults).
-struct SlipstreamRestoreAnchor {
-    let height: BlockHeight
-    let treeState: TreeState?
-}
+// `SlipstreamRestoreAnchor` itself lives in the core target
+// (Initializer+SlipstreamAnchor.swift) because `Initializer.slipstreamAnchorSource`
+// names it; only this engine-driven resolver belongs here.
```
nuttycom: Need more information about this move.

Cargo.toml
```diff
@@ -134,10 +132,17 @@ cc = "1.0"
 [features]
 default = []
 
+# ZODL Slipstream engine (AGPL-3.0-only) and its FFI surface. OFF by default —
+# the released libzcashlc.xcframework carries zero AGPL code. Enabled only for
+# the libzcashlc-zodl-slipstream artifact and local builds via
+# Scripts/init-local-ffi.sh --slipstream / rebuild-local-ffi.sh --slipstream.
+slipstream = ["dep:slipstream-core", "dep:tokio"]
```
nuttycom: Instead of this, we will need something like:
```
[target.'cfg(zodl_slipstream)'.dependencies]
slipstream-core = { version = "0.1.1", package = "zodl-slipstream"}

File: Sources/ZcashLightClientKit/Initializer+SlipstreamAnchor.swift
Comment (pacu) [inline Sources/ZcashLightClientKit/Initializer+SlipstreamAnchor.swift:1]: this extraction does not make much sense the Initialized still preserves the slipstream related property which is not something that should happen. Clients not using ZODL Slipstream should not get anything releated to it.

File: Makefile [hunk 6]
```diff
@@ -26,8 +26,20 @@ CARGO ?= cargo
 BUILD_SUPPORT := BuildSupport
 SCRIPTS := Scripts
 
+# SLIPSTREAM=1 builds/links the ZODL Slipstream (AGPL-3.0-only) variant: the
+# ffi-* targets compile the Rust with the `slipstream` cargo feature into the
+# parallel products-slipstream/ tree, and configure-local-ffi maintains the
+# .zodl-slipstream-variant marker Package.swift keys the target graph off.
+# See BuildSupport/Makefile and Sources/ZODLSlipstream/NOTICE.md.
+SLIPSTREAM ?= 0
```
pacu: the name is ZODL Slipstream. Features, config flags, product identifiers or anything related to it must have its complete name "ZODL Slipstream"

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -157,7 +157,7 @@ public class Initializer {
     /// `nil` (the legacy `SDKSynchronizer` path) keeps the host-side fetch below,
     /// byte-for-byte (the old sync path is frozen).
     /// Signature: (isRestore, birthday, latest bundled checkpoint height) → anchor.
-    var slipstreamAnchorSource: ((Bool, BlockHeight, BlockHeight) async -> SlipstreamRestoreAnchor?)?
+    package var slipstreamAnchorSource: ((Bool, BlockHeight, BlockHeight) async -> SlipstreamRestoreAnchor?)?
```
pacu: Non-ZODL-Slipstream variants should not get this code or this type.

rust/src/lib.rs
pacu: Done in a59dc29d — the gate is now `#[cfg(all(zodl_slipstream, feature = "zodl-slipstream"))]`, with `zodl_slipstream` declared in `[lints.rust] check-cfg` and set via `RUSTFLAGS` by every build path.

The feature is still there alongside the cfg because only a feature can make the *dependency* optional (cargo cannot gate a dependency on a cfg): the feature decides whether the crate is linked, the cfg decides whether our FFI surface is compiled. Neither alone yields an engine build.

Verified at the object level, since checking only our `zcashlc_slipstream_*` wrappers would prove the engine is unreachable but not that AGPL code is absent:

| build | engine object files in archive | tests |
|---|---|---|
| default | 0 | 198 pass |
| `--all-features` (no cfg) | **0** | 198 pass |
| `RUSTFLAGS="--cfg zodl_slipstream" --features zodl-slipstream` | 16 | 234 pass |

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

rust/build.rs
pacu: Right — updated in a59dc29d. `with_define()` always writes a `key = value` entry, so a bare cfg cannot be expressed through it; the defines now go in through `cbindgen::Config` directly, one entry for `zodl_slipstream` and one for `feature = zodl-slipstream`, both mapping to `ZCASHLC_ZODL_SLIPSTREAM`.

cbindgen renders the conjunction as `#if (defined(ZCASHLC_ZODL_SLIPSTREAM) && defined(ZCASHLC_ZODL_SLIPSTREAM))`, which `unifdef` still resolves correctly in both directions (clean header: 0 slipstream references; variant header: 13 declarations).

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Scripts/init-local-ffi.sh
pacu: Updated in a59dc29d. The local-FFI scripts and both Makefiles now export `RUSTFLAGS="${RUSTFLAGS:-} --cfg zodl_slipstream"` alongside `--features zodl-slipstream`, appending rather than clobbering a caller-supplied `RUSTFLAGS` (dry-run check: `make -n cargo-macos ZODL_SLIPSTREAM=1 RUSTFLAGS="-C target-cpu=native"` yields `RUSTFLAGS="-C target-cpu=native --cfg zodl_slipstream"`). The flag is also renamed `--zodl-slipstream`, per the full-product-name convention.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZODLSlipstream/SlipstreamEngine.swift
pacu: Fair — and on reflection the move was wrong, so it is withdrawn rather than justified. `SlipstreamRestoreAnchor` is back in `Sources/ZODLSlipstream/` where it started, and `Initializer+SlipstreamAnchor.swift` is deleted.

The reason it had leaked into core was that `Initializer` named the type in its provisioning hook. That hook is now engine-agnostic — `provisioningAnchorSource`, returning `(height: BlockHeight, treeState: TreeState?)?`, both types core already owns — so core gains no new type and names no engine, and the add-on maps its own anchor type into the tuple when installing the closure.

It stays a lazily-invoked closure rather than a value passed into `initialize()`: the anchor is only needed once `listAccounts()` shows the wallet is unprovisioned, so resolving it eagerly would spend — and leak — a network round trip on every launch of an already-provisioned wallet.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Initializer+SlipstreamAnchor.swift
pacu: Agreed, and fixed in a59dc29d — the file is deleted and `SlipstreamRestoreAnchor` moved back into `Sources/ZODLSlipstream/`.

`Initializer` no longer carries anything engine-specific: the hook is now `provisioningAnchorSource`, returning `(height: BlockHeight, treeState: TreeState?)?` — a tuple of types core already owns — so a wallet built without ZODL Slipstream gains no type and reaches no engine code. Outside two inert spots noted below, the word no longer appears in core at all.

Some seam has to exist, because the anchor must be resolved *during* `initialize()`, only once `listAccounts()` proves the wallet is unprovisioned — resolving it eagerly would spend and leak a network round trip on every launch of a provisioned wallet. What your comment requires, and what this now does, is that the seam is not branded.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Makefile
pacu: Done in a59dc29d — full name everywhere: cargo feature `zodl-slipstream`, cfg `zodl_slipstream`, C define `ZCASHLC_ZODL_SLIPSTREAM`, make variable `ZODL_SLIPSTREAM`, script flag `--zodl-slipstream`, `products-zodl-slipstream/` tree, `test-offline-zodl-slipstream` target, `zodlSlipstreamVariantPinned` / `zodlSlipstreamFFIURL` / `zodlSlipstreamFFIChecksum` in the manifest, and `ZODLSlipstreamOfflineTests` / `ZODLSlipstreamDarksideTests` (directories renamed to match).

One thing I did NOT rename, so you can decide: the pre-existing public Swift types — `SlipstreamSynchronizer`, `SlipstreamEngine`, `SlipstreamSnapshot`. They predate this PR, they are already namespaced by the `ZODLSlipstream` module (`ZODLSlipstream.SlipstreamSynchronizer`), and renaming them breaks call sites in zodl-ios. Say the word and I will rename them here and update zodl-inc/zodl-ios#1991 to match.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Initializer.swift
pacu: Fixed in a59dc29d — that property is gone. The replacement names no engine and introduces no type:

```swift
package var provisioningAnchorSource:
    ((Bool, BlockHeight, BlockHeight) async -> (height: BlockHeight, treeState: TreeState?)?)?
```

`resolveSlipstreamAnchor` became `resolveProvisioningAnchor`, and the `[slipstream]` log tags in the init flow are now `[init]`. A clean-variant build has no `ZODLSlipstream` target in the package graph at all (verified: 809 tests pass, 0 slipstream targets), so non-ZODL-Slipstream clients get neither the code nor the type.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -569,8 +575,8 @@ public class Initializer {
             let recoverUntil = chainTip.map { "tip \($0)" } ?? "unknown"
             logger.info(
                 walletBirthday != nil
-                    ? "[slipstream] init flow: RESTORE — birthday \(self.walletBirthday), recover_until=\(recoverUntil)"
```
pacu, line 572: this rename is unhelpful

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -150,14 +150,21 @@ public class Initializer {
...
+    /// fetch, byte-for-byte (the old sync path is frozen).
+    ///
+    /// It is resolved lazily, here, rather than by the caller ahead of time: the
+    /// anchor is only needed once `listAccounts()` proves the wallet is
+    /// unprovisioned, and resolving it eagerly would spend — and leak — a network
+    /// round trip on every launch of an already-provisioned wallet.
+    ///
+    /// Signature: (isRestore, birthday, latest bundled checkpoint height) →
+    /// (provisioning height, tree state at that height).
+    package var provisioningAnchorSource:
```
pacu: this rename is unhelpful

Sources/ZcashLightClientKit/Utils/DBActor.swift
```diff
@@ -22,7 +22,7 @@ import Foundation
 ///
 /// WHAT THIS ACTOR GUARANTEES: no two Swift-initiated writes ever interleave.
 ///
-/// WHAT IT DOES NOT AND CANNOT GUARANTEE: the slipstream engine writes to the same database
+/// WHAT IT DOES NOT AND CANNOT GUARANTEE: an alternative sync engine may write to the same database
```
pacu: this rename is unhelpful

Sources/ZcashLightClientKit/Transaction/TxResubmitter.swift
```diff
@@ -8,8 +8,8 @@ import Foundation
 /// Engine-independent resubmission core: prunes stale submit plans and
 /// resubmits the transactions that qualify, throttled to at most once per
 /// 300-second window. Shared by the old sync pipeline's `TxResubmissionAction`
-/// and `SlipstreamSynchronizer`, so the retry policy lives in exactly one place.
-final class TxResubmitter {
+/// and any alternative synchronizer, so the retry policy lives in exactly one place.
```
pacu: this rename is unhelpful

Sources/ZcashLightClientKit/Block/DatabaseStorage/SimpleConnectionProvider.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 import SQLite
 
 class SimpleConnectionProvider: ConnectionProvider {
-    /// Seconds SQLite retries a locked DB before erroring. The Slipstream engine writes `data.db` (WAL
+    /// Seconds SQLite retries a locked DB before erroring. An alternative sync engine may write `data.db` (WAL
```
pacu: this rename is unhelpful

Sources/ZcashLightClientKit/Migration/OrchardMigrationHost.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
 
 /// The per-synchronizer owner of all Orchard -> Ironwood migration machinery.
 ///
-/// `SDKSynchronizer` (and, later, `SlipstreamSynchronizer`) each hold exactly one host. It owns the
+/// `SDKSynchronizer` (and any alternative synchronizer) each hold exactly one host. It owns the
```
pacu: this rename is unhelpful

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -47,7 +47,7 @@ public struct LightWalletEndpoint {
 }
 
 // Sendable: a pure value type (String/Int/Bool fields). [v0.7 P1b] Alternate-endpoint
-// lists cross the SlipstreamSynchronizer actor boundary, which makes hosts building
+// lists cross an alternative synchronizer's actor boundary, which makes hosts building
```
pacu: this rename is unhelpful

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -586,8 +592,8 @@ public class Initializer {
         } else {
             logger.info(
                 existingAccounts.isEmpty
-                    ? "[slipstream] init flow: OPEN — no seed supplied, not creating an account"
```
pacu, line 589: this rename is unhelpful

Sources/ZcashLightClientKit/Initializer.swift
```diff
@@ -615,13 +621,13 @@ public class Initializer {
...
-    /// `treeState` = the reorg-safe recent server tree state, or nil offline (the caller
-    /// keeps the bundled checkpoint defaults).
-    private func resolveSlipstreamAnchor(
-        _ anchorSource: (Bool, BlockHeight, BlockHeight) async -> SlipstreamRestoreAnchor?,
+    /// [v2.1 E-6] Resolve the provisioning anchor. RESTORE ⇒ `chainTip` = the
+    /// recover_until height (the anchor source's policy guarantees one: live tip, or an
+    /// offline max(bundled checkpoint, birthday+1) — never NULL). NEW ⇒ `treeState` =
+    /// the reorg-safe recent server tree state, or nil offline (the caller keeps the
+    /// bundled checkpoint defaults).
+    private func resolveProvisioningAnchor(
```
pacu: this rename is unhelpful

Sources/ZcashLightClientKit/Synchronizer.swift
```diff
@@ -1415,7 +1415,7 @@ public extension Synchronizer {
     //
     // Default implementations so adding the migration group to the protocol is not a
     // source-breaking change for downstream/stacked conformers (in particular the
-    // `SlipstreamSynchronizer` stack, until it carries its own implementations). Conformers with
+    // alternative synchronizer stack, until it carries its own implementations). Conformers with
```
pacu: this rename is unhelpful

Sources/ZODLSlipstream/SlipstreamSynchronizer.swift
```diff
@@ -292,15 +293,18 @@ public actor SlipstreamSynchronizer: Synchronizer {
         let anchorServer = currentEndpoint
         let anchorNetwork = initializer.network
         let anchorTorDir = await slipstreamTorDirPath()
-        initializer.slipstreamAnchorSource = { isRestore, birthday, fallbackCheckpoint in
```
pacu, line 295: this rename is unhelpful

Sources/ZcashLightClientKit/Utils/DBActor.swift
pacu: Agreed — reverted in 6c7d312a. This one was the worst of the set: the comment exists to explain *why* the busy timeout is there, and "the slipstream engine writes to the same database" is the reason. Rewriting it to "an alternative sync engine **may** write" turned a load-bearing fact into speculation, so a future reader could reasonably conclude the hazard is hypothetical and drop the timeout. This file is now byte-identical to `main`.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Block/DatabaseStorage/SimpleConnectionProvider.swift
pacu: Agreed — reverted in 6c7d312a. This one was the worst of the set: the comment exists to explain *why* the busy timeout is there, and "the slipstream engine writes to the same database" is the reason. Rewriting it to "an alternative sync engine **may** write" turned a load-bearing fact into speculation, so a future reader could reasonably conclude the hazard is hypothetical and drop the timeout. This file is now byte-identical to `main`.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Transaction/TxResubmitter.swift
pacu: Agreed — reverted in 6c7d312a. Replacing the concrete `SlipstreamSynchronizer` with "any alternative synchronizer" made the sentence ungreppable: a reader can no longer jump to the conformer it is talking about, and gains nothing in exchange.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Migration/OrchardMigrationHost.swift
pacu: Agreed — reverted in 6c7d312a. Replacing the concrete `SlipstreamSynchronizer` with "any alternative synchronizer" made the sentence ungreppable: a reader can no longer jump to the conformer it is talking about, and gains nothing in exchange.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Synchronizer.swift
pacu: Agreed — reverted in 6c7d312a. Replacing the concrete `SlipstreamSynchronizer` with "any alternative synchronizer" made the sentence ungreppable: a reader can no longer jump to the conformer it is talking about, and gains nothing in exchange.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Initializer.swift
pacu: Agreed — reverted in 6c7d312a. Replacing the concrete `SlipstreamSynchronizer` with "any alternative synchronizer" made the sentence ungreppable: a reader can no longer jump to the conformer it is talking about, and gains nothing in exchange.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Initializer.swift
pacu: Agreed — reverted in 6c7d312a. `slipstreamAnchorSource` and `resolveSlipstreamAnchor` have their names back. Renaming did not remove anything from the clean product, it only obscured who installs the hook and why — and the AGPL constrains the code compiled into a distributed binary, not the vocabulary of MIT-licensed source. What does NOT come back is the engine's *type*: `SlipstreamRestoreAnchor` stays in the ZODLSlipstream module and this hook keeps returning a tuple of types core already owns, which is the structural separation your earlier comment asked for.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Initializer.swift
pacu: Agreed — reverted in 6c7d312a. `slipstreamAnchorSource` and `resolveSlipstreamAnchor` have their names back. Renaming did not remove anything from the clean product, it only obscured who installs the hook and why — and the AGPL constrains the code compiled into a distributed binary, not the vocabulary of MIT-licensed source. What does NOT come back is the engine's *type*: `SlipstreamRestoreAnchor` stays in the ZODLSlipstream module and this hook keeps returning a tuple of types core already owns, which is the structural separation your earlier comment asked for.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZODLSlipstream/SlipstreamSynchronizer.swift
pacu: Agreed — reverted in 6c7d312a. `slipstreamAnchorSource` and `resolveSlipstreamAnchor` have their names back. Renaming did not remove anything from the clean product, it only obscured who installs the hook and why — and the AGPL constrains the code compiled into a distributed binary, not the vocabulary of MIT-licensed source. What does NOT come back is the engine's *type*: `SlipstreamRestoreAnchor` stays in the ZODLSlipstream module and this hook keeps returning a tuple of types core already owns, which is the structural separation your earlier comment asked for.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Initializer.swift
pacu: Agreed — reverted in 6c7d312a. The `[slipstream]` tag is the token you filter logs on to follow the engine's provisioning path; `[init]` matches every other init line and loses that.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

Sources/ZcashLightClientKit/Initializer.swift
pacu: Agreed — reverted in 6c7d312a. The `[slipstream]` tag is the token you filter logs on to follow the engine's provisioning path; `[init]` matches every other init line and loses that.

---
🤖 Drafted and posted by [Claude Code](https://claude.com/claude-code) on @pacu's behalf.

PR #1965, opened 2026-08-13
Fix created transaction readback for broadcast

zcash-harry:
## Summary

- return newly created transaction bytes directly from the authoritative wallet store through Rust FFI
- decouple transaction submission from the derived `v_transactions` history projection
- continue shielding and sends when the creation-time history overview is unavailable
- preserve every available history event in multi-transaction proposals when another overview is missing

## Observed failure mechanism

After `zcashlc_create_proposed_transactions` created and stored a transaction, the Swift SDK immediately reconstructed the broadcast payload by querying `v_transactions`. For the affected wallets, that derived history view did not contain the newly stored transaction, so the SDK threw `ZTREE0001 transactionRepositoryEntityNotFound` before submission.

`v_transactions` is rooted in received-output and received-output-spend relations. A wallet-created transaction with neither relation can therefore exist in the wallet store with its serialized bytes while remaining absent from that history projection. A database-level synthetic reproduction confirms this state and confirms that the new FFI lookup returns the same transaction bytes.

The exact reason the reported transactions lacked a matching transparent spend relation is not yet established. That is a separate librustzcash investigation; it does not make a derived history projection an appropriate dependency of broadcasting.

The broadcast path now reads the serialized transaction from the wallet store. The history view is used only to enrich the creation-time event, so a missing overview no longer prevents submission or suppresses other available overviews in the same batch.

Linear: https://linear.app/zodl/issue/MOB-1703/error-when-shielding-mining-payouts-ios-v39

## User impact

Shielding transparent mining payouts and normal sends can proceed even when the transaction-history projection has not exposed the newly created row.

## Validation

- `make ffi-all`
- `cargo check --lib`
- `cargo test transaction_data_tests` — 2 tests, 0 failures
- `swift test --filter BroadcasterTests` — 17 tests, 0 failures
- `swift test --filter OfflineTests` — 870 tests, 0 failures
- `git diff --check`

`make lint` was unavailable locally because SwiftLint is not installed.


Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -1271,7 +1271,7 @@ struct ZcashRustBackend: ZcashRustBackendWelding {
     func createProposedTransactions(
         proposal: FfiProposal,
         usk: UnifiedSpendingKey
-    ) async throws -> [Data] {
+    ) async throws -> [CreatedTransaction] {
```
str4d: Rather than modifying this FFI wrapper method to not return txids (which doesn't reduce the number of FFI calls), we should add a separate FFI wrapper method that takes a txid and returns whatever information we have about it, if any (`CreatedTransaction` would then not be the right name for it, as it would work for any txid including received txs).

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -1308,6 +1308,37 @@ struct ZcashRustBackend: ZcashRustBackendWelding {
         return txIds
     }
 
+    @DBActor
+    func getTransaction(txId: Data) async throws -> TransactionData? {
+        guard txId.count == 32 else {
```
dannywillems: Very nit: we can use a constant instead of the magic number 32.

PR #1966, opened 2026-08-13
Add a database-level reproduction of the missing v_transactions row

dannywillems:
`v_transactions` is built from the union of received outputs and spends of received outputs; `sent_notes` reaches it only through a LEFT JOIN and so cannot produce a row on its own. A transaction is therefore visible in history only if one of its inputs was recorded, or it has a wallet-internal shielded output. A stored, valid transaction that satisfies neither is dropped, and the send path read that omission as "no such transaction" and raised ZTREE0001 before broadcasting.

The existing coverage for this stubs the repository error, so it pins the SDK's reaction to a configured stub rather than to a database that actually drops the row. These tests migrate a fixture wallet to the current schema and exercise the real view: a fixture transaction is found, a transaction with no notes rows is not (mined and unmined alike), and the same row is shown to be fully readable in `transactions`.

Two further tests cover the readback this branch is stacked on: the row the view drops is returned by `getTransaction`, and an unknown txid reports absence rather than an error. The stored bytes are copied from a real fixture transaction rather than fabricated, because the readback parses the `raw` column and re-encodes it, so the broadcast bytes are only equal to the stored bytes if that round trip is exact.

## What these tests do not establish

They do not show which operation produces such a row. They insert one directly.

An earlier version of this description claimed shielding and cross-pay have neither a recorded input nor a wallet-internal output. That was wrong, and thanks to @pacu for catching it. At the pinned revision the write path should record a transparent spend through `mark_transparent_utxo_spent`, and an ordinary shielding transaction's output *is* wallet-internal change, recorded through `put_received_note` — so it should not reach this state. The Ironwood arms are present in both the view and the write path there.

The likelier origins are a silent miss in those two markers, which both no-op on an unknown outpoint or nullifier and whose return value only feeds `detectable_via_scanning`, or an exact-amount transaction with no change. The shape that structurally lacks an internal output is shielding to an *external* receiver with no change, which is what librustzcash's own `shielding_coinbase_to_orchard_receiver_delivers_via_ironwood` models; whether the mining-payout flow takes that path is unverified.

Root cause is open and belongs upstream. This PR is about the SDK's behaviour once the row is missing, not about why it goes missing.

### DIFF
--- a/Tests/OfflineTests/MissingEntityReproTests.swift
+++ b/Tests/OfflineTests/MissingEntityReproTests.swift
@@ -12,10 +12,11 @@
 //  history only if at least one of its inputs was already recorded, or it has a wallet-internal
 //  shielded output (change or self-send).
 //
-//  Shielding and cross-pay both have NO wallet-internal output, which removes the second half of
-//  that rule and leaves the input match as the only thing creating the row. These tests pin the
-//  resulting behaviour: the transaction is fully present in `transactions`, and invisible through
-//  the view.
+//  A shielding or cross-pay send normally has both halves: recorded input spends and a
+//  wallet-internal change output. The field failure means neither was recorded for the affected
+//  wallets — which write went missing is an open librustzcash question. These tests pin the
+//  resulting behaviour rather than its upstream cause: a transaction in that state is fully
+//  present in `transactions`, and invisible through the view.
 //
 
 import SQLite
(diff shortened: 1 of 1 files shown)


Tests/OfflineTests/MissingEntityReproTests.swift
```diff
@@ -0,0 +1,246 @@
...
+            orphan.raw,
+            "the broadcast bytes must equal the stored bytes: the readback parses and re-encodes "
+            + "the row, so any divergence here would broadcast a different transaction than the "
+            + "one the wallet recorded"
+        )
+    }
+
+    /// A txid the wallet has never stored is reported as absent rather than as an error, so the
+    /// caller can tell "not stored" from "lookup failed".
+    func testGetTransactionReturnsNilForAnUnknownTxId() async throws {
```
dannywillems, line 448: This is mostly a test for the Rust backend. It is unrelated to actually reproducing the bug.

Chlup (approved):
utACK

PR #1969, opened 2026-08-13
Fix created transaction readback for broadcast

zcash-harry:
## Summary

- return newly created transaction bytes directly from the authoritative wallet store through Rust FFI
- decouple transaction submission from the derived `v_transactions` history projection
- continue shielding and sends when the creation-time history overview is unavailable
- preserve every available history event in multi-transaction proposals when another overview is missing

## Observed failure mechanism

After `zcashlc_create_proposed_transactions` created and stored a transaction, the Swift SDK immediately reconstructed the broadcast payload by querying `v_transactions`. For the affected wallets, that derived history view did not contain the newly stored transaction, so the SDK threw `ZTREE0001 transactionRepositoryEntityNotFound` before submission.

`v_transactions` is rooted in received-output and received-output-spend relations. A wallet-created transaction with neither relation can therefore exist in the wallet store with its serialized bytes while remaining absent from that history projection. A database-level synthetic reproduction confirms this state and confirms that the new FFI lookup returns the same transaction bytes.

The exact reason the reported transactions lacked a matching transparent spend relation is not yet established. That is a separate librustzcash investigation; it does not make a derived history projection an appropriate dependency of broadcasting.

The broadcast path now reads the serialized transaction from the wallet store. The history view is used only to enrich the creation-time event, so a missing overview no longer prevents submission or suppresses other available overviews in the same batch.

Linear: https://linear.app/zodl/issue/MOB-1703/error-when-shielding-mining-payouts-ios-v39

## User impact

Shielding transparent mining payouts and normal sends can proceed even when the transaction-history projection has not exposed the newly created row.

## Validation

- `make ffi-all`
- `cargo check --lib`
- `cargo test transaction_data_tests` — 2 tests, 0 failures
- `swift test --filter BroadcasterTests` — 17 tests, 0 failures
- `swift test --filter OfflineTests` — 870 tests, 0 failures
- `git diff --check`

`make lint` was unavailable locally because SwiftLint is not installed. GitHub's SwiftLint check passes.

Replaces #1965 so the change can target `feature/ironwood-slipstream`; GitHub would not allow the base of the existing stacked PR to be changed.


rust/src/ffi.rs
```diff
@@ -829,6 +829,95 @@ impl SymmetricKeys {
...
+pub unsafe extern "C" fn zcashlc_free_transaction_data(ptr: *mut FfiTransactionData) {
+    if !ptr.is_null() {
+        let transaction_data = unsafe { Box::from_raw(ptr) };
+        free_ptr_from_vec(transaction_data.raw, transaction_data.raw_len);
+        drop(transaction_data);
+    }
+}
+
+#[cfg(test)]
+mod transaction_data_tests {
```
dannywillems: Is it expected to have a module for testing in the middle of the file? It doesn't seem to be the case for the other extern definition.

rust/src/ffi.rs
```diff
@@ -829,6 +829,95 @@ impl SymmetricKeys {
 
 pub type TxIds = SymmetricKeys;
 
+/// Transaction data stored by the wallet, with the bytes required for broadcast.
+///
+/// A null `raw` pointer represents an unknown transaction or one whose raw bytes are unavailable.
+/// `expiry_height` is the consensus value; zero means that transaction expiry is disabled.
+#[repr(C)]
+pub struct FfiTransactionData {
```
dannywillems: Nit: prefixing with `Ffi` might not be informational as it could be initially, as we're already in the `ffi` module, and other structures do not seem to follow this.

rust/src/ffi.rs
```diff
@@ -829,6 +829,95 @@ impl SymmetricKeys {
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn zcashlc_free_transaction_data(ptr: *mut FfiTransactionData) {
+    if !ptr.is_null() {
+        let transaction_data = unsafe { Box::from_raw(ptr) };
+        free_ptr_from_vec(transaction_data.raw, transaction_data.raw_len);
+        drop(transaction_data);
+    }
+}
+
+#[cfg(test)]
```
dannywillems: I don't know if these tests actually bring any value. They are just testing the constructors, checking that the fields are set to particular values. It doesn't exist for the other structures, if I'm not mistaken.

dannywillems (approved):
utACK 4ad0c5ad6d32568d43bf73c9f68615c97dbccf31

pacu (approved):
Approving at `4ad0c5ad` per my re-review: both blocking findings are resolved (best-effort event enrichment; wallet-store-aware plan pruning), and the parsed-txid integrity check closes the round-trip concern. Remaining notes are non-blocking follow-ups. #1966, adapted to this base, should land as the companion test coverage.

PR #1971, opened 2026-08-14
Merge `feature/ironwood-slipstream` changes back to `main`

Chlup:
This PR brings `main` up to date with the `feature/ironwood-slipstream` work. Note that work for the 3.0.0 SDK release will continue to be based on and merge to `feature/ironwood-slipstream`; the purpose of this is to integrate all outstanding work so that we can begin working on coinholder polling on `main` without duplicating or conflicting with ongoing work on the feature branch.


LukasKorba (approved):
utACK

PR #1987, opened 2026-08-19
Remove ZODL Slipstream so 3.0.0 ships MIT-only

### LINKED ISSUE #1986: Remove ZODL Slipstream so 3.0.0 ships MIT-only
The `zodl-slipstream` crate is AGPL-licensed and landed on the 3.0.0 line during<br>development. The SDK is MIT, so shipping 3.0.0 as-is would place AGPL-derived<br>code inside artifacts this repository presents as MIT-only.

Remove Slipstream and its entire footprint including anything added to generalized protocols or<br>interfaces solely to support it, so that the 3.0.x line reads as though the engine<br>had never landed. No history rewrite.

Verification must show no AGPL-licensed crate in the resolved dependency graph for<br>every Apple target the xcframework is built from, and no Slipstream symbols in the<br>built binaries.

pacu:
Closes #1986


nuttycom (approved):
utACK 9fdb552b77f81418fe538f15ad650e5e6ef75cfc

PR #1989, opened 2026-08-21
Better Rust errors

dannywillems:
Fixes MOB-1201:
https://linear.app/zodl/issue/MOB-1201/error-from-rust-layer-when-calling-zcashrustbackendcreatetoaddress

## The problem

Two defects compounded into an error report that carried no information.

**The rust error string was captured and then discarded.** `lastErrorMessage(fallback:)` already
pulled the rust text into the `ZcashError` associated value, but `errorDescription` rendered only
`"\(code.rawValue): \(message)"`, where `message` is a static per-case sentence generated from the
case's doc comment. So every report read exactly `"ZRUST0002: Error from rust layer when calling
ZcashRustBackend.createToAddress"`, with the actual cause sitting unread in the payload.

**Four distinct FFI entry points all threw `rustCreateToAddress`.** On `main`,
`ZcashRustBackend.swift:372`, `:400`, `:429` and `:1296` throw the same code, so a report could not
say whether the failure happened while BUILDING a proposal or while signing an already-confirmed
one. `SDKSynchronizer` and `SlipstreamSynchronizer` each carried a `catch`/re-throw to recover the
URI case after the fact; both are deleted here.

## What changed

The rust layer classifies these failures instead of flattening them to a string.
`rust/src/error_report.rs` maps `zcash_client_backend`'s `Error` onto a stable `ErrorKind`
discriminant plus a constant message, and carries both across the FFI in `FfiErrorReport`.

The redaction is enforced by typing, not by discipline. `classify` is generic over the six `Error`
type parameters and takes **no `Display` or `Debug` bound on any of them**, so the payloads holding
amounts, addresses and note identifiers are literally unrenderable inside it; parametricity leaves
no way to observe them. The `message` field is `&'static str`, which makes `format!` ill-typed at
the assignment. Amounts travel out of band as integers for the single case that needs them.

On the Swift side `RedactedRustError` carries the kind, the redacted message and the optional
amounts, and `ZcashError.detail` appends the message to `errorDescription`. `detail` matches on the
payload TYPE, so the raw strings still carried by the other `rust*` cases cannot reach a report by
accident. Each entry point gets its own code, and the two conditions a wallet should render as
ordinary wallet state get dedicated cases:

| code | case |
|---|---|
| `ZRUST0151` | `rustProposeTransfer` |
| `ZRUST0057` | `rustProposeTransferFromURI` |
| `ZRUST0129` | `rustProposeSendMaxTransfer` |
| `ZRUST0152` | `rustProposeOrchardToIronwoodMigration` |
| `ZRUST0002` | `rustCreateToAddress`, now the signing step only |
| `ZRUST0153` | `rustProposalScanRequired` |
| `ZRUST0154` | `rustProposalInsufficientFunds(available:required:)` |

The unredacted rust text is still logged on-device at `debug!`.

### On reusing arti's `safelog`

@str4d suggested looking at the Tor crates. I read `safelog` and did not use it, because it does
not fit this shape: `Sensitive<T>` and `Redactable` help when you own the `Display` impl of the
sensitive type, but here the sensitive data is interpolated by UPSTREAM impls. The string
`"Insufficient balance (have {}, need {})"` is formatted inside `zcash_client_backend`, and there is
no hook mid-`Display` to intercept it. Withholding the bound so the payload cannot be rendered at
all is sound where wrapping it after the fact is not, and it costs no dependency.

## Breaking changes

The associated value of `rustCreateToAddress`, `rustProposeTransferFromURI` and
`rustProposeSendMaxTransfer` changes from `String` to `RedactedRustError`. Pattern matches that bind
the payload stop compiling. `ZcashError` gains `detail: String?`, so any snapshot test pinning an
exact `localizedDescription` needs updating. Both documented in MIGRATING.md.

## Scope

This surfaces the cause; it does not fix it. The conditions this now names are `ScanRequired` and
`InsufficientFunds`, and
[librustzcash#2329](https://github.com/zcash/librustzcash/issues/2329) (rewind affecting
spendability) is still open. `rustProposalInsufficientFunds(available:required:)` gives us the two
numbers that will confirm or refute it from a real report.

`zcashlc_propose_shielding` is deliberately untouched and still builds `anyhow!` strings. Extending
the classifier there is a follow-up, not a hidden part of this change.

Shipping is two-step: `Package.swift` pins the `2.8.0-rc.2` XCFramework, so the Swift half is inert
until a build carrying `zcashlc_take_last_error_report` is published.

## Test plan

Verified on this base (rebased onto `main` @ 31a35c56), not only on the branch point:

- `cargo test --lib error_report` - 6 passed, 0 failed. Covers classification of `ScanRequired` and
  `InsufficientFunds`, the out-of-band amounts, the no-error case, FFI round-trip of kind and
  amounts, the unclassified fallback, and that every arm yields a constant message.
- `swift build` - clean.
- `swift test --filter "ZcashErrorLocalizedTests|MigrationLogicTests"` - 70 passed, 0 failed.
- Sourcery 2.3.0 regeneration is byte-identical to the committed
  `ZcashError.swift` / `ZcashErrorCode.swift`.

`testEveryRedactedCarrierRendersItsDetail` is a regression guard on the Stencil template, and it was
mutation-proved rather than assumed: the first version of my filter compared a Sourcery `TypeName`
OBJECT against a string literal, matched nothing, and generated an empty `detail` switch, which
would have shipped this fix as a silent no-op while every single-case test still passed.
Reintroducing that filter makes the new test fail on all five codes; the fix is
`value.typeName.name`.

Manual verification on testnet: trigger a send while the wallet is mid-scan and confirm the failure
now surfaces as `ZRUST0153` rather than `ZRUST0002`; trigger a send above the spendable balance and
confirm `ZRUST0154` carries both `Zatoshi` amounts.

Not run: `DarksideTests` (needs a local lightwalletd). `PaymentURIFulfillmentTests` is touched only
in an `XCTFail` message.


Sources/ZcashLightClientKit/Error/Sourcery/ZcashError.stencil
```diff
@@ -25,6 +25,21 @@ public enum ZcashError: Equatable, Error {
         }
     }
 
+    /// The redacted detail carried by this error, if any.
+    ///
+    /// Only a `RedactedRustError` payload is rendered here. That is the whole point of the type:
+    /// a raw `String` from `lastErrorMessage(fallback:)` may still hold an amount or an address,
+    /// so it must not reach `errorDescription` and from there an error report. Widening this to
+    /// other payload types means redacting them at the rust boundary first.
+    public var detail: String? {
```
dannywillems, line 34: Important for review.

Sources/ZcashLightClientKit/Rust/ZcashRustBackend.swift
```diff
@@ -2710,6 +2723,62 @@ nonisolated func lastErrorMessage(fallback: String) -> String {
     }
 }
 
+/// Takes the last Rust error as a classified, redacted report, clearing it exactly as
+/// `lastErrorMessage(fallback:)` does.
+///
+/// Use this instead of `lastErrorMessage(fallback:)` wherever the resulting `ZcashError` may be
+/// shown to a user or submitted in an error report. The two must not be combined for one
+/// failure: whichever runs first consumes the error.
+nonisolated func lastErrorReport(fallback: String) -> RedactedRustError {
```
dannywillems, line 2732: Important part too

nuttycom (approved):
utACK 1b5a2ec32ed8b01bac6fac7f4edb7183759697d3

PR #1997, opened 2026-08-29
Release zcash-swift-wallet-sdk 2.7.0

### LINKED ISSUE #1996: zcash-swift-wallet-sdk release version 2.7.0

nuttycom:
Closes #1996

Release `2.7.0`, following `2.7.0-rc.3`.

The base of this pull request is `release/2.7.0`, which starts at `2.5.2`,
the last published release. Its diff is therefore exactly what users receive
relative to that release, rather than the intervening development history.

## This is a draft: the artifact phase has not run

Reviewing this diff is the first of two phases. Once it looks right, a
maintainer runs

    ./Scripts/prepare-release.sh build <remote> 2.7.0

which extends this branch with:

- the version bump, across `Cargo.toml`, `Cargo.lock` and
  `BuildSupport/platform-Info.plist`;
- the `libzcashlc` XCFramework, built for all five Apple architectures and
  uploaded as a draft GitHub release;
- `Package.swift`, rewritten so its binary target points at that release and
  carries its checksum.

When that has happened a comment will be added here and this pull request will
be marked ready for review. **Do not merge before then.** Merging early tags a
release whose `Package.swift` still points at the previous version's
XCFramework.

After merging, `./Scripts/release.sh <remote> 2.7.0` verifies the
checksum against the uploaded asset, signs and pushes the tag, and publishes
the release.


Scripts/lib/release-lib.sh
```diff
@@ -0,0 +1,283 @@
+# Shared helpers for the release scripts. Source this; do not execute it.
+#
+#   . "$(dirname "$0")/lib/release-lib.sh"
+#
+# Everything here is either a pure text transform or a preflight predicate, so
+# that Scripts/tests/run-tests.sh can exercise it without a network, a git
+# remote, or a GitHub token.
+#
+# Written for bash 3.2, which is what macOS ships.
```
pacu: This is not correct. MacOS has bash 3.2 but its official shell is zsh. 

It can be said that until Apple removes it, it can be guaranteed that bash 3.2 will be present in macOS systems but it's not "what it ships with". That's zsh

Scripts/tests/fixtures/CHANGELOG.md
pacu, line 1: This is not really a changelog 😓

Scripts/tests/fixtures/CHANGELOG-empty-unreleased.md
pacu, line 1: This is not really a changelog 😓

Sources/ZcashLightClientKit/Block/Actions/UpdateSubtreeRootsAction.swift
pacu: Filename header doesn't match the file name

Sources/ZcashLightClientKit/Checkpoint/Checkpoint.swift
pacu, line 50: The initializer should be implicitly defined by swift compiler

pacu (commented):
Left some small suggestions non-blocking

Scripts/tests/fixtures/CHANGELOG.md
nuttycom: It's a fixture for script testing.

pacu (approved):

PR #1999, opened 2026-09-03
Release zcash-swift-wallet-sdk 2.7.0

### LINKED ISSUE #1996: zcash-swift-wallet-sdk release version 2.7.0

nuttycom:
Replacement for #1997 because I botched the release process.

Closes #1996

Release `2.7.0`, following `2.7.0-rc.3`.

The base of this pull request is `release/2.7.0`, which starts at `2.5.2`,
the last published release. Its diff is therefore exactly what users receive
relative to that release, rather than the intervening development history.

## This is a draft: the artifact phase has not run

Reviewing this diff is the first of two phases. Once it looks right, a
maintainer runs

    ./Scripts/prepare-release.sh build <remote> 2.7.0

which extends this branch with:

- the version bump, across `Cargo.toml`, `Cargo.lock` and
  `BuildSupport/platform-Info.plist`;
- the `libzcashlc` XCFramework, built for all five Apple architectures and
  uploaded as a draft GitHub release;
- `Package.swift`, rewritten so its binary target points at that release and
  carries its checksum.

When that has happened a comment will be added here and this pull request will
be marked ready for review. **Do not merge before then.** Merging early tags a
release whose `Package.swift` still points at the previous version's
XCFramework.

After merging, `./Scripts/release.sh <remote> 2.7.0` verifies the
checksum against the uploaded asset, signs and pushes the tag, and publishes
the release.


pacu (approved):
utACK

PR #2001, opened 2026-09-03
Merge v2.7.x into v3.0.x

nuttycom:
Forward-merges `maint/v2.7.x` into `maint/v3.0.x`, then reworks `CHANGELOG.md` so the
3.0.0 notes are the delta a 2.7.0 consumer actually needs.

Three commits, reviewable in order: the merge itself, then two changelog passes.

## Conflict resolution

| Conflict | Resolution |
| --- | --- |
| `Cargo.toml`, `Cargo.lock`, `Package.swift` | Version records keep 3.0.0. |
| `Tests/DarksideTests/*` (12 files) | Accepted 2.7.x's deletion of the suite; 3.0.x's edits were API adaptation to tests that no longer exist. The `.gitkeep` naming the deleting commit remains. |
| `Scripts/prepare-release.sh`, `Scripts/release.sh` | Took 2.7.x's two-phase rewrite over `Scripts/lib/release-lib.sh` wholesale. It already subsumes 3.0.x's pre-release detection (`is_prerelease` / `prerelease_flag`, applied in both directions, which is stronger than what 3.0.x had). |
| `docs/ci.md` | Took 2.7.x's; 3.0.x's block described the retired flow. |
| `.github/workflows/swift.yml` | Kept 3.0.x's enumerated `actions/cache`. 2.7.x still used `Swatinem/rust-cache`, which caches `~/.cargo` wholesale — including `~/.cargo/git`. Dropped that step's `if: cache-hit != 'true'` guard, because 2.7.x's new `make test-rust` step runs unconditionally and needs the cache. |
| `CHANGELOG.md` | Reconstructed rather than patched — git had aligned 3.0.x's `2.7.0-rc.1` tail against 2.7.x's `2.7.0` section, so both sides needed the same `## Removed` / `## Fixed` context. |

### Carried forward by hand

2.7.x's script rewrite does **not** carry 3.0.x's fork-publishing fix, so `produce_artifacts`
now falls back through `GITHUB_REPOSITORY` before the canonical repo:

```sh
: "${GH_REPO:=${GITHUB_REPOSITORY:-$DEFAULT_REPO}}"
```

Without it a fork's `workflow_dispatch` 403s against upstream. Its doc comment was sitting
above the wrong function in the incoming file; it moved onto `produce_artifacts`.

## Two things that needed attention beyond the conflicts

**`BuildSupport/platform-Info.plist` auto-merged wrongly, without a conflict.** 3.0.x still
carried a stale `0.8.1`; 2.7.x bumped it to `2.7.0` and reordered the keys, so git silently
stamped **2.7.0 onto the 3.0.x framework**. Resolved to `3.0.0`, keeping 2.7.x's key layout
(`set_plist_version` addresses by key via PlistBuddy, and the test fixture assumes that
layout). All four version records — `Cargo.toml`, `Cargo.lock`, `Package.swift`, the plist —
now agree.

**The combination broke one of 2.7.x's new script tests.**
`test_build_ffi_workflow_interpolations_are_env_assignments_only` required *every* `${{ }}`
in `build-ffi.yml` to be a bare `KEY: ${{ … }}` line; 3.0.x's cache key is the first
legitimate interpolation in a `with:` block. Its comment states the property it actually
wants — no interpolation spliced into a `run:` script, the zizmor template-injection class —
so the check now tests that instead and leaves YAML values alone. Verified with positive
controls: it catches injection both in a `run: |` block scalar and in an inline `run:`.
Reverting the cache change was the alternative and is worse.

## CHANGELOG rework

The 2.8.0 release candidates and the 2.6.0 alphas were never published, so the section a
reader meets above 3.0.0 must be one that was. That makes 2.7.0 the baseline.

- **Added** the 2.7.0 section; **dropped** `2.7.0-rc.1`–`rc.3`, which it consolidates.
- **Removed** `2.8.0-rc.1`–`rc.3` and `2.6.0-alpha.1`–`alpha.6`. The alphas only ever
  described the voting surface — added and withdrawn between them — and checkpoints the
  3.0.0 block already spans.
- **Rebased the 3.0.0 section on 2.7.0**, 620 lines to 289.

Ordering is now `Unreleased → 3.0.0 → 2.7.0 → 2.5.2 → …`.

### Note for review

This edits an already-published `# 3.0.0` heading, which `AGENTS.md` otherwise forbids. It
is deliberate: the section was written as the delta from `2.8.0-rc.3`, so removing that
section left it describing a range no reader can start from.

The larger cuts, each verified against the two tags' trees:

- **The migration privacy sync buffer's removal.** `migrationPrivacySyncBufferDuration`
  appears in zero files at 2.7.0 *and* zero at 3.0.0 — ~20 lines about withdrawing something
  that never shipped.
- **The librustzcash git pin.** The entry named branch `kris/tmp-respread-plus-adapter-2` at
  rev `fa3e1de5`; 3.0.0 pins rev `13ce6c4e`. Wrong revision, and the versions it resolves to
  were already stated in the entry below it.
- **The migration surface's own bug fixes.** The surface is new in 3.0.0, so a fix to it
  names no change a 2.7.0 consumer can observe. The properties they leave behind — inputs
  marked spent at proving time, anchors marked durable, proving yielding between proofs — are
  stated once, where the surface is introduced.
- **The retired ZRUST codes**, the upstream PR numbers, the design rationale, and the quoted
  design rulings.
- **The six witnesses-fix gate defects**, compressed to the one thing a consumer acts on: the
  repair check runs again on the next launch.

### Found during the audit

`ZcashError.rustSetTransactionStatus` is **`ZRUST0111` at 2.7.0 and `ZRUST0107` at 3.0.0**,
and `ZRUST0111` now names `rustMigrationTakeBroadcastTransaction`. Code matching on the
string does not stop compiling — it matches the wrong error. This was documented nowhere; it
is now a `## Changed` entry.

`# Unreleased` carries one entry: `ZIP318Kind.canonicalCrossingPayment`, the only public API
change the merge itself brings onto this line. 3.0.0's notes described the case, but its
shipped enum did not have it.

## Verification

- Changelog checked **both directions** against the 2.7.0 and 3.0.0 trees: every added public
  type and all 37 added `Synchronizer` members are named in the section, and every symbol the
  section names differs between the trees. No bullet in 3.0.0 is >55% similar to one in 2.7.0.
- `Scripts/tests/run-tests.sh` — 74 assertions, 4 failures, all `plutil` / `PlistBuddy` not
  existing on Linux. They pass on `macos-15`.
- `cargo metadata --locked` succeeds, so manifest and lockfile agree.
- Auto-merged `Makefile`, `AGENTS.md`, `build-ffi.yml`, `.gitignore` audited against both
  sides; no stale references to the removed `Scripts/start-release.sh`.

**Not verified locally:** `swift build` / `make check`, which need the Apple XCFramework.
Worth a run on a Mac before merge — CI covers it.

---

🤖 Description written with [Claude Code](https://claude.com/claude-code) assistance.


pacu (approved):
utACK

PR #2002, opened 2026-09-03
Merge v3.0.x into main

nuttycom:
Forward-merges `maint/v3.0.x` — which already carries the `maint/v2.7.x` merge from #2001 — into
`main`. One merge commit.

## The consequential part: Slipstream leaves `main`

#1987 removed the alternative sync engine on the 3.0.x line, and this merge carries that onto
`main`: the synchronizer, engine wrapper and FFI value types, their offline and darkside suites,
the architecture proposal, `rust/src/retained_marks.rs`, the composite action that minted a read
token for the private engine repository, and the `slipstream-core` dependency with the `gpu`
feature and `[patch.crates-io]` pin that served it.

`main` had touched Slipstream twice since the removal, and neither was engine work:
`SlipstreamSynchronizer.swift` lost its URI-proposal error wrapper as part of MOB-1201's sweep, and
the dependency pin moved to `slipstream-internal` on 2026-08-07, before #1987. So the deletion
costs nothing, but it is a feature leaving `main` and worth a deliberate look rather than a nod.

**Voting is unaffected.** It lives only on `main`, and the merge keeps it: `zcash_voting` stays an
unconditional dependency, so 3.0.x's commented-out `voting` feature and its `cfg(zcash_voting)`
lint allowance stay gone. 58 `zcashlc_voting_*` symbols are exported by Rust and 58 are referenced
from Swift, matching exactly.

## Conflict resolution

| Conflict | Resolution |
| --- | --- |
| `Slipstream/SlipstreamSynchronizer.swift`, `.github/actions/authorize-slipstream/action.yml` | Took the deletion, per above. |
| `Tests/DarksideTests/PaymentURIFulfillmentTests.swift` | Took the deletion; the rest of that suite was retired on 2.7.x. |
| `Cargo.toml` | Union of both sides' deletions: `main` dropped the `voting` feature line when it reinstated voting, 3.0.x dropped `gpu`, so `[features]` keeps only `default`. `[patch.crates-io]` loses the slipstream pin. |
| `Cargo.lock` | Re-resolved rather than hand-merged — see below. |
| `CHANGELOG.md` | Unreleased rebased on 3.0.0 — see below. |

### `Cargo.lock`

The two conflict hunks looked like the wgpu subtree, but one of them contained `multimap 0.10.1`,
which is `prost-build 0.14`'s dependency and not wgpu's — resolving them by hand would have dropped
a live crate. The lock is instead re-resolved from `main`'s. Against `main` it drops exactly the
slipstream and wgpu subtree and moves `libzcashlc` to 3.0.0; no other crate changes version, and
`cargo metadata --locked` exits 0.

## `MIGRATING.md` auto-merged wrongly, without a conflict

`main`'s top heading was `to _Unreleased_` and 3.0.x's was `to 3.0.0`. Git took 3.0.x's heading with
`main`'s sections underneath, filing the `RedactedRustError` guide and the voting guides under a
release that shipped neither.

Resolved by restoring a `to _Unreleased_` heading over the three unreleased sections and starting
`to 3.0.0` at the eleven that are 3.0.0's. The voting restoration guide also moves up out of the
`v2.8.0-rc.1` block, where the auto-merge left it — below 3.0.0, so a reader upgrading from 3.0.0
would never find it — carrying the two voting delta sections with it as subsections. References to
`2.7.0-rc.1`, which this line no longer records, become `2.7.0`, the release that removed voting.

Every content line of the auto-merged file is accounted for: nothing was dropped that does not have
a reworded replacement.

## `CHANGELOG.md`

`main`'s Unreleased section was written against `2.8.0-rc.3`, a section this line does not carry.
Rebased on 3.0.0, which is now the release above it.

Its second half turned out to be the 3.0.0 release content — the migration group, the `Broadcaster`
redesign, custom networks, the balance and proposal additions, and the fixes to all of them — and
is dropped as subsumed. The librustzcash pin entry named a revision this line does not use. What
survives is what a 3.0.0 consumer actually meets:

- shielded voting restored (3.0.0 shipped none, so the notes describe the surface as it stands
  rather than tracking how it changed while unreleased),
- MOB-1201's classified proposal errors,
- `ZcashSDK.nu63ConsensusBranchID`,
- `ZIP318Kind.canonicalCrossingPayment`.

571 lines to 89.

## Verification

Checked both directions against the `3.0.0` tag:

- No public declaration is removed. The `Synchronizer` protocol is identical at 96 members.
- The 210 added public declarations are all voting or MOB-1201; the four added error codes are
  `ZRUST0151`–`ZRUST0154`; every symbol the new section names is a real delta.
- The whole-tree diff against 3.0.0 touches only voting, the MOB-1201 error path, and those two
  constants — which is what confirms none of `main`'s Unreleased `## Fixed` entries were genuine
  deltas rather than fixes already inside 3.0.0.

Also: the three generated error files agree at 269 cases; `Cargo.toml`, `Cargo.lock`,
`Package.swift` and `platform-Info.plist` all read 3.0.0; `Scripts/tests/run-tests.sh` gives 74
assertions with 4 failures, all `plutil` / `PlistBuddy` absent on Linux, unchanged from before the
merge; and no reference to any removed symbol, script or document survives anywhere in the tree.

**Not verified locally:** `make check` and `make lint` — no `swift` or `swiftlint` on the machine
this was resolved on, and the package needs the Apple XCFramework. Worth a run on a Mac; CI covers
it.

**Note for whoever builds this branch:** the published 3.0.0 XCFramework that `Package.swift` names
predates the voting FFI, so a fresh checkout needs a local FFI build to link. CI builds the FFI from
source, so CI is unaffected.

## Test plan

- CI: `swift.yml` (`make build`, `make test-offline`, `make test-rust`), `swiftlint.yml`,
  `proto-sync.yml`, `zizmor.yml`.
- Locally on a Mac: `make ffi-macos && make configure-local-ffi && make check`, then `make lint`.
  The FFI build is required, not optional — this merge changes `rust/src/`.
- Manual, on testnet: sync a wallet to confirm the Slipstream removal left the default sync path
  intact, and exercise one voting round end to end, since voting and the removed engine were the
  two surfaces this merge had to keep apart.

---

🤖 Opened with [Claude Code](https://claude.com/claude-code) assistance.


Chlup (approved):
utACK

pacu (approved):
utACK

